PR #2, opened 2023-04-01
feat: foreword

### LINKED ISSUE #1: Foreword
Foreword: A little bit of history about Cairo, who this book is for, and what can you expect to find in the book.

Estimated: ± 1 hour

enitrat:
Closes #1


fricoben (approved):
Guud. 

Be careful tho to not work too much before we find an agreement with the starkware team in order to make our docs compatible with what they're building

PR #5, opened 2023-04-01
feat: gh-pages CI

enitrat:


enitrat:
Closing because there was a default config available in the Github settings

PR #19, opened 2023-04-04
feat: control flow

### LINKED ISSUE #16: Control Flow
A section on the Cairo control flow: if statements, loops(still alpha, might be for another issue)
This section doesn't include match statements, which are specific to Cairo (and Rust)

Estimated time: 1 hour

hurrikaanig:
Close #16


src/ch02-05-control-flow.md
```diff
@@ -0,0 +1,64 @@
...
+
+<span class="filename">Filename: main.cairo</span>
+
+```rust
+use debug::PrintTrait;
+
+fn main() {
+    let number = 3;
+
+    if number == 5 {
```
enitrat, line 17: The text below expects the condition checked to be "number < 5" here, either change the text or this line

src/ch02-05-control-flow.md
```diff
@@ -0,0 +1,64 @@
...
+```rust
+    let number = 5;
+```
+
+```console
+$ cairo-run main.cairo
+condition was true
+```
+
+It’s also worth noting that the condition in this code must be a bool. If the condition isn’t a bool, we’ll get an error.
```
enitrat, line 47: What kind of error? Include the error the console gives you

src/ch02-05-control-flow.md
```diff
@@ -0,0 +1,64 @@
...
+Because if is an expression, we can use it on the right side of a let statement to assign the outcome to a variable.
+
+<span class="filename">Filename: main.cairo</span>
+
+```rust
+use debug::PrintTrait;
+
+fn main() {
+    let condition = true;
+    let number = if condition { 5 } else { 6 };
```
enitrat, line 60: include a print statement in the example so that the reader can try it himself

enitrat (changes requested):
You will also need to add a part that demonstrates how `if... else if... else` are used

enitrat (approved):
💯

PR #21, opened 2023-04-04
Data type page

Cheelax:
Time spent: 3h


src/SUMMARY.md
```diff
@@ -11,9 +11,8 @@
   - [Installation](ch01-01-installation.md)
   - [Hello, World!](ch01-02-hello-world.md)
   - [Hello, Scarb!](ch01-03-hello-scarb.md)
-
-- [Common Programming Concepts](ch02-00-common-programming-concepts.md)
-
+  - [Common Programming Concepts](ch02-00-common-programming-concepts.md)
```
enitrat, line 14: you will need to rebase this with main, this should be in section 2. (cf file in https://github.com/cairo-book/cairo-book.github.io/pull/18)

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
+## Data Types
+
+Every value in Cairo is of a certain *data type*, which tells Cairo what kind of
+data is being specified so it knows how to work with that data. his section covers two subsets of data types: scalars and compounds.
```
enitrat, line 4: > his

This

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+
+Keep in mind that Cairo is a *statically typed* language, which means that it
+must know the types of all variables at compile time. The compiler can usually infer the desired type based on the value and its usage. In cases
+when many types are possible, we can use a cast method where we specify the desired output type.
+
+```Rust
+let guess = cast("42",u32);
```
enitrat, line 11: What? this is not valid cairo code

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+You’ll see different type annotations for other data types.
+
+### Felts Type
+
+In Cairo when you don't specify the type of a variable/argument, its type is a field element (represented by the keyword `felt`). In the context of Cairo, when we say “a field element” we mean an integer in the range -P/2 < x < P/2
```
enitrat, line 18: In Cairo, if you don't specify the type of a variable or argument, its type defaults to a field element, represented by the keyword `felt252`.

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+let guess = cast("42",u32);
+```
+
+You’ll see different type annotations for other data types.
+
+### Felts Type
+
+In Cairo when you don't specify the type of a variable/argument, its type is a field element (represented by the keyword `felt`). In the context of Cairo, when we say “a field element” we mean an integer in the range -P/2 < x < P/2 
+ where 
+ is a very large (prime) number (currently it is a 252-bit number, which is a number with 76 decimal digits). When we add, subtract or multiply and the result is outside the range above, there is an overflow, and the appropriate multiple of P
```
enitrat, line 20: where P is a very large prime number

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+ is added or subtracted to bring the result back into this range (in other words, the result is computed modulo P).
+
+The most important difference between integers and field elements is division: Division of field elements (and therefore division in Cairo) is not the integer division you have in many programming languages, where the integral part of the quotient is returned (so you get `7 / 3 = 2`). As long as the numerator is a multiple of the denominator, it will behave as you expect (`6 / 3 = 2`). If this is not the case, for example when we divide `7/3`, it will result in a field element `x` that will satisfy `3 * x = 7`. It won’t be `2.3333` because x has to be an integer. If this seems impossible, remember that if `3 * x` is outside the range -P/2 < x < P/2 
+, an overflow will occur which can bring the result down to 7. It’s a well-known mathematical fact that unless the denominator is zero, there will always be a value x satisfying `denominator * x = numerator`.
```
enitrat, line 24: If this seems implausible, keep in mind that if 3 * x falls outside the range -P/2 < x < P/2, an overflow occurs, potentially bringing the result down to 7

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Scalar Types
```
enitrat, line 26: Felts are scalar types and thus should be included

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+
+A *scalar* type represents a single value. Cairo has three primary scalar types:
```
enitrat, line 28: You need to mention the fact that these types are custom types defined in the core library and they're built from felts and add security features

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+
+| Length    | Unsigned |
+|-----------|----------|
+| 8-bit     | `u8`     |
+| 16-bit    | `u16`    |
+| 32-bit    | `u32`    |
+| 64-bit    | `u64`    |
+| 128-bit   | `u128`   |
+| 256-bit   | `u256`   |
+|arch(32bit)| `usize`  |
```
enitrat, line 49: `usize` are hardcoded to 32-bit (for now)

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+
+Each variant has an explicit size.
+As variables are unsigned, they can't contain a negative number. This code will cause the program to panic: 
+```Swift
+fn sub_u8s(x: u8, y: u8) -> u8 {
+    x - y
+}
+
+main() {
```
enitrat, line 58: fn main(){

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+| 16-bit    | `u16`    |
+| 32-bit    | `u32`    |
+| 64-bit    | `u64`    |
+| 128-bit   | `u128`   |
+| 256-bit   | `u256`   |
+|arch(32bit)| `usize`  |
+
+Each variant has an explicit size.
+As variables are unsigned, they can't contain a negative number. This code will cause the program to panic: 
+```Swift
```
enitrat, line 53: use ```rust here (no capitalization)

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+}
+
+main() {
+    sub_u8s(1,3)
+}
+```
+Unsigned ints can store numbers from 0 to 2<sup>n</sup> - 1,
+so a `u8` can store numbers from 0 to 2<sup>8</sup> - 1, which equals 0 to 255.
+
+Additionally, the `usize` type depends on the architecture of the
```
enitrat, line 65: Not true yet, I would just remove this

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+64 bits if you’re on a 64-bit architecture and 32 bits if you’re on a 32-bit
+architecture.
+
+You can write integer literals in any of the forms shown in Table 3-2. Note
+that number literals that can be multiple numeric types allow a type suffix,
+such as `57_u8`, to designate the type.
+
+<span class="caption">Table 3-2: Integer Literals in Cairo</span>
+
+| Numeric literals | Example       |
```
enitrat, line 76: There are also octal and binary literals
https://github.com/starkware-libs/cairo/commit/51ebb3ffd243d6eaee8111a44ee2b0bcda3d7ee9

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+The primary situation in which you’d use `usize` is when indexing some sort of collection.
+
+[//]: # "TODO: write panic chapter"
+
+> ##### Integer Overflow
+>
+> Let’s say you have a variable of type `u8` that can hold values between 0 and
+> 255. If you try to change the variable to a value outside that range, such as
+> 256, *integer overflow* will occur, which can result in one of two behaviors.
+> When you’re compiling in debug mode, Cairo includes checks for integer overflow
```
enitrat, line 91: Compiling in debug mode is not a thing, this is only valid from Rust. The compiler doesn't check for overflows at compile time.
Remove the part about panics

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+> 9.
+
+#### Numeric Operations
+
+Cairo supports the basic mathematical operations you’d expect for all the number
+types: addition, subtraction, multiplication, division, and remainder. Integer
+division truncates toward zero to the nearest integer. The following code shows
+how you’d use each numeric operation in a `let` statement:
+
+```rust
```
enitrat, line 105: You're demonstrating operations with felts. Use u128 instead, we want readers to use integer types as much as possible

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+must know the types of all variables at compile time. The compiler can usually infer the desired type based on the value and its usage. In cases
+when many types are possible, we can use a cast method where we specify the desired output type.
+
+```Rust
+let guess = cast("42",u32);
+```
+
+You’ll see different type annotations for other data types.
+
+### Felts Type
```
enitrat, line 16: You need to add a sentence explaining that while felts are the data type used by default in Cairo programs, programmers should abstain from using the `felt252` type when possible for security measures.

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+    let t = true;
+
+    let f: bool = false; // with explicit type annotation
+}
+```
+
+[//]: # "TODO: control flow section"
+The main way to use Boolean values is through conditionals, such as an `if`
+expression. We’ll cover how `if` expressions work in Cairo in the [“Control
+Flow”][control-flow]<!-- ignore --> section.
```
enitrat, line 148: You can add the reference (at the bottom of the md page) even if the ref doesn't exist yet.

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+
+#### The Short String Type
+
+Cairo’s `short sting` type is the language’s most primitive alphabetic type, the length is at most 31 characters as they are stored in a felt. Here are
```
enitrat, line 152: Cairo doesn't have a native type for strings, but you can store characters forming what we call a "short string" inside `felt252`s

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+fn main() {
+    let tup: (u32,u64,bool) = (10,20,true);
+}
+```
+
+The variable `tup` binds to the entire tuple because a tuple is considered a
+single compound element. To get the individual values out of a tuple, we can
+use pattern matching to destructure a tuple value, like this:
+
+```Rust
```
enitrat, line 187:
```rust
use debug::PrintTrait;
fn main() {
    let tup = (500, 6, true);

    let (x, y, z) = tup;

    if y == 6 {
        'y is six!'.print();
    }
}
```

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+It is possible to remove an element of an array by doing: 
+```Rust
+    let mut a = create_array();
+
+    a.pop_front().unwrap();
+```
+[//]: # "TODO: stack and heap"
+Arrays are useful when you want your data allocated on the stack rather than
```
enitrat, line 260: Remove this part, there are no such things as stack and heap in Cairo

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+the heap (we will discuss the stack and the heap more in [Chapter
+4][stack-and-heap]<!-- ignore -->).
+
+You write an array’s type by passing the type like this:
```
enitrat, line 264: You can pass the expected type of items inside the array when instantiating the array like this

```rust
let mut arr = ArrayTrait::<u128>::new();
```

(consistency with the above example)

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+You write an array’s type by passing the type like this: 
+
+```Rust
+let mut arr = array_new::<u128>();
+```
+
+##### Accessing Array Elements
+
+You can access elements of an array using indexing,
+like this:
```
enitrat, line 273: If we're presenting arrays, we need to present `get()`, `at()`, and explain their differences.

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+}
+```
+
+In this example, the variable named `first` will get the value `0` because that
+is the value at index `0` in the array. The variable named `second` will get
+the value `1` from index `1` in the array.
+
+##### Invalid Array Element Access
+
+By using the ArrayTrait::get() function because it returns an Option type, meaning that if you're trying to access an index out of bounds, it will return None instead of exiting your program, meaning that you can implement error management functionalities.
```
enitrat, line 293: formatting: `ArrayTrait::get()`

src/ch02-02-data-types.md
```diff
@@ -0,0 +1,318 @@
...
+
+
+
+```Rust
```
enitrat, line 297: incorrect snippet

```rust
use array::ArrayTrait;
use box::BoxTrait;

fn main() -> u128 {
    let mut arr = ArrayTrait::<u128>::new();
    arr.append(100_u128);
    let length = arr.len();
    match arr.get(length - 1_usize) {
        Option::Some(x) => {
            *x.unbox()
        },
        Option::None(_) => {
            let mut data = ArrayTrait::new();
            data.append('out of bounds');
            panic(data)
        }
    } // returns 100
}
```

enitrat (changes requested):
Thank you for the PR!

I made some comments about the changes to bring.

src/ch02-02-data-types.md
Cheelax: Can you confirm if you are talking about others scalar types or Compound Types too?

src/ch02-02-data-types.md
Cheelax: added this at the end of felt description:  The felt252 type is a fundamental type that serves as the basis for creating custom types in the core library. All other types, except felt252, are built on top of this foundational type by combining felts with additional security features.

It is highly recommended for programmers to use the custom types instead of the felt252 type whenever possible, as the custom types come with added security features that provide extra protection against potential vulnerabilities in the code. By utilizing these custom types, programmers can ensure that their programs are more secure and less susceptible to attacks or other security threats.

enitrat: > by combining felts

Not 100% accurate, u128 is built on top of felt252 but it's not a combination of felts

enitrat: > utilizing

using

> custom types

suggestion: standard/core types? I think it's best to be more precise, we're using the types from the corelib

src/ch02-02-data-types.md
```diff
@@ -1,36 +1,42 @@
...
- is added or subtracted to bring the result back into this range (in other words, the result is computed modulo P).
-
-The most important difference between integers and field elements is division: Division of field elements (and therefore division in Cairo) is not the integer division you have in many programming languages, where the integral part of the quotient is returned (so you get `7 / 3 = 2`). As long as the numerator is a multiple of the denominator, it will behave as you expect (`6 / 3 = 2`). If this is not the case, for example when we divide `7/3`, it will result in a field element `x` that will satisfy `3 * x = 7`. It won’t be `2.3333` because x has to be an integer. If this seems impossible, remember that if `3 * x` is outside the range -P/2 < x < P/2 
-, an overflow will occur which can bring the result down to 7. It’s a well-known mathematical fact that unless the denominator is zero, there will always be a value x satisfying `denominator * x = numerator`.
 
 ### Scalar Types
 
 A *scalar* type represents a single value. Cairo has three primary scalar types:
```
enitrat, line 24: four: felts!

src/ch02-02-data-types.md
```diff
@@ -1,36 +1,42 @@
...
 integers, booleans, and characters. You may recognize
```
enitrat, line 25: characters don't exist so back to three, lol

src/ch02-02-data-types.md
```diff
@@ -251,29 +241,43 @@ fn main() {
...
 ```
 
 ##### Accessing Array Elements
 
-You can access elements of an array using indexing,
-like this:
+You can access elements of an array using indexing.
+You can use `get()` or `at()` which have different functionning.
 
+The `get` function returns an `Option<Box<@T>>`, which means it returns a pointer to a box containing a reference to the element at the specified index if that element exists in the array. If the element doesn't exist, get returns `None`. This allows checking if the index is valid before accessing the element, which can help avoid segmentation faults.
```
enitrat, line 274: > reference

it's not a reference, it's a _snapshot_

src/ch02-02-data-types.md
```diff
@@ -251,29 +241,43 @@ fn main() {
...
 
-```Rust
+The `at` function, on the other hand, directly returns a reference to the element at the specified index using the `unbox()` operator to extract the value stored in a box. If the index is out of bounds, a panic error occurs. This function is more concise than `get`, but it doesn't provide any index checking, so it's more prone to segmentation faults.
```
enitrat, line 276: snapshot here as well

src/ch02-02-data-types.md
enitrat: This method is useful when you expect to access indices that may not be within the array's bounds and want to handle such cases gracefully without panics.

src/ch02-02-data-types.md
enitrat, line 276: > so it's more prone to segmentation faults.

segmentation faults don't exist in Cairo.

> This function is more concise than `get`, but it doesn't provide any index checking, so it's more prone to segmentation faults.

Replace by

You should only use `at` when you want the program to panic if the provided index is out of the array's bounds, which can prevent unexpected behavior.

src/ch02-02-data-types.md
```diff
@@ -251,29 +241,43 @@ fn main() {
...
+
+In summary, `get` is safer because it returns an `Option` that allows checking if the index is valid, while at is more concise but can cause segmentation faults if the index is out of bounds.
```
enitrat, line 278: Well, the whole point of using `get` is to provide error management, we should mention that.

replace by:

In summary, use `at` when you want to panic on out-of-bounds access attempts, and use `get` when you prefer to handle such cases gracefully without panicking.

src/ch02-02-data-types.md
```diff
@@ -251,29 +241,43 @@ fn main() {
...
-let mut arr = array_new::<u128>();
+```rust
+let mut arr = ArrayTrait::<u128>::new();
 ```
 
 ##### Accessing Array Elements
 
-You can access elements of an array using indexing,
-like this:
+You can access elements of an array using indexing.
```
enitrat, line 271: This is not true, you can't access elements with indexing... (if you meant like arr[2])

src/ch02-02-data-types.md
```diff
@@ -290,23 +294,20 @@ the value `1` from index `1` in the array.
 
 ##### Invalid Array Element Access
 
-By using the ArrayTrait::get() function because it returns an Option type, meaning that if you're trying to access an index out of bounds, it will return None instead of exiting your program, meaning that you can implement error management functionalities.
+By using the `ArrayTrait::get()` function because it returns an Option type, meaning that if you're trying to access an index out of bounds, it will return None instead of exiting your program, meaning that you can implement error management functionalities.
```
enitrat, line 297: Redundant, you can remove the section, but keep the example below to show how to manage errors manually

enitrat (changes requested):

src/ch02-02-data-types.md
```diff
@@ -1 +1,321 @@
...
+    let mut a = ArrayTrait::new();
+    a.append(0);
+    a.append(1);
+
+    let first = a.get(0_usize);
+    let second = a.get(1_usize);
+}
+```
+
+In this example, the variable named `first` will get the value `0` because that
```
enitrat, line 291: but if we use `get` we don't get the value, do we?

let's use `at` in this example, as `get` is used below.

```
let first = *a.at(0_usize);
let second = *a.at(1_usize);
```

enitrat (approved):
great, thank you!

PR #29, opened 2023-04-06
Feat/intro

Th0rgal:
This should be a good start. In the future it could be a good idea to add links to relevant parts of the book. What do you think? :)


src/ch00-00-introduction.md
```diff
@@ -1 +1,22 @@
...
+
+Another aspect that sets the language apart is its memory model. In Cairo, memory access is immutable, meaning that once a value is written to memory, it cannot be changed. Cairo 1 allows to abstract this constraint by simulating mutability but it still requires developers to think carefully about how they manage memory and data structures in their programs if they want to optimize performances.
```
enitrat, line 17: > Cairo 1 allows to abstract this constraint by simulating mutability

I don't really agree, for example Array values are immutable. I think it would be better reworded so that the reader is not confused

enitrat (approved):
LGTM once comment addressed!

src/ch00-00-introduction.md
Th0rgal: should be better now :)

enitrat: perfect!

PR #30, opened 2023-04-07
feat: method-syntax

### LINKED ISSUE #25: Method syntax
Introduce the method syntax. In Cairo, **methods are defined in the context of a trait**, unlike Rust where it can be defined for a struct/enum/trait

- Defining methods
- Explain the struct/trait/impl pattern
- Functions associated with a type (type of the `self` parameter)
- Multiple trait implementation for the same Struct

This section will need to introduce `Traits` and `Impls` but stay succint and make a reference to a future chapter about traits and generic types. You can add a TODO as a comment so we don't forget it later.
Estimated: 2 to 3 hours

enitrat:
Closes #25 

Time spent: 2 hours

### DIFF
--- a/src/ch04-03-method-syntax.md
+++ b/src/ch04-03-method-syntax.md
@@ -50,7 +50,10 @@ block with the signature of the method that we want to implement. Traits are not
 a specific type; only the `self` parameter of the method defines which type it can be used
 with. Then, we define an `impl` (implementation) block for `RectangleTrait`, that defines
 the behavior of the methods implemented. Everything within this `impl` block will be
-associated with the type of the `self` parameter of the method called.
+associated with the type of the `self` parameter of the method called. While it is technically
+possible to define methods for multiple types within the same `impl` block, it is not
+a recommended practice, as it can lead to confusion. We recommend that the type of the `self` parameter
+statys consistent within the same `impl` block.
 Then we move the `area` function within the `impl` curly brackets and change the first (and in this case, only)
 parameter to be `self` in the signature and everywhere within the body. In
 `main`, where we called the `area` function and passed `rect1` as an argument,
--- a/src/ch04-03-method-syntax.md
+++ b/src/ch04-03-method-syntax.md
@@ -3,11 +3,11 @@
 _Methods_ are similar to functions: we declare them with the `fn` keyword and a
 name, they can have parameters and a return value, and they contain some code
 that’s run when the method is called from somewhere else. Unlike functions,
-methods are defined within the context of a struct and their first parameter is
-always `self`, which represents the instance of the struct the method is being
+methods are defined within the context of a type and their first parameter is
+always `self`, which represents the instance of the type the method is being
 called on. For those familiar with Rust, Cairo's approach might be confusing,
-as methods cannot be defined directly on structs. Instead, you must define a trait
-and an implementation associated with the struct for which the method is intended.
+as methods cannot be defined directly on types. Instead, you must define a trait
+and an implementation associated with the type for which the method is intended.
 
 ### Defining Methods
 
@@ -42,26 +42,27 @@ fn main() {
 }
 ```
 
-<span class="caption">Listing 5-13: Defining an `area` method to use on the
+<span class="caption">Listing 4-13: Defining an `area` method to use on the
 `Rectangle` </span>
 
 To define the function within the context of `Rectangle`, we start by definining a `trait`
 block with the signature of the method that we want to implement. Traits are not linked to
-a specific struct; only the `self` parameter of the method defines which type it can be used
+a specific type; only the `self` parameter of the method defines which type it can be used
 with. Then, we define an `impl` (implementation) block for `RectangleTrait`, that defines
 the behavior of the methods implemented. Everything within this `impl` block will be
 associated with the type of the `self` parameter of the method called.
 Then we move the `area` function within the `impl` curly brackets and change the first (and in this case, only)
 parameter to be `self` in the signature and everywhere within the body. In
 `main`, where we called the `area` function and passed `rect1` as an argument,
-we can instead use _method syntax_ to call the `area` method on our `Rectangle`
+we can instead use the _method syntax_ to call the `area` method on our `Rectangle`
 instance. The method syntax goes after an instance: we add a dot followed by
 the method name, parentheses, and any arguments.
 
 Methods must have a parameter named `self` of the type they will be applied to for their first parameter.
 Note that we used the `@` snapshot operator in front of the `Rectangle` type in the function signature.
-By doing so, we indicate that this method borrows an immutable snapshot of the `Rectangle` instance.
-Methods can take ownership of `self`, borrow `self` immutably using snapshots as we’ve done here, or borrow `self` mutably,
+By doing so, we indicate that this method takes an immutable snapshot of the `Rectangle` instance, which is
+automatically created by the compiler when passing the instance to the method.
+Methods can take ownership of `self`, use `self` with snapshots as we’ve done here, or use a mutable reference to `self`
 using the `ref self: T` syntax, just as they can any other parameter.
 
 We chose `self: @Rectangle` here for the same reason we used `@Rectangle` in the function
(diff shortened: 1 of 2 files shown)


src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,311 @@
+## Method Syntax
+
+_Methods_ are similar to functions: we declare them with the `fn` keyword and a
+name, they can have parameters and a return value, and they contain some code
+that’s run when the method is called from somewhere else. Unlike functions,
+methods are defined within the context of a struct and their first parameter is
```
wraitii, line 6: This is incorrect for Cairo since 'methods' are defined in the context of a trait only

wraitii:
Might be a bit of a meta-comment but I would argue Cairo 1 does not have methods / This shouldn't in the struct chapter at all, since it has nothing to do with structs.

src/ch04-03-method-syntax.md
enitrat: This is what I initially wrote but I changed it back to this. They're defined in a trait/impl combo, but they remain associated with a struct

wraitii: Well they're not associated with a struct. You can define methods on felt252 or u8 which aren't structs, even on tuples which also aren't structs.

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,311 @@
...
+desired output. Methods can take multiple parameters that we add to the
+signature after the `self` parameter, and those parameters work just like
+parameters in functions.
+
+### Associated Functions
+
+All functions defined within an `impl` block are called _associated functions_
+because they’re associated with the type corresponding to the `self` argument. We can define
+associated functions that don’t have `self` as their first parameter (and thus
+are not methods) because they don’t need an instance of the type to work with.
```
wraitii, line 232: This block no longer makes sense in Cairo because of this distinction: functions are associated with the type corresponding to `self`, but you don't need `self`...

src/ch04-03-method-syntax.md
enitrat: ok sorry - with types. I will change to `types`

enitrat:
Changing it as draft until we get to the traits/impl sections. It would be a better fit there

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,309 @@
+## Method Syntax
+
+_Methods_ are similar to functions: we declare them with the `fn` keyword and a
+name, they can have parameters and a return value, and they contain some code
+that’s run when the method is called from somewhere else. Unlike functions,
+methods are defined within the context of a type and their first parameter is
+always `self`, which represents the instance of the type the method is being
+called on. For those familiar with Rust, Cairo's approach might be confusing,
+as methods cannot be defined directly on types. Instead, you must define a trait
```
wraitii, line 9: I think it should actually be `struct` in this sentence?

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,309 @@
...
+    rect1.area().print();
+}
+```
+
+<span class="caption">Listing 4-13: Defining an `area` method to use on the
+`Rectangle` </span>
+
+To define the function within the context of `Rectangle`, we start by definining a `trait`
+block with the signature of the method that we want to implement. Traits are not linked to
+a specific type; only the `self` parameter of the method defines which type it can be used
```
wraitii, line 50: Technically, it's just the first argument I think. There are some traits that don't use `self` for this in the core lib, like addition traits.

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,309 @@
...
+with. Then, we define an `impl` (implementation) block for `RectangleTrait`, that defines
+the behavior of the methods implemented. Everything within this `impl` block will be
+associated with the type of the `self` parameter of the method called.
```
wraitii, line 53: Unfortunately this is confusing to read, because it's actually on a per-method basis. The impl block is just a way to make generic types available (but there are none here)

Though I kind of agree that it would be suuuuper confusing to define methods on different types in the same `impl` block it's technically legal.

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,309 @@
...
+Then we move the `area` function within the `impl` curly brackets and change the first (and in this case, only)
+parameter to be `self` in the signature and everywhere within the body. In
+`main`, where we called the `area` function and passed `rect1` as an argument,
+we can instead use the _method syntax_ to call the `area` method on our `Rectangle`
+instance. The method syntax goes after an instance: we add a dot followed by
+the method name, parentheses, and any arguments.
+
+Methods must have a parameter named `self` of the type they will be applied to for their first parameter.
```
wraitii, line 61: If the name doesn't have to be `self` this should perhaps be changed as well.

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,309 @@
...
+using just `self` as the first parameter is rare; this technique is usually
+used when the method transforms `self` into something else and you want to
+prevent the caller from using the original instance after the transformation.
+
+Observe the use of the desnap operator `*` within the area method when accessing the struct's members.
+This is necessary because the struct is passed as a snapshot, and all of its field values are of type `@T`,
+requiring them to be desnapped in order to manipulate them.
+
+The main reason for using methods instead of functions, in addition to
+providing method syntax and not having to repeat the type of `self` in every
```
wraitii, line 82: I don't understand the second bit of this sentence, in Cairo 1, you have to.

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,309 @@
...
+method’s signature, is for organization. We’ve put all the things we can do
+with an instance of a type in one combination of `trait` & `impl` blocks, rather than making future users
+of our code search for capabilities of `Rectangle` in various places in the
+library we provide.
```
wraitii, line 86: I also don't think Cairo provides this guarantee? You in fact mention below that you can add more methods and traits to a type ?

src/ch04-03-method-syntax.md
```diff
@@ -0,0 +1,309 @@
...
+could choose to provide an associated function named `square` that would have
+one dimension parameter and use that as both width and height, thus making it
+easier to create a square `Rectangle` rather than having to specify the same
+value twice:
+
+<span class="filename">Filename: src/lib.cairo</span>
+
+```rust
+trait RectangleTrait {
+    fn square(size:u64) -> Rectangle;
```
wraitii, line 244: Would be nice to have one without parameters, because I believe this technically defines a method _square_ on `u64` that returns a Rectangle...

In fact I think this might be a cairo 1 anti pattern, you want your 'constructors' to take no parameters, which is a _mess_.

wraitii (commented):
Seems much clearer to me but there are still some edge-cases / oddities that might be worth addressing

src/ch04-03-method-syntax.md
enitrat: Traits functions that don't use `self` are not methods. Addition traits are special because they are not called on `a` or `b`, so `add`. 
On the other hand, `add_eq` * is * a method called on a `self` parameter and compared to a second value

src/ch04-03-method-syntax.md
enitrat: I might add a disclaimer that it's bad practice to do so then :)

src/ch04-03-method-syntax.md
enitrat: Thanks, it slipped through.

src/ch04-03-method-syntax.md
enitrat: Yes you can, but in this specific case we don't and grouped everything in one place. I added a sentence about this.

src/ch04-03-method-syntax.md
enitrat: It initially was `struct`, but that's too restrictive, since methods can be defined for struct, enums, and corelib types as well... idk

wraitii: I believe the ` types` here refers to rust? I'm not sure if you can `imply  a rust enum

src/ch04-03-method-syntax.md
wraitii: You are correct, but this actually only applies to the `trait` definition, not to `impl` where you can rename it however you want

src/ch04-03-method-syntax.md
enitrat: > While it is technically
possible to define methods for multiple types within the same `impl` block, it is not
a recommended practice, as it can lead to confusion.

wraitii: I think it might be worth to just skip that sentence here ? Maybe just go deeper in the topics in traits ?

wraitii (approved):

Issue #26, opened 2023-04-05
Defining an Enum


enitrat:
- Enum variants, values
- enums combined with custom types
- traits implementations for enums
- The Option Enum and its advantages

Th0rgal:
I can work on this :)

PR #32, opened 2023-04-08
Feat/enums

Th0rgal:
Hey,

Here is my page about Enums (see #26 ).  I tried to give working examples that make sense. The last one is a bit too long in my opinion but hopefully we should be able to make it shorter when we will have working loops with automatic gas management (may be we already do and I missed it). 

What do you think? :)


src/ch05-01-enums.md
```diff
@@ -0,0 +1,141 @@
...
+
+    if index >= arr.len() {
+        return Option::None(());
+    }
+
+    if *arr.at(index) == value {
+        return Option::Some(index);
+    }
+
+    find_value_recursive(arr, value, index + 1_usize)
```
fricoben, line 113: A bit difficult considering that we shouldn't see a lot of recursive loops anymore.

src/ch05-01-enums.md
```diff
@@ -0,0 +1,141 @@
...
+            if index == 1_usize {
+                // it worked should get printed
+                'it worked'.print();
+            }
+        },
+        Option::None(()) => {
+            'not found'.print();
+        },
+    }
+}
```
fricoben, line 139: Maybe show what's printed when your run ?

fricoben (changes requested):
Some questions and propositions

src/ch05-01-enums.md
```diff
@@ -0,0 +1,141 @@
...
+        match self {
+            Animal::Elephant((weight, _)) => {
+                weight
+            },
+            Animal::Ant(()) => {
+                0_u128
+            },
+            Animal::Giraffe((weight, _, _)) => {
+                weight
+            },
```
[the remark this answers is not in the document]
Th0rgal: exactly, I think it should be explained better in the iris docs about match

src/ch05-01-enums.md
Th0rgal: Right, as explained in the note this is temporary, we can update the example when loops are available

src/ch05-01-enums.md
Th0rgal: Alright, I will remove it from the source code comment and add it at the end, "run output" or something

src/ch05-01-enums.md
```diff
@@ -0,0 +1,141 @@
...
+Enums can also be used to store more interesting data associated with each variant. For example:
+```rs
+enum Animal {
+    Elephant : (u128, u128),
+    Ant : (),
+    Giraffe : (u128, u128, u128),
+}
+```
+
+In this example, the Animal enum has three variants: `Elephant`, `Ant` and `Giraffe`.
```
hurrikaanig, line 38: Forgot quotes on Animal

src/ch05-01-enums.md
```diff
@@ -0,0 +1,143 @@
...
+
+```rs
+trait AnimalWeight {
+    fn get_weight(self: Animal) -> u128;
+}
+
+impl AnimalWeightImpl of AnimalWeight {
+    fn get_weight(self: Animal) -> u128 {
+        match self {
+            Animal::Elephant((weight, _)) => {
```
enitrat, line 55: might be worth explaining what `_` stands for

src/ch05-01-enums.md
```diff
@@ -0,0 +1,143 @@
...
+let direction = Direction::North(());
+```
+
+It's easy to write code that acts differently depending on the variant of an enum instance, in this example to run specific code according to a Direction. You can learn more about it on the [The Match Control Flow Construct page](ch05-02-the-match-control-flow-construct.md).
+
+## Enums Combined with Custom Types
+Enums can also be used to store more interesting data associated with each variant. For example:
+
+```rs
+enum Animal {
```
enitrat, line 31: I personally think this example doesn't demonstrate the power of enums very well and is quite confusing.
It would work better with something more constrained. I actually like the rustbook example with Messages more, because 
- It implies that you can only have 4 types of messages in your application and makes it more obvious
- Each message has a different function and must then be treated differently (which is not clear with animals)
- Each enum variant value is a different type, which explains well what you're trying to demonstrate here
- The `message.call` example (which I used in Starklings) is I think very explicit

I'll leave it up to you to decide if you want to change it :)

fricoben:
close #32

src/ch05-01-enums.md
Th0rgal: I agree, tbh I couldn't find one example that worked very well for both the `Custom Types` and the `Traits implementation` sections. Do you have an idea in mind of a trait that would make sense for the Messages enum?

src/ch05-01-enums.md
Th0rgal: will do

src/ch05-01-enums.md
enitrat: Why not a simple `message.process` that acts differently based on the variant received?

Th0rgal: let's do it

Th0rgal:
should be good now @enitrat

enitrat (approved):
Great work :100:

PR #52, opened 2023-04-14
feat: testing cairo programs

### LINKED ISSUE #38: Testing Cairo programs
This will chapter cover testing **pure Cairo** programs. 
- Test modules ( `mod tests {...}` ) should be annotated with the `#[cfg(tests)]` attribute to reduce the size of the compiled code
- How to check values with the `assert` function
- How to expect failure from a test (with/without an expected error message)
- Differentiating unit tests and integration tests

Note: Unlike Rust, unit tests should not be with the source code.
Estimated time: 4 hours

irisdv:
Closes #38

### DIFF
--- a/src/ch09-01-how-to-write-tests.md
+++ b/src/ch09-01-how-to-write-tests.md
@@ -8,25 +8,39 @@ Tests are Cairo functions that verify that the non-test code is functioning in t
 - Run the code you want to test.
 - Assert the results are what you expect.
 
-Let’s look at the features Cairo provides specifically for writing tests that take these actions, which include the test attribute, the assert function, and and the should_panic attribute.
+Let’s look at the features Cairo provides specifically for writing tests that take these actions, which include the `test` attribute, the `assert` function, and and the `should_panic` attribute.
 
 ### The Anatomy of a Test Function
 
 At its simplest, a test in Cairo is a function that’s annotated with the `test` attribute. Attributes are metadata about pieces of Cairo code; one example is the derive attribute we used with structs in Chapter 4. To change a function into a test function, add `#[test]` on the line before `fn`. When you run your tests with the `cairo-test` command, Cairo builds a test runner binary that runs the annotated functions and reports on whether each test function passes or fails.
 
-Let's create a new project called `adder` that will add two numbers:
+Let's create a new project called `adder` that will add two numbers using Scarb with the command `scarb new adder`:
 
 ```shell
 adder
 ├── cairo_project.toml
+├── Scarb.toml
 └── src
-    ├── cairo_project.toml
-    └── main.cairo
+    └── lib.cairo
 ```
 
-In `main.cairo`, let's add a firt test, as showned in Listing 9-1.
+<!-- TODO: remove when Scarb test work -->
 
-<span class="filename">Filename: main.cairo</span>
+> Note: You will notice here a `cairo_project.toml` file.
+> This is the configuration file for "vanilla" Cairo projects (i.e. not managed by Scarb),
+> which is required to run the `cairo-test .` command to run the code of the crate.
+> It is required until Scarb implements this feature. The content of the file is:
+>
+> ```toml
+> [crate_roots]
+> adder = "src"
+> ```
+>
+> and indicates that the crate named "adder" is located in the `src` directory.
+
+In _lib.cairo_, let's add a firt test, as shown in Listing 9-1.
+
+<span class="filename">Filename: lib.cairo</span>
 
 ```rust
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


src/ch09-01-how-to-write-tests.md
```diff
@@ -0,0 +1,503 @@
...
+    fn add_three_and_two() {
+        let result = 3 + 2;
+        assert(result == 5, 'result is not 5');
+    }
+}
+```
+
+Listing 9-10: Two tests with two different names
+
+We can pass the name of any test function to `cairo-test` to run only that test using the `-f` flag:
```
enitrat, line 453: I think it's worth mentioning that it filters tests that **contain** the current string

src/ch09-02-test-organization.md
```diff
@@ -0,0 +1,91 @@
+# Testing Organization
+
+We'll think about tests in terms of two main categories: unit tests and integration tests. Unit tests are small and more focused, testing one module in isolation at a time, and can test private functions. Integration tests are entirely external to your library and use your code in the same way any other external code would, using only the public interface and potentially exercising multiple modules per test.
+
+Writing both kinds of tests is important to ensure that the pieces of your library are doing what you expect them to, separately and together.
+
+## Unit Tests
+
+The purpose of unit tests is to test each unit of code in isolation from the rest of the code to quickly pinpoint where code is and isn’t working as expected. You’ll put unit tests in the `src` directory in each file with the code that they’re testing.
```
enitrat, line 9: This is worth discussing but I think we shouldn't actually do this.

The code might end up on block explorers for verification purposes. Even for libraries I think we should keep a clear separation between source files and tests files. Plus, given that Cairo doesn't have that visibility problem (pub/private), we don't really need to have test modules with the sources.

IMO we have a `src/` and `tests/` folders, and unit & integration tests are in the `tests/` folder

@Th0rgal @wraitii let's debate this :)

src/ch09-02-test-organization.md
```diff
@@ -0,0 +1,91 @@
...
+        assert(result == 4, 'result is not 4');
+    }
+}
+```
+
+The attribute `cfg` stands for configuration and tells Cairo that the following item should only be included given a certain configuration option. In this case, the configuration option is `test`, which is provided by Cairo for compiling and running tests. By using the `cfg` attribute, Cairo compiles our test code only if we actively run the tests with `cairo-test`. This includes any helper functions that might be within this module, in addition to the functions annotated with `#[test]`.
+
+## Integration Tests
+
+In Cairo, integration tests are entirely external to your library. They use your library in the same way any other code would. Their purpose is to test whether many parts of your library work together correctly. Units of code that work correctly on their own could have problems when integrated, so test coverage of the integrated code is important as well. To create integration tests, you first need a `tests` directory.
```
enitrat, line 36: > are entirely external to your library.

Hm, I wouldn't word it this way, since visibility doesn't exist in Cairo, I think you can remove this sentence

src/ch09-02-test-organization.md
```diff
@@ -0,0 +1,91 @@
...
+└── tests
+    ├── cairo_project.toml
+    ├── lib.cairo
+    └── integration_test.cairo
+```
+
+In your `tests` directory, create a `cairo_project.toml` file. This file will tell Cairo where our `adder` module is located and that this directory is the `tests` module:
+
+```rust
+[crate_roots]
```
enitrat, line 56: we should have a single `cairo_project.toml` at the project root that declares multiple crates

```toml
[crate_roots]
adder = "src"
tests = "tests"
```

instead

src/ch09-02-test-organization.md
```diff
@@ -0,0 +1,91 @@
...
+
+#[test]
+fn internal() {
+    assert(main::internal_adder(2_u32, 2_u32) == 4_u32, 'internal_adder failed');
+}
+```
+
+Each file in the tests directory is a separate crate, so we need to bring our library into each test crate’s scope. For that reason we add `use adder::main` at the top of the code, which we didn’t need in the unit tests.
+
+We don’t need to annotate any code in _tests/integration_test.cairo_ with `#[cfg(test)]`. Cairo treats the tests directory specially and compiles files in this directory only when we run `cairo-test`. Run `cairo-test` now:
```
enitrat, line 80: hmmm I am not sure, let's keep it for now.

In `lib.cairo`

```rust
#[cfg(tests)]
mod integration_tests
```
mod tests

enitrat (commented):
Amazing PR, tyvm, I left some comments about
- Alpha.7 changes
- Things we need to discuss
- Small inaccuracies

🔥

src/ch09-02-test-organization.md
wraitii: I don't really mind either way TBH. I think putting some level of unit tests next to the source code is convenient, but it's more convenient to _write_ the test than once they're actually _written_.

enitrat: > but it's more convenient to write the test than once they're actually written.

I didn't get that

enitrat (approved):
LGTM :fire:

src/ch09-02-test-organization.md
wraitii: What I meant is that being able to write the tests in the same file is very good for workflow: it minimises context-switching. You write the function, the text below, ideally you can see both at once, and adjust as needed.
Then once the test is written, it could be moved somewhere else, as it's now a regression-test.

PR #53, opened 2023-04-14
Traits

### LINKED ISSUE #37: Cairo's trait system
- How to define a trait
- How to implement a trait on a type
- Managing different implementations for the same trait (e.g. you can have one impl in `mod a` and another one in `mod b`. You will need to import the correct one to use the method syntax on your type)

Estimated time: 2 hours

shramee:
Fixes #37

### DIFF
--- a/src/ch04-04-traits-in-cairo.md
+++ b/src/ch04-04-traits-in-cairo.md
@@ -1,0 +1,98 @@
+# Traits in Cairo
+
+Traits specify functionality blueprints that can be implemented. The specification includes a set of function signatures containing type annotations for parameters and return value.
+
+## Defining a Trait
+
+To define a trait, you use the keyword `trait` followed by the name of the trait in `PascalCase` then the function signatures in a pair of curly braces.
+
+Here's what a simple trait declaration looks like.
+
+```rust
+trait ShapeGeometry {
+	fn boundary( self: Rectangle ) -> u64;
+	fn area( self: Rectangle ) -> u64;
+}
+```
+
+## Implementing a Trait
+
+The trait can then be implemented with `impl` keyword with the name of the implementation (to refer to when using methods/functions) like this,
+
+```rust
+impl RectangleGeometry of ShapeGeometry {
+	fn boundary( self: Rectangle ) -> u64 {
+        2_u64 * (self.height + self.width)
+    }
+	fn area( self: Rectangle ) -> u64 {
+		self.height * self.width
+	}
+}
+```
+
+## Parameter `self`
+
+In the example above, `self` is a special keyword. When parameter with name `self` is used, the implemented functions are also attached to the instances of the type as methods. Here's an illustration,
+
+When `ShapeGeometry` trait is implemented,
+Function `area` from `ShapeGeometry` trait can be called in two ways,
+
+```rust
+let rect = Rectangle { ... }; // Rectangle instantiation
+
+// First way, as a method on the struct instance
+let area1 = rect.area();
+// Second way, from the implementation
+let area2 = RectangleGeometry::area(rect);
+// `area1` has same value as `area2`
+area1.print();
+area2.print();
+```
+
+And the implementation of `area` method will be able to access the instance via `self` parameter.
+
+## Traits with generic types
+
+You would want to write a trait when you want multiple types to implement some functionality in a standard way. To do this, we use generic types.
+
+In the example below, we use generic type `T` and our method signatures can use this alias which can be provided during implementation.
+
+```rust
+// Here T is an alias type which will be provided buring implementation
+trait ShapeGeometry<T> {
+	fn boundary( self: T ) -> u64;
+	fn area( self: T ) -> u64;
+}
+
+// Implementation RectangleGeometry passes in <Rectangle>
+// to implement the trait for that type
+impl RectangleGeometry of ShapeGeometry::<Rectangle> {
+	fn boundary( self: Rectangle ) -> u64 {
+        2_u64 * (self.height + self.width)
+    }
+	fn area( self: Rectangle ) -> u64 {
+		self.height * self.width
+	}
+}
+
+// We might have another struct Triangle
+// which can use the same trait spec
+impl CircleGeometry of ShapeGeometry::<Circle> {
+	fn boundary( self: Circle ) -> u64 {
+        (2_u64 * 314_u64 * self.radius) / 100_u64
+    }
+	fn area( self: Circle ) -> u64 {
+		(314_u64 * self.radius * self.radius) / 100_u64
+	}
+}
+
+fn main() {
+	let rect = Rectangle { height: 5_u128, width: 7_u128 };
+    rect.area().print(); // 35
+    rect.boundary().print(); // 24
+
+	let circ = Circle { radius: 5_u128 };
+    circ.area().print(); // 78
+    circ.boundary().print(); // 31
+}
+```
--- a/src/ch04-04-traits-in-cairo.md
+++ b/src/ch04-04-traits-in-cairo.md
@@ -1,12 +1,12 @@
 # Traits in Cairo
 
-Traits specify functionality blueprints that can be implemented. The specification includes a set of function signatures containing type annotations for parameters and return value.
+Traits specify functionality blueprints that can be implemented. The blueprint specification includes a set of function signatures containing type annotations for the parameters and return value. This sets a standard to implement the specific functionality.
 
 ## Defining a Trait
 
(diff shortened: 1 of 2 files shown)


shramee:
Sure can! Thanks for taking a look 🤗

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
+# Traits in Cairo
+
+Traits specify functionality blueprints that can be implemented. The specification includes a set of function signatures containing type annotations for parameters and return value.
```
enitrat, line 3: Can you be more specific? It's not very clear to someone that is not familiar with traits. What can you do with them?

From the rust book
> A trait defines functionality a particular type has and can share with other types. We can use traits to define shared behavior in an abstract way.

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Defining a Trait
+
```
enitrat, line 6: Can you add an explanation here? What do you define a `ShapeGeometry` trait for? 
It must be very explicit here, you must mention that all shapes share common methods such as areas, etc, and defining a geometry trait allows us to specify a shared behavior for all shapes

Once again quoting the Rust book for this

> A type’s behavior consists of the methods we can call on that type. Different types share the same behavior if we can call the same methods on all of those types. Trait definitions are a way to group method signatures together to define a set of behaviors necessary to accomplish some purpose.
...

+ Suggestion
> For example, let's say that we have multiple structs representing shapes. We want our application to be able to perform geometry operations on these shapes ...

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
...
+Here's what a simple trait declaration looks like.
+
+```rust
+trait ShapeGeometry {
+	fn boundary( self: Rectangle ) -> u64;
+	fn area( self: Rectangle ) -> u64;
+}
+```
+
```
enitrat, line 17: Can you add some explanation here?

> Here, we declare a trait using the trait keyword and then the trait’s name, which is ShapeGeometry in this case. Inside the curly brackets, we declare the method signatures that describe the behaviors of the type that implement this trait, which in this case is...

- traits can have multiple functions
-  instead of providing an implementation within curly brackets, we use a semicolon at the end of the declaration

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
...
+## Implementing a Trait
+
+The trait can then be implemented with `impl` keyword with the name of the implementation (to refer to when using methods/functions) like this,
+
```
enitrat, line 21: Be more explicit. You must detail what happens in your code snippet.

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
...
+    }
+	fn area( self: Rectangle ) -> u64 {
+		self.height * self.width
+	}
+}
+```
+
+## Parameter `self`
+
+In the example above, `self` is a special keyword. When parameter with name `self` is used, the implemented functions are also attached to the instances of the type as methods. Here's an illustration,
```
enitrat, line 35: When a parameter with the name

enitrat: Can you insert a link to ch04-03-method-syntax.html#defining-methods?

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
...
+let area2 = RectangleGeometry::area(rect);
+// `area1` has same value as `area2`
+area1.print();
+area2.print();
+```
+
+And the implementation of `area` method will be able to access the instance via `self` parameter.
+
+## Traits with generic types
+
```
enitrat, line 55: Refering to my first comment:

> A trait defines functionality a particular type has and can share with other types. We can use traits to define shared behavior in an abstract way.

I think we should emphasize that non-generic traits are implemented for a single type, and that to define shared behavior common to multiple types we must use a generic trait syntax.

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
...
+	let circ = Circle { radius: 5_u128 };
+    circ.area().print(); // 78
+    circ.boundary().print(); // 31
+}
+```
+
+## Managing and using external trait implementations
+
+To use traits methods, you need to make sure the correct implementation(s) are imported. In the code above we imported `PrintTrait` from `debug` with `use debug::PrintTrait;` to use `print()` methods.
```
enitrat, line 104: Can you add a sentence that explains that when importing a trait, the compiler will look for implementation in the same modules as the trait, and that you don't need to explicitly import those?

src/ch04-04-traits-in-cairo.md
```diff
@@ -0,0 +1,156 @@
...
+
+In some cases you might need to import not only the trait but also the implementation if they are declared in separate modules.
+If `CircleGeometry` was in a separate module/file `circle` then to use `boundary` on `circ: Circle`, we'd need to import `CircleGeometry` in addition to `ShapeGeometry`.
+
+If the code was organised into modules like this,
+
+```rust
```
enitrat, line 111: Great!

enitrat (changes requested):
Thank you for the PR. I think we need to be more explicit and descriptive, as this book must remain understandable by people with only very basic programming knowledge.

I left comments that need to be addressed

src/ch04-04-traits-in-cairo.md
shramee: @enitrat Cairo traits IMO are significantly different from Rust traits. To me, they don't seem to be coupled with  `types` in Cairo.

enitrat: Yes, however this becomes true with generic traits.

src/ch04-04-traits-in-cairo.md
shramee: We already have this,
```
In some cases you might need to import not only the trait but also the implementation if they are declared in separate modules.
```

src/ch04-04-traits-in-cairo.md
shramee: > `Here's what a simple trait declaration looks like.`

Swapped with,

> For example, let's say that we have multiple structs representing shapes. We want our application to be able to perform geometry operations on these shapes, So we define a trait `ShapeGeometry` that contains a blueprint to implement geometry operations on the shape like this,

src/ch04-04-traits-in-cairo.md
shramee: Added this,

> Here our trait `ShapeGeometry` declares signatures for two methods `boundary` and `area`. When implemented, both these functions should return a `u64` and except parameters as specified by the trait.

src/ch04-04-traits-in-cairo.md
shramee: Added,

> In the code above, `RectangleGeometry` implements the trait `ShapeGeometry` defining what the methods `boundary` and `area` should do. Note the the function parameters and return value types are identical to the trait specification.

src/ch04-04-traits-in-cairo.md
shramee:
```diff
-In the example above, `self` is a special keyword. When parameter with name `self` is used, the implemented functions are also attached to the instances of the type as methods. Here's an illustration,
+In the example above, `self` is a special keyword. When a parameter with name `self` is used, the implemented functions are also [attached to the instances of the type as methods](ch04-03-method-syntax.md#defining-methods). Here's an illustration,
```

src/ch04-04-traits-in-cairo.md
shramee: Replaced,

>You would want to write a trait when you want multiple types to implement some functionality in a standard way. To do this, we use generic types.

with

> Usually we want to write a trait when we want multiple types to implement a functionality in a standard way. However, in the example above the signatures are static and cannot be used for multiple types. To do this, we use generic types when defining traits.

src/ch04-04-traits-in-cairo.md
shramee: > Traits specify functionality blueprints that can be implemented. The specification includes a set of function signatures containing type annotations for parameters and return value.

swapped with,

> Traits specify functionality blueprints that can be implemented. The blueprint specification includes a set of function signatures containing type annotations for parameters and return value. This helps standardise the way to implement some specific functionality.

shramee: Just made it more verbose, let me know if this makes more sense.

enitrat (approved):
LGTM :)

PR #54, opened 2023-04-15
Feat/generic data types

### LINKED ISSUE #36: Generic Data Types
This part should introduce how to use Generic data types in Cairo, including:
- How to define generic functions
- How to define generic function, while restricting the T type that can be used with this function
- How to define generic structs
- How to define generic enums
- How to define generic methods (generic trait + generic impl on a type T)

Estimated time: 4 hours

rodrigo-pino:
fix #36


enitrat:
Can you merge 01-02-03 in a single page please?

src/ch07-00-generic-types.md
```diff
@@ -0,0 +1,9 @@
+# Generic Types
+
+Every programming language has tools for effectively handling the duplication of concepts. In Cairo, one such tool is generics: abstract stand-ins for concrete types or other properties. We can express the behaviour of generics or how they relate to other generics without knowing what will be in their place when compiling and running the code.
+
+Functions, structs, enums and traits can incorporate generic types as part of their definition instead of a concrete types like `u32` or `ContractAddress`.
+
+Generics allow us to replace specific types with a placeholder that represents multiple types to remove code duplication.
+
+For each concrete type that replaces a generic type the compiler creates a new definition, reducing development time for the programmer, but code duplication at compile level still exists. This may be of importance if you are writing Starknet contracts and using a generic for multiple types which will cause contract size to increment.
```
enitrat, line 9: nice!

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+    l2.append(4);
+    l2.append(5);
+
+    // There is no need to specify the concrete type of T because
+    // it is inferred by the compiler
+    let l3 = largest_list(l1, l2);
+}
+```
+
+The `largest_list` function compares two lists of the same type and returns the one with more elements and drops the other. If you compile the previous code, you will notice that it will fail with an error saying that there are no traits defined for droping an array of a generic type. This happens because the compiler has now way to guarantee that an `Array<T>` is droppable when executing the `main` function. In order to drop an array of `T`, the compiler must first know how to drop `T`. This can be fixed by specifiying in the function signature that `T` implements the drop trait.
```
enitrat, line 40: > has now way 

has no way

enitrat: >  that T implements

maybe we can rephrase to: that T must implement

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+
+```rust
+use array::ArrayTrait;
+
+// Specify generic type T between the angulars
+// Also specify that the T implements the Drop trait
+fn largest_list<T, impl TDrop: Drop<T>>(l1: Array<T>, l2: Array<T>) -> Array<T> {
```
enitrat, line 47: Instead of rewriting the whole example, can you include only the line to be changed? That way it would be more obvious that nothing else has changed.

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+    // TDrop because they are infered by the compiler
+    let l3 = largest_list(l1, l2);
+}
+```
+
+The new `largest_list` function includes in its definition the requirement that whatever generic type is placed there, it must be droppable. Note that the `main` function remained unchanged, the compiler is smart enough to deduct which concrete type is being used and if it implements the `Drop` trait.
+
+### Constraints for Generic Types
+
+When defining generic types, it is useful to have information about them. Knowing which traits a generic type implements allow us to utilize them more effectively in a functions logic at the cost of constraining the generic types that can be used with the function. We saw an example of this previously by adding the `TDrop` implementation as part of the generic arguments of `largest_list`. While `TDrop` was added to satisfy the compilers requirements, we can also add constraints to benefit our function logic.
```
enitrat, line 76: > utilize

use
idk why but chatGPT loves the word `utilize` lol

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+}
+
+impl WalletDrop<T, impl TDrop : Drop<T>> of Drop<Wallet<t>>;
+
+fn main() {
+   let w = Wallet{ balance: 3_u128};
+}
+```
+
+We avoid using the `derive` macro for `Drop` implementation of `Wallet` and instead define our own `WalletDrop` implementation. Notice that we must define, just like functions, an aditional generic type for `WalletDrop` saying that `T` implements the `Drop` trait as well. We are basically saying that the struct `Wallet<T>` is droppable as long as `T` is also droppable.
```
enitrat, line 171: > aditional

additional

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+
+Finally, if we want to add a field to `Wallet` representing its Cairo address and we want that field to be different than `T` but generic as well can simply add another generic type between the `<>`:
```
enitrat, line 173: nice!

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+    assert(w.balance() == 50, 0);
+}
+```
+
+We first define `WalletTrait<T>` trait using a generic type `T` which defines a method that returns a snapshot of the field `address` from `Wallet`. Then we give an implementation for the trait in `WalletImpl<T>`. Note that you need to include a generic type in both definitions of the trait and the implementation.
+
+We can also specify constraints on generic types when defining methods on the type. We could, for example, implement methods only for `Wallet<u128>` instances rather than `Wallet<T>`. In the code example we define an implementation for wallets which have a concrete type of `u128` for the `balance` field.
+
+```rust
+trait WalletRecieveTrait {
```
enitrat, line 248: WalletReceiveTrait

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+    fn recieve(ref self: Wallet<u128>, value: u128);
```
enitrat, line 249: receive

src/ch07-01-generic-data-types.md
```diff
@@ -0,0 +1,321 @@
...
+}
+
+impl WalletRecieveImpl of WalletRecieveTrait {
```
enitrat, line 252: WalletReceiveImpl

enitrat (approved):
This is an insanely well written PR that covers everything there is to know about genericity in Cairo, tyvm!!! 💯 

I left some very minor comments, once addressed we can merge 🔥

src/ch07-01-generic-data-types.md
rodrigo-pino: Yeah lol, I had "use" initially 😅

src/ch07-01-generic-data-types.md
rodrigo-pino: Removed everything else but the updated function definition

Issue #47, opened 2023-04-13
Cairo Projects: Bringing Paths into Scope with the use Keyword


enitrat:
- Full paths and shortcuts to paths (e.g. import a module and then call `module::foo`)
- Idiomatic paths (full paths for Traits, Structs, Impls, Enums)
-Using external packages (with Scarb)
- Re-export names in module files (in module a you can import foo with `use a::b::c::foo`, so that you can now import foo in other places using only `use a::foo`

Estimated time: 2 hours

PR #55, opened 2023-04-15
feat: add paths explainations

### LINKED ISSUE #46: Cairo Projects: Paths to refer to an item
- Absolute paths
- Relative paths
Reference: https://github.com/starkware-libs/cairo/blob/main/docs/reference/src/components/cairo/modules/language_constructs/pages/path.adoc
Estimated time: 1 hour

fricoben:
close #46 

I tried to show the path possibilities that I know, if you know more of them feel free to tell me I'll add them :)

### DIFF
--- a/src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
+++ b/src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
@@ -1,0 +1,163 @@
+## Paths for Referring to an Item in the Module Tree
+
+To show Cairo where to find an item in a module tree, we use a path in the same way we use a path when navigating a filesystem. To call a function, we need to know its path.
+
+A path can take two forms:
+
+- An _absolute path_ is a full path starting from a crate root, in Rust you use the `crate` keyword to start an absolute path, in Cairo absolute paths are not implemented yet.
+- A _relative path_ starts depends on the current module and uses `super`, or an identifier in the current module.
+
+Both absolute and relative paths are followed by one or more identifiers
+separated by double colons (`::`).
+
+To illustrate this notion let's take back our example for the restaurant we used in the last chapter. We have a module named `front_of_house` that contains a module named `hosting`. The `hosting` module contains a function named `add_to_waitlist`. We want to call the `add_to_waitlist` function from the `eat_at_restaurant` function. We need to tell Cairo the path to the `add_to_waitlist` function so it can find it.
+
+<span class="filename">Filename: src/lib.cairo</span>
+
+```rust
+mod front_of_house {
+    mod hosting {
+        fn add_to_waitlist() {}
+
+        fn seat_at_table() {}
+    }
+
+    mod serving {
+        fn take_order() {}
+
+        fn serve_order() {}
+
+        fn take_payment() {}
+    }
+}
+
+use front_of_house::hosting::add_to_waitlist; // used for Relative Method 2
+
+pub fn eat_at_restaurant() {
+    // Absolute path
+    crate::front_of_house::hosting::add_to_waitlist(); // ❌ Doesn't compile in Cairo
+
+    // Relative path Method 1
+    front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+
+    // Relative path Method 2
+    add_to_waitlist(); // ✅ Compiles
+
+    // Relative path Method 3
+    super::lib::front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+
+}
+```
+
+The first time we call the `add_to_waitlist` function in `eat_at_restaurant`,
+we use an absolute path. The `add_to_waitlist` function is defined in the same
+crate as `eat_at_restaurant`. In rust you can use the `crate` keyword to
+start an absolute path, in Cairo it's not possible that's the reason why this code won't compile.
+
+We then include three methods to import with a relative path.
+
+For the first method, the path starts with `front_of_house`, the name of the module
+defined at the same level of the module tree as `eat_at_restaurant`. Here the
+filesystem equivalent would be using the path
+`./front_of_house/hosting/add_to_waitlist`. Starting with a module name means
+that the path is relative to the current module.
+
+The second method is similar to the first one but we use the `use` keyword to import the function. This is the recommended way to import functions as it permits to use the function without having to specify the path.
+
+The third method is similar to the first one but we use the `super` keyword to start the path from the parent module. Here the filesystem equivalent would be using the path `../front_of_house/hosting/add_to_waitlist`.
+
+### Starting Relative Paths with `super`
+
+Choosing whether to use a `super` or not is a decision you’ll make
+based on your project, and depends on whether you’re more likely to move item
+definition code separately from or together with the code that uses the item.
+For example, if we move the `front_of_house` module and the `eat_at_restaurant`
+function into a module named `customer_experience` like so:
+
+<span class="filename">Filename: src/lib.cairo</span>
+
+```rust
+mod customer_experience {
+    mod front_of_house {
+        mod hosting {
+            fn add_to_waitlist() {}
+
+            fn seat_at_table() {}
+        }
+
+        mod serving {
+            fn take_order() {}
+
+            fn serve_order() {}
+
+            fn take_payment() {}
+        }
+    }
+
+    fn eat_at_restaurant() {
+        front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+        super::lib::front_of_house::hosting::add_to_waitlist(); // ❌ Doesn't compile
+        super::customer_experience::front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+    }
+}
+```
+
+We need to update the path that uses `super` because it won't compile as we added the `customer_experience` module. Now super is relative to the `customer_experience` module and not the `lib` file.
+
+Here is the error message:
+
+```bash
+error: Invalid path.
+ --> lib.cairo:20:22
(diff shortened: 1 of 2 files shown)


src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,163 @@
+## Paths for Referring to an Item in the Module Tree
+
+To show Cairo where to find an item in a module tree, we use a path in the same way we use a path when navigating a filesystem. To call a function, we need to know its path.
+
+A path can take two forms:
+
+- An _absolute path_ is a full path starting from a crate root, in Rust you use the `crate` keyword to start an absolute path, in Cairo absolute paths are not implemented yet.
```
enitrat, line 7: I would remove the comparison with Rust here - absolute paths start with the crate name

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,163 @@
...
+
+        fn take_payment() {}
+    }
+}
+
+use front_of_house::hosting::add_to_waitlist; // used for Relative Method 2
+
+pub fn eat_at_restaurant() {
+    // Absolute path
+    crate::front_of_house::hosting::add_to_waitlist(); // ❌ Doesn't compile in Cairo
```
enitrat, line 38: I would remove this as well

enitrat: absolute path should be something like `restaurant::front_of_house::...` since our crate is named "restaurant", that's what should be shown here

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,163 @@
...
+
+    // Relative path Method 3
+    super::lib::front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+
+}
+```
+
+The first time we call the `add_to_waitlist` function in `eat_at_restaurant`,
+we use an absolute path. The `add_to_waitlist` function is defined in the same
+crate as `eat_at_restaurant`. In rust you can use the `crate` keyword to
```
enitrat, line 54: same, remove mention to rust, add a mention that you must use the crate name

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,163 @@
...
+    mod serving {
+        fn take_order() {}
+
+        fn serve_order() {}
+
+        fn take_payment() {}
+    }
+}
+
+use front_of_house::hosting::add_to_waitlist; // used for Relative Method 2
```
enitrat, line 34: I would say that the `use` keyword to bring items into scope should not be explained here but in #47

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,163 @@
...
+
+            fn serve_order() {}
+
+            fn take_payment() {}
+        }
+    }
+
+    fn eat_at_restaurant() {
+        front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+        super::lib::front_of_house::hosting::add_to_waitlist(); // ❌ Doesn't compile
```
enitrat, line 99: remove this one as well

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,163 @@
...
+        super::customer_experience::front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+    }
+}
+```
+
+We need to update the path that uses `super` because it won't compile as we added the `customer_experience` module. Now super is relative to the `customer_experience` module and not the `lib` file.
+
+Here is the error message:
+
+```bash
```
enitrat, line 109:
```
```console
```

enitrat (changes requested):
Thanks for the PR, however I would change some stuff
- Remove mentions to Rust's `crate::`, instead mention that absolute paths use the crate's name as the root module
- Remove anything that has to do with `super::lib`, this is not intuitive. Examples with super should refer to an explicit parent module directly

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
fricoben: On my end doing something like this in lib.cairo

```rust 
mod customer_experience {
    mod front_of_house {
        mod hosting {
            fn add_to_waitlist() {}

            fn seat_at_table() {}
        }

        mod serving {
            fn take_order() {}

            fn serve_order() {}

            fn take_payment() {}
        }
    }

    fn eat_at_restaurant() {
        front_of_house::hosting::add_to_waitlist();
        super::lib::front_of_house::hosting::add_to_waitlist();
        lib::customer_experience::front_of_house::hosting::add_to_waitlist();
    }
}
```

the `lib::customer_experience::front_of_house::hosting::add_to_waitlist();` does not compile so we can't say there is an absolute path here. That's why I said there is no absolute path I thought would come in the future.

enitrat: yeah but the crate name here is not `lib`, it's `restaurant`. It's the one defined in `cairo_project.toml`

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
fricoben: https://github.com/cairo-book/cairo-book.github.io/pull/55/files/14f85aab09b2a502d5692c37c5062ab978869d91#r1167565620

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
fricoben: [This does not compile for me as I said here](https://github.com/cairo-book/cairo-book.github.io/pull/55/files/14f85aab09b2a502d5692c37c5062ab978869d91#r1167565620)

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
fricoben: Understood here I didn't set the restaurant module well

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,153 @@
...
+    front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+
+    // Relative path Method 2
+    super::lib::front_of_house::hosting::add_to_waitlist(); // ✅ Compiles
+}
+```
+
+The first time we call the `add_to_waitlist` function in `eat_at_restaurant`,
+we use an absolute path. The `add_to_waitlist` function is defined in the same
+crate as `eat_at_restaurant`. In Cairo it's not possible you need to specify the crate module identifier,
```
enitrat, line 49: Something along the lines of:
"In Cairo, absolute paths start from the crate root, which you need to refer to by using the crate name."

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,153 @@
...
+in our example it's `restaurant`.
+
+We then included two methods to import `add_to_waitlist` with a relative path.
```
enitrat, line 52: two "ways" instead of methods maybe?

enitrat: We then included two ways of using `add_to_waitlist` with a relative path.

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,153 @@
...
+Choosing whether to use a `super` or not is a decision you’ll make
+based on your project, and depends on whether you’re more likely to move item
+definition code separately from or together with the code that uses the item.
+For example, if we move the `front_of_house` module and the `eat_at_restaurant`
+function into a module named `customer_experience` like so:
+
+<span class="filename">Filename: src/lib.cairo</span>
+
+```rust
+mod customer_experience {
```
enitrat, line 73: I think we should change this example and just use something better to demonstrate how super is useful.

```rust
fn deliver_order() {}

mod back_of_house {
    fn fix_incorrect_order() {
        cook_order();
        super::deliver_order();
    }

    fn cook_order() {}
}
```

here you can see directly that you access a parent's module easily using `super` , which wasn't the case previously

src/ch06-03-paths-for-referring-to-an-item-in-the-module-tree.md
```diff
@@ -0,0 +1,153 @@
...
+}
+```
+
+In order to compile here we needed to use the `super` keyword, the super permits to access the `front_of_house` module from the `dining` module. The `super` keyword is relative to the current module and will take the parent module so we can't keep the same path with `::lib::`.
+
+You could also use the `super` keyword here like so:
+
+```rust
+mod dining {
+    use super::front_of_house;
```
enitrat, line 145: we don't want to use `use` yet in this part

enitrat (approved):
LGTM

PR #59, opened 2023-04-16
feat: add what is a crate

### LINKED ISSUE #44: Cairo Projects: Packages and Crates
- Explain what a crate is
- Create a package with `scarb` and explain the package structure (`lib.cairo` default crate root module..)

Estimated time: 1 hour

Th0rgal:
This introduces users to crates and closes #44 

It's a short and simple one. I don't think it's worth rewriting a copy of the scarb doc but tell me if you think something would be relevant here.

### DIFF
--- a/src/ch06-01-packages-and-crates.md
+++ b/src/ch06-01-packages-and-crates.md
@@ -1,0 +1,31 @@
+# Packages and Crates
+
+## What is a crate?
+Cairo packages can be organized into crates to modularize and group related code together. A crate is a compilation unit that contains Cairo code, functions, and data structures. This allows you to split your code into smaller, reusable parts, and to manage dependencies in a more structured way.
+
+## Creating a Package with Scarb
+You can create a new Cairo package using the scarb command-line tool. To create a new package, run the following command:
+```bash
+scarb new my_crate
+```
+
+This command will generate a new package directory named my_crate with the following structure:
+```
+my_crate/
+├── Scarb.toml
+└── src
+    └── lib.cairo
+```
+
+- `src/` is the main directory where all the Cairo source files for the package will be stored.
+- `lib.cairo` is the default root module of the crate, which is also the main entry point of the package. By default, it is empty.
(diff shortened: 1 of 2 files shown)


src/ch06-01-packages-and-crates.md
```diff
@@ -0,0 +1,33 @@
+# Packages and Crates
+
+## What is a crate?
+Cairo packages can be organized into crates to modularize and group related code together. A crate is a compilation unit that contains Cairo code, functions, and data structures. This allows you to split your code into smaller, reusable parts, and to manage dependencies in a more structured way.
```
enitrat, line 4: It is not clear what the difference between packages and crates is here.

I like the description of the Rust book, that I adapted here:

> A crate is the smallest amount of code that the Cairo compiler considers at a time. Even if you run `cairo-compile` rather than `scarb build` and pass a single source code file, the compiler considers that file to be a crate. Crates can contain modules, and the modules may be defined in other files that get compiled with the crate, as we’ll see in the coming sections.

src/ch06-01-packages-and-crates.md
```diff
@@ -0,0 +1,33 @@
...
+```toml
+[package]
+name = "my_crate"
+version = "0.1.0"
+
+[dependencies]
+# foo = { path = "vendor/foo" }
+```
+
+As you develop your package, you may want to organize your code into multiple Cairo source files. You can do this by creating additional `.cairo` files within the `src` directory or its subdirectories. It will be time to read [Defining Modules to Control Scope](./ch06-02-defining-modules-to-control-scope.md).
```
enitrat, line 33: I would remove that

enitrat (commented):
Content to add:
Description of what the _crate root_. is:

> The crate root is the `lib.cairo` source file that the Cairo compiler starts from and makes up the root module of your crate (we’ll explain modules in depth in the [“Defining Modules to Control Scope”](ch06-02-defining-modules-to-control-scope) section).

What is a Package, exactly?
(to adapt)
> It's a bundle of one or more crates with a `Scarb.toml` file that describes how to build those crates

Th0rgal:
Thank you @enitrat, I added your changes

enitrat (approved):
:100:

PR #65, opened 2023-04-17
feat: add use keyword usage

### LINKED ISSUE #47: Cairo Projects: Bringing Paths into Scope with the use Keyword

Th0rgal:
This closes #47 

I wasn't able to complete the `Using external packages (with Scarb)` section. It might be a problem on my side but I couldn't  use the cairo lib with the quaireaux example. I think it might not be ready so in the meantime I added a note.

What do you think?

### DIFF
--- a/src/ch06-04-bringing-paths-into-scope-with-the-use-keyword.md
+++ b/src/ch06-04-bringing-paths-into-scope-with-the-use-keyword.md
@@ -1,0 +1,58 @@
+# Bringing Paths into Scope with the `use` Keyword
+
+Writing out paths to call functions can be repetitive and inconvenient. To create a shortcut to a path, we can use the `use` keyword, which allows us to use the shorter name within the scope.
+
+## Full Paths and Shortcuts to Paths
+Using the `use` keyword, we can create shortcuts for paths, which can be used throughout the scope. To illustrate this, let's continue with the restaurant example:
+
+<span class="filename">Filename: src/lib.cairo</span>
+```rs
+mod front_of_house {
+    pub mod hosting {
+        pub fn add_to_waitlist() {}
+    }
+}
+
+use crate::front_of_house::hosting;
+
+pub fn eat_at_restaurant() {
+    hosting::add_to_waitlist(); // ✅ Shorter path
+}
+```
+
+By adding `use crate::front_of_house::hosting` in the crate root, `hosting` is now a valid name in that scope, as if the `hosting` module had been defined in the crate root. The paths brought into scope with `use` also check privacy, like any other paths.
+
+## Idiomatic Paths
+While it's possible to bring a function directly into scope with `use`, it is considered more idiomatic to bring the function's parent module into scope instead. This makes it clear that the function is not locally defined while still minimizing the repetition of the full path.
+
+For `Structs`, `Enums`, `Traits` and `Implementations`, it's idiomatic to specify the full path when using the `use` keyword.
+
+
+## Re-exporting Names in Module Files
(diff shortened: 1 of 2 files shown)


src/ch06-04-bringing-paths-into-scope-with-the-use-keyword.md
```diff
@@ -0,0 +1,58 @@
+# Bringing Paths into Scope with the `use` Keyword
+
+Writing out paths to call functions can be repetitive and inconvenient. To create a shortcut to a path, we can use the `use` keyword, which allows us to use the shorter name within the scope.
```
enitrat, line 3: To avoid repeting the word "use", let's rephrase it this way:

> We can create a shortcut to a path with the use keyword

src/ch06-04-bringing-paths-into-scope-with-the-use-keyword.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Full Paths and Shortcuts to Paths
+Using the `use` keyword, we can create shortcuts for paths, which can be used throughout the scope. To illustrate this, let's continue with the restaurant example:
+
+<span class="filename">Filename: src/lib.cairo</span>
+```rs
+mod front_of_house {
+    pub mod hosting {
```
enitrat, line 11: `pub` doesn't exist, please make sure the examples work!

src/ch06-04-bringing-paths-into-scope-with-the-use-keyword.md
```diff
@@ -0,0 +1,58 @@
...
+        pub fn add_to_waitlist() {}
+    }
+}
+
+use crate::front_of_house::hosting;
```
enitrat, line 16: `use crate`? You need to make sure your examples 😄

src/ch06-04-bringing-paths-into-scope-with-the-use-keyword.md
```diff
@@ -0,0 +1,58 @@
...
+fn eat_at_restaurant() {
+    add_to_waitlist(); // ✅ Re-exported name
+}
+```
+
+With this change, external code can now use the path `restaurant::add_to_waitlist()` instead of `restaurant::front_of_house::hosting::add_to_waitlist()`.
+
+Re-exporting is useful when the internal structure of your code differs from how programmers calling your code would think about the domain. This allows you to write your code with one structure, but expose a different structure, making your library well-organized for both programmers working on the library and programmers calling the library.
+
+## Using External Packages in Cairo with Scarb
```
enitrat, line 55: Ok for now

enitrat (changes requested):
The examples don't compile, and there's an overall lack of explanation. It feels like a poor version of the rustbook, because the concepts that translate between the two languages are poorly explained here.

You should directly adapt the Rust Book's page to Cairo, it's of good quality, and we don't need to be creative here.

enitrat:
Also don't forget to number the listings

Th0rgal:
> Also don't forget to number the listings And we can now rename imports with `as`

Alright, is it with alpha 7?

enitrat:
> > Also don't forget to number the listings And we can now rename imports with `as`
> 
> Alright, is it with alpha 7?

yes

enitrat (approved):

PR #73, opened 2023-04-19
feat:loops

enitrat:


src/ch02-05-control-flow.md
```diff
@@ -108,3 +108,139 @@ $ cairo-run main.cairo
...
+        }
+        'again!'.print();
+    }
+}
+```
+
+When we run this program, we’ll see `again!` printed over and over continuously
+until we stop the program manually, because the stop condition is never reached..
+While the compiler prevents us from writing programs without a stop condition (`break()` statement),
+the stop condition might never be reached, resulting in an infinite loop.
```
fricoben, line 144: Will the compiler see here that `i>10` will never happen ?

src/ch02-05-control-flow.md
```diff
@@ -108,3 +108,139 @@ $ cairo-run main.cairo
...
+
+[DEBUG]	again                          	(raw: 418346264942)
+
+[DEBUG]	again                          	(raw: 418346264942)
+
+Run panicked with err values: [375233589013918064796019]
+Remaining gas: 1050
+```
+
+> Note: Cairo prevents us from running program with infinite loops by including a gas meter. The gas meter is a mechanism that limits the amount of computation that can be done in a program. By setting a value to the `--available-gas` flag, we can set the maximum amount of gas available to the program. Gas is a unit of measurementsthat expresses the computation cost of an instruction. When the gas meter runs out, the program will stop. In this case, the program panicked because it ran out of gas, as the stop condition was never reached.
```
fricoben, line 162: Can you briefly define `Panick` here ? from my understanding the program runs out of gas (with the default max gas value).

src/ch02-05-control-flow.md
fricoben: Answered after but might be worth to say that the compiler won't see it directly

src/ch02-05-control-flow.md
```diff
@@ -108,3 +108,139 @@ $ cairo-run main.cairo
...
+    let mut i:usize = 0;
+    loop {
+        if i > 10{
+            break();
+        }
+        'again'.print();
+        i+=1;
+    }
+}
+```
```
fricoben, line 181: Maybe be you can give an example with the `--available-gas` flag and a very high condition on `i`

src/ch02-05-control-flow.md
```diff
@@ -108,3 +108,139 @@ $ cairo-run main.cairo
...
+        if counter == 10 {
+            break counter * 2;
+        }
+        counter += 1;
+    };
+
+    'The result is '.print();
+    result.print();
+}
+```
```
fricoben, line 234: You wrote the function two times (the second one might be the on you want to keep).

src/ch02-05-control-flow.md
```diff
@@ -108,3 +108,139 @@ $ cairo-run main.cairo
...
+
+## Summary
+
+You made it! This was a sizable chapter: you learned about variables, data types, functions, comments,
+`if` expressions, and loops! To practice with the concepts discussed in this chapter,
+try building programs to do the following:
+
+- Generate the *n*th Fibonacci number.
```
fricoben, line 242: n-th

fricoben (changes requested):
Some comments and questions. 

Learnt things, thx 🔥

src/ch02-05-control-flow.md
enitrat: the program runs out of gas, and thus panics with an out of gas error

enitrat: you understood it well :)

fricoben (approved):
LGTM

PR #84, opened 2023-04-24
Starknet contracts: ABIs and cross-contract interactions

Darlington02:


src/ch09-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,59 @@
...
+
+1. Must be appended with the `[abi]` macro.
+2. Your interface functions should have no implementations.
+3. You must explicitly declare the function's decorator.
+4. Your interface should not declare a constructor.
+5. Your interface should not declare state variables.
+
+Here's a sample interface for an ERC20 token contract:
+
+```rust
```
enitrat, line 21: Can you use OZ's implementation here? https://github.com/OpenZeppelin/cairo-contracts/blob/cairo-1/src/openzeppelin/token/erc20.cairo

src/ch09-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,59 @@
...
+    #[external]
+    fn transfer_from(sender: ContractAddress, recipient: ContractAddress, amount: u256);
+
+    #[external]
+    fn approve(spender: ContractAddress, amount: u256);
+}
+```
+
+## ABIs
+ABI stands for Application Binary Interface. ABIs gives a smart contract the ability to communicate and interact with external applications or other smart contracts. ABIs can be likened to APIs in traditional web development, which helps data flow between applications and servers.
```
enitrat, line 53: > ABIs can be likened to APIs in traditional web development, which helps data flow between applications and servers.

You should specify that an ABI is just a specification that defines how to encode and decode data when calling smart contract methods

src/ch09-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,59 @@
+# ABIs and Contract Interfaces
+
+Cross-contract interactions between smart contracts on a blockchain is a common practice which enables us build flexible contracts that can speak with each other.
+
+Achieving this on Starknet, requires something we call an interface.
+
+## Interface
+An interface, is a list of a contract's function definitions without implementations. In other words, an interface specfies the function declarations (name, parameters, visibility and return value) contained in a smart contract without including the function body.
+
+The Starknet plugin generates a contract interface for your defined contracts, a trait you can find in `<contract_name>::__abi`. But you can also write a contract interface manually to interact with external contracts you do not have direct access to.
```
enitrat, line 10: > The Starknet plugin generates a contract interface for your defined contracts, a trait you can find in `<contract_name>::__abi`.

This sentence is not very clear 🤔

src/ch09-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,59 @@
...
+Cross-contract interactions between smart contracts on a blockchain is a common practice which enables us build flexible contracts that can speak with each other.
+
+Achieving this on Starknet, requires something we call an interface.
+
+## Interface
+An interface, is a list of a contract's function definitions without implementations. In other words, an interface specfies the function declarations (name, parameters, visibility and return value) contained in a smart contract without including the function body.
+
+The Starknet plugin generates a contract interface for your defined contracts, a trait you can find in `<contract_name>::__abi`. But you can also write a contract interface manually to interact with external contracts you do not have direct access to. 
+For your Cairo code to qualify as an interface, it must meet the following requirements:
+
```
enitrat, line 12: I think the most important thing here is missing: Interfaces are implemented with traits in Cairo. It might be worth adding a ref link to the Traits sections, where we specify that traits specify functionality blueprints.

src/ch09-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,59 @@
...
+4. Your interface should not declare a constructor.
+5. Your interface should not declare state variables.
+
+Here's a sample interface for an ERC20 token contract:
+
+```rust
+use starknet::ContractAddress;
+
+#[abi]
+trait IERC20 {
```
enitrat, line 25: same, use OZ's (as we want to avoid the use of the `felt252` type generally, we should avoid using it in the book as well, except temporarily for short strings)

src/ch09-02-contract-dispatcher.md
```diff
@@ -0,0 +1,92 @@
...
+
+    #[external]
+    fn transfer_from(sender: ContractAddress, recipient: ContractAddress, amount: u256);
+
+    #[external]
+    fn approve(spender: ContractAddress, amount: u256);
+}
+```
+
+Contracts annotated with the `abi` macro/attribute are programmed to automatically generate and export the relevant dispatcher logic on compilation. The compiler also generates a new trait, two new structs(one for contract calls, and the other for library calls) and their implementation of this trait. Our interface is expanded into something like this:
```
enitrat, line 42: > with the `abi` macro/attribute

just attribute

src/ch09-02-contract-dispatcher.md
```diff
@@ -0,0 +1,92 @@
...
+    fn transfer(self: T, recipient: ContractAddress, amount: u256);
+}
+
+#[derive(Copy, Drop)]
+struct IERC20Dispatcher {
+    contract_address: starknet::ContractAddress,
+}
+
+impl IERC20DispatcherImpl of IERC20DispatcherTrait::<IERC20Dispatcher> {
+    fn get_name(self: T) -> felt252 {
```
enitrat, line 56: > fn get_name(self: T) -> felt252 {

I think `T` should be `IERC20Dispatcher` here

enitrat (changes requested):
Great PR! Thank you 💯 

A few things to address:
- Can you move the chapter09 before the Appendix in the summary
- We should have an entire section dedicated to smart contracts, something like `ch09-starknet-smart-contracts`. The content you wrote would be a subsection of this section. Can you do this change as well? 
- I think we could merge these pages together to improve the structure of the book
- Add Listing numbering (e.g. Listing 9-1: ERC20 Interface by Openzeppelin)

enitrat:
Also now that I think about it using the IERC20 example might not be ideal as it's a bit too long and it's probably better if we can restrict the scope so that the reader can experiment with the examples themselves. 

I'll leave it up to you, but this part will probably be reworked later with smarter examples. In the first Contract chapter we will very likely write a simple contract, and it would be cool if the user could write a second contract that makes calls to the contract he wrote previously, and finally test all of this in a third section.

Darlington02:
> Great PR! Thank you 💯
> 
> A few things to address:
> 
> * Can you move the chapter09 before the Appendix in the summary
> * We should have an entire section dedicated to smart contracts, something like `ch09-starknet-smart-contracts`. The content you wrote would be a subsection of this section. Can you do this change as well?
> * I think we could merge these pages together to improve the structure of the book
> * Add Listing numbering (e.g. Listing 9-1: ERC20 Interface by Openzeppelin)

So here's the structure I ended up creating..
- Starknet smart contracts  9
       - ABIs and Contract Interactions  9.1
             - ABIs and Interfaces 9.1.1
             - Contract Dispatchers 9.1.2
             - Library Dispatchers and system calls 9.1.3
       
The idea is sections under starknet contracts can go in as subsections, but can also be parent sections to other subsections too...does this make sense to do? Cause I feel like cramming all the subsections under let's say the ABIs and Contract Interactions into one page, might be too overwhelming for the reader

src/ch09-01-abis-and-interfaces.md
Darlington02: Uhmm, I wanted to go that route, but OZ's implementation considers the functions as all internals (which they actually are), so there are no visibilities specified, which I thought might not directly communicate the point to the user (that specifying visibility is important).

Or should I just use them, then let the user know there's no visibility specified because the functions are all internal functions?

src/ch09-01-abis-and-interfaces.md
Darlington02: done

src/ch09-01-abis-and-interfaces.md
Darlington02: felts are only returned by the `get_name` and `get_symbol` functions, which returns short strings. the `get_total_supply` was a mistake..was supposed to return a u256 :)

but if you think it's better to go with OZ's implementation, let me know and I'd switch before updating the PR

src/ch09-01-abis-and-interfaces.md
enitrat: hmm, I see
Let's do both actually: Keep it as you did right now, but just modify the function signatures to match OZ's

src/ch09-01-abis-and-interfaces.md
enitrat: See comment above

src/ch09-01-abis-and-interfaces.md
Darlington02: alright, on it :)

enitrat:
> Cause I feel like cramming all the subsections under let's say the ABIs and Contract Interactions into one page, might be too overwhelming for the reader

Ok, let's keep it like this for now. We'll see if we end up with too many subsections, or if some can be merged, whether we group them or not.
However I still think at least ContractDispatcher and LibraryDispatcher could be grouped together

src/ch09-02-contract-dispatcher.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+**NB:** The expanded code for our IERC20 interface is a lot more robust, but to keep this chapter concise and straight to the point, we focused on one view function `get_name`, and one external function `transfer`.
+
+It's also worthy of note that, all these is abstracted behind the scenes, thanks to the power of Cairo plugins.
+
+## Calling Contracts using the Contract Dispatcher
+Calling another contract using the Contract interface dispatcher, calls the contract's logic in it's context, and in most cases may alter the contract's state. Here's an example:
+
+```rust
```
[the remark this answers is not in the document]
Darlington02: updated

src/ch09-03-library-dispatcher-and-system-calls.md
```diff
@@ -0,0 +1,71 @@
...
+fn transfer_token(
+    recipient: ContractAddress, amount: u256
+) -> felt252 {
+    IERC20LibraryDispatcher { class_hash: starknet::class_hash_const::<0x1234>() }.transfer(recipient, amount);
+} 
+```
+
+As you can see, we had to first import the `IERC20DispatcherTrait` and `IERC20LibraryDispatcher` which was generated and exported on compiling our interface, then we make calls to the methods implemented for the `IERC20LibraryDispatcher` struct (`get_name`, `transfer`, etc), passing in the `class_hash` parameter which represents the class of the contract we want to call.
+
+## Calling Contracts using low-level System calls
```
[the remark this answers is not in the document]
Darlington02: done

Darlington02:
> 

yes, we could have a dedicated section, that's practical oriented, walking the reader through all they've learnt from writing to deploying

src/ch09-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,59 @@
...
+    #[view]
+    fn get_name() -> felt252;
```
enitrat: these should be named `name()` , `symbol()`, without get prefixes (you will need to modify these names in the text below as well)

src/ch09-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,59 @@
...
+    fn transfer(recipient: ContractAddress, amount: u256);
+
+    #[external]
+    fn transfer_from(sender: ContractAddress, recipient: ContractAddress, amount: u256);
+
+    #[external]
+    fn approve(spender: ContractAddress, amount: u256);
+}
+```
+
```
enitrat: Add Listing numbering here

src/ch09-02-contract-dispatcher.md
```diff
@@ -0,0 +1,92 @@
+# Contract Dispatcher
+
+Each time a contract interface is created on Starknet, two dispatchers are automatically created and exported:
+1. The Contract Dispatcher
+2. The Library Dispatcher
+
+In this chapter, we are going to extensively discuss how the Contract dispatcher works and it's usage.
+
+To effectively break down the concepts in this chapter, we are going to be using the IERC20 interface from the previous chapter:
+
```
enitrat: And here instead of copy-pasting, just say that we are going to use the IERC20 interface from Listing x.x

src/ch09-00-01-abis-and-interfaces.md
```diff
@@ -0,0 +1,65 @@
...
+    fn transfer(recipient: ContractAddress, amount: u256) -> bool;
+
+    #[external]
+    fn transfer_from(sender: ContractAddress, recipient: ContractAddress, amount: u256) -> bool;
+
+    #[external]
+    fn approve(spender: ContractAddress, amount: u256) -> bool;
+}
+```
+
```
enitrat, line 57: Can you place the Listing here below the example, as done in previous chapters?

src/ch09-00-02-contract-dispatcher-library-dispatcher-and-system-calls.md
```diff
@@ -0,0 +1,173 @@
...
+
+Each time a contract interface is created on Starknet, two dispatchers are automatically created and exported:
+1. The Contract Dispatcher
+2. The Library Dispatcher
+
+In this chapter, we are going to extensively discuss how these dispatchers works and their usage.
+
+To effectively break down the concepts in this chapter, we are going to be using the IERC20 interface from the previous chapter:
+
+<span class="caption">Listing 9-2: A simple ERC20 Interface</span>
```
enitrat, line 11: If this is the same as Listing 9-1, we can just refer to Listing 9-1 instead of defining Listing 9-2, which is the same as 9-1

src/ch09-00-02-contract-dispatcher-library-dispatcher-and-system-calls.md
```diff
@@ -0,0 +1,173 @@
...
+
+impl IERC20LibraryDispatcherImpl of IERC20DispatcherTrait::<IERC20LibraryDispatcher> {
+    fn get_name(self: IERC20LibraryDispatcher) -> felt252 {
+        // starknet::syscalls::library_call_syscall  is called in here
+    }
+    fn transfer(self: IERC20LibraryDispatcher, recipient: ContractAddress, amount: u256) {
+        // starknet::syscalls::library_call_syscall  is called in here
+    }
+}
+```
```
enitrat, line 130: Listing 9-2

src/ch09-00-02-contract-dispatcher-library-dispatcher-and-system-calls.md
```diff
@@ -0,0 +1,173 @@
...
+#[external]
+fn transfer_token(
+    address: starknet::ContractAddress, selector: felt252, calldata: Array<felt252>
+) -> Span::<felt252> {
+    starknet::call_contract_syscall(
+        :address, entry_point_selector: selector, calldata: calldata.span()
+    ).unwrap_syscall()
+} 
+```
+
```
enitrat, line 172: Listing 9-4

enitrat (approved):
💯

PR #95, opened 2023-04-30
Fix Type casting error description

zkcarter:
As example code show that felt252 is smaller than a u256, we can use the into() method, which felt252 is source type and u256 is destination type.
But above descriptionis exactly opposite.


enitrat (approved):
thanks 💯

PR #96, opened 2023-04-30
feat: add panic chapter and reorganize starknet and errors chapter

### LINKED ISSUE #78: Error Handling: Unrecoverable Errors with `panic()`
- Explain what `panic` does.
- Explain what actually happens when a Cairo program panics. (read this [article](https://medium.com/nethermind-eth/under-the-hood-of-cairo-1-0-exploring-sierra-7f32808421f5) to understand what panics _actually_ are).

Given that panics are not very exploitable nor informative in Cairo, this chapter is pretty short.

Estimated time: 1 hour

zarboq:
Closes #78


src/ch09-01-unrecoverable-errors-with-panic.md
```diff
@@ -0,0 +1,38 @@
+# Unrecoverable Errors with panic
+
+Sometimes, during program execution, unforeseen issues can occur, leading to runtime errors. In Cairo, these errors can be handled through the panic macro. There are two primary ways to trigger a panic in Cairo: unintentionally, by performing actions that cause the code to panic (e.g., accessing an array beyond its bounds), or explicitly, by using the panic macro.
```
enitrat, line 3: I wouldn't call that `handled`. You're just saying "ok, there's an error, i cant do anything about it, let's exit the program."

Please re-do this first sentence

enitrat: panic is not a macro but a function of the core library

src/ch09-01-unrecoverable-errors-with-panic.md
```diff
@@ -0,0 +1,38 @@
...
+use array::ArrayTrait;
+
+fn main() {
+    let mut data = ArrayTrait::new();
+    data.append(2);
+    panic(data);
+}
+```
+
+When you run the program, you’ll see something like this:
```
enitrat, line 21: Running the program will produce the following output

src/ch09-01-unrecoverable-errors-with-panic.md
```diff
@@ -0,0 +1,38 @@
...
+
+```console
+$ cairo-run test.cairo 
+Run panicked with err values: [2]
+```
+
+The call to panic causes the error message.
```
enitrat, line 28: Calling the `panic` function causes the program to stop its execution and return early with the error `2`. 

can you add a print statement right after the panic in the example to show that it's not executed?

src/ch09-01-unrecoverable-errors-with-panic.md
```diff
@@ -0,0 +1,38 @@
...
+
+Another option available is to use panic_with_felt252. Let's consider an example:
+
+```rust
+fn main() {
+    panic_with_felt252(2);
+}
+```
+
+Running this program will display the same error message as before. However, if you don't require the use of an array and need to return multiple values within your error, panic_with_felt252 provides a more concise alternative.
```
enitrat, line 38: > However, if you don't require the use of an array and need to return multiple values within your error,

rephrase. Also, mention te cases in which  would we want to use an array

enitrat (changes requested):

enitrat:
Just realized that it's not very clear what we can use `panic` for. 
Can you add a subsection about the `assert` statements to demonstrate a good use of `panic`? 
`assert` are basically panics `if cond not true then panic with message` statements. 

In your example, use something meaningful, for example someone tries to access a resource and he's not authorized.

src/ch09-01-unrecoverable-errors-with-panic.md
```diff
@@ -38,4 +38,25 @@ fn main() {
...
+Here is an example of its usage:
+
+```rust
+fn number_not_zero(number: u8) -> bool {
+    number != 0
+}
+
+fn main() {
+    let my_number: u8 = 0;
+    assert(number_not_zero(my_number), 'number is zero');
```
enitrat, line 56: just number !=0

enitrat (approved):
lfg 💯

PR #97, opened 2023-04-30
Single tuple trailing comma

shramee:
Tuple with a single value needs a trailing comma.


enitrat:
Actually, this enum variant can hold a single value that is not inside a tuple. I'll make the required modifications.

PR #124, opened 2023-05-05
added iterative example

prix0007:
Adds an Iterative example in the enums section.

### DIFF
--- a/src/ch05-01-enums.md
+++ b/src/ch05-01-enums.md
@@ -124,6 +124,24 @@ fn find_value_recursive(
     find_value_recursive(arr, value, index + 1_usize)
 }
 
+fn find_value_iterative(arr: @Array<felt252>, value: felt252) -> Option<usize> {
+    let length = arr.len();
+    let mut index = 0_usize;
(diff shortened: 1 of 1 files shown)


enitrat:
Can you just add some context above the example and say that we demonstrate two different ways, using a) recursive functions and b) loops?

Thank you for this contribution!

prix0007:
Makes sense.
Thanks for mentioning, updated accordingly

src/ch05-01-enums.md
```diff
@@ -95,6 +95,10 @@ The `Option` enum is helpful because it allows you to explicitly represent the p
 
 To give you an example, here is a function which returns the index of the first element of an array with a given value, or None if the element is not present.
 
+We are demonstrating two approaches for above function:
```
enitrat, line 98: typo: for the above function

enitrat (approved):

src/ch05-01-enums.md
prix0007: fixed, thanks

PR #129, opened 2023-05-08
Update ch02-02-data-types.md

Giwook-Han:
The previous code was able to run because the compiler inferred the type of the variable. However, this could be confusing to understand.


enitrat:
Hm, i'm not sure we should change it. The default type when nothing is specified is `felt252`, this is explained a bit earlier I believe? + this is a good example of type inference
@fricoben wdyt?

enitrat:
Alright, let's favor strong defined types for clarity, thanks! :)

enitrat (approved):

PR #130, opened 2023-05-09
feat: i18n support for mdbook

glihm:
This PR adds support for i18n mdbook extension.

The idea is to also help contributors to work on translations. For that the README
was extended and a script `translation.sh` was added to facilitate the work.

The CI is also reworked to use rust cache, and drop the build time from 5 minutes to 30 seconds. The CI always build the english first and then the translations.


glihm:
We have ongoing work to port our previous translations to the repository. The idea of this PR is to integrate the new structure and new files to then start PRs with translations only.

Also, as added in the README, we suggest that any PR related to only translation must be prefixed by `i18n`. This may help maintainers to know that only `po` files were changed, and the structure of the book is completely intact.

enitrat:
Sweet! Where was the CI reworked? I don't see any modification in the yml files

glihm:
Just pushed, thank you for pointing this out @enitrat.

enitrat (approved):

enitrat:
Thanks, that's super cool :)

PR #138, opened 2023-05-11
i18n: Spanish Version es.po

Nadai2010:
Hello, great work of art once again.

Great contribution also from @glihm  and @dpinones  to be able to have the versions in several languages and in a clean way.

I have migrated what we had from the Spanish version, and added and improved the ones that were missing... It has been a hard and double road, but the truth is that I love the content. 

Can you check how you see it? We think the only new 9.1 is missing from @Darlington02


Nadai2010:
Thanks, I have passed it to DRAFTS, in case any member in Spanish can review it.

Although we can improve and change with updates, or wrong words if any.

Also now the flow with po, is much better for this kind of work and updates without bothering us, more than the account. 

Whatever you think is best we will do, thank you.

glihm:
@enitrat we have gone through a first internal revision. Today we should be able to rebase and push the changes to this PR, which will be re-open. :+1:

Nadai2010:
Thanks @glihm and @enitrat  for the review and help, what a great ecosystem and teams are being created. We will be adding the new changes now that the translation flow is much better. Great work

The Spanish speaking version is ready to be launched. 🚀

enitrat (approved):

Issue #137, opened 2023-05-10
feat: keywords


enitrat:
This appendix will list the reserved keywords in Cairo, and what they're used for.

For example, the `return` keyword is used to return from a function.
The expected output is similar to the Rust book's [keywords appendix](https://doc.rust-lang.org/book/appendix-01-keywords.html) page.

You can refer to this page of the [cairo documentation](https://github.com/starkware-libs/cairo/blob/main/docs/reference/src/components/cairo/modules/language_constructs/pages/keywords.adoc) Your objective here is to provide context for these keywords and explain what they're used for.

Estimated time: 1 hour

rkdud007:
want to try this one 👋

rkdud007:
yo think you can close this issue :)

PR #143, opened 2023-05-12
appendix keywords ( block : implicits )

rkdud007:
issue #137 

done appendix with keywords 
except for keyword `implicits` :/


src/appendix-01-keywords.md
```diff
@@ -0,0 +1,83 @@
...
+---
+
+### Reserved keywords
+
+These keywords aren't used yet, but they are reserved for future use.
+They have the same restrictions as strict keywords.
+The reasoning behind this is to make current programs forward compatible with future versions of
+Cairo by forbidding them to use these keywords.
+
+- `as`
```
zarboq, line 56: Is it normal that 'as' is both in strict and reserved ?

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,83 @@
...
+- `assert`
```
zarboq, line 57: I think assert is already usable

rkdud007: hey i got keyword list from

 https://github.com/starkware-libs/cairo/blob/a0ead7c0b8e297d281c7213151cd43ac11de5042/crates/cairo-lang-parser/src/lexer.rs#L332

and i think assert is not usable currently :)

src/appendix-01-keywords.md
rkdud007: ye my bad commit again haha

src/appendix-01-keywords.md
enitrat: Where did you get the reserved keyword list from? I can, for example, create a variable of name `assert` without any problem. `assert` is a function from the core library and not a keyword per se.

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+### Strict keywords
+
+These keywords can only be used in their correct contexts.
+They cannot be used as names of any items.
+
+- `as` - perform primitive casting, disambiguate the specific trait containing
+- `const` - define constant items or constant raw pointers
+- `false` - Boolean false literal
+- `true` - Boolean true literal
+- `extern` - link an external function or variable
```
enitrat, line 23: unlike in rust, `extern` functions are designating functions defined at the compiler level, corresponding to Sierra libfuncs, which can be compiled down to CASM.

For example, if you want to create a function using a hint available at the cairo1 level, you would need to: 

 - Write the CASM code corresponding to this function, hint included
 - Create a Sierra libfunc that can be compiled to this function
 - Export this libfunc at the cairo1-level, by importing its declaration with the `extern` keyword.
 -

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+- `type` - define a type alias or associated type
```
enitrat, line 24: what is an associated type? I'm not sure it exists yet in Cairo. This is not valid:

```rust

trait ABC{
    type A;
    type B;
}
```
But we can define type aliases indeed ;)
```rust
type abc = u128;

fn main() -> abc {
    let x: abc = 3;
    x
}
```

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+- `fn` - define a function or the function pointer type
```
enitrat, line 25: we don't have function pointers

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+- `return` - return from function
+- `match` - match a value to patterns
+- `if` - branch based on the result of a conditional expression
+- `loop` - loop unconditionally
+- `continue` - continue to the next loop iteration
+- `break` - exit a loop immediately
+- `else` - fallback for `if` and `if let` control flow constructs
+- `use` - bring symbols into scope
+- `ref` - bind by reference
+- `mut` - denote mutability in references, raw pointers, or pattern bindings
```
enitrat, line 42: > denote mutability in references, raw pointers, or pattern bindings

just denote variable mutability

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+- `nopanic` - Functions marked with this notation, means that the function will never panic.
+<!-- need help for impliccit -->
+- `implicits`
```
enitrat, line 45: For implicits, you can introduce them as as special kind of function parameters that are required to perform certain actions (like, for example, checking whether a value is inside a range, or checking how much gas is left during a run). This is due to the layout of the Cairo VM.

more info here https://www.cairo-lang.org/docs/how_cairo_works/builtins.html?highlight=implicits

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+- `where`
+- `while`
+- `with`
+- `yield`
+
+---
+
+### Contextual keywords
+
+Some grammar productions may make use of new keywords not listed here. Such keywords have special meaning only in these certain contexts. Outside these places, these character sequences are treated as regular identifiers, thus it is possible to declare a function or variable with such names.
```
enitrat, line 78: I didn't really understand this paragraph. Is there an example of this?

src/appendix-01-keywords.md
rkdud007: i didn't find reserved keyword list from cairo repo itself, so refered from https://github.com/starkware-libs/cairo/blob/main/docs/reference/src/components/cairo/modules/language_constructs/pages/keywords.adoc here , 

and hmm like i just mentioned above, got keyword from `lexer.rs` but seems like `assert` is from somewhere else, will check more

enitrat: > so refered from [here](https://github.com/starkware-libs/cairo/blob/main/docs/reference/src/components/cairo/modules/language_constructs/pages/keywords.adoc) 

Great, thanks! I tried to name variables with these reserved keywords and it worked. Perhaps that's a bug and we should open an issue for it.

enitrat:
https://github.com/starkware-libs/cairo/issues/3176

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+- contextual
+
+---
+
+### Strict keywords
+
+These keywords can only be used in their correct contexts.
+They cannot be used as names of any items.
+
+- `as` - perform primitive casting, disambiguate the specific trait containing
```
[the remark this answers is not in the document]
rkdud007: so base on upper example, can say `as` as Type Assertions?  
btw i tried to find `into` or `try_into` as type cast example but couldn't find one. would you have anything?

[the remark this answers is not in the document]
rkdud007: can say `as` is for type assertion right ?
and `into` and `try_into` is method so, i guess there is no type casting keyword :)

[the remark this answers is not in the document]
rkdud007: * i think its just rename import not type assertion

src/appendix-01-keywords.md
```diff
@@ -0,0 +1,82 @@
...
+- `const` - define constant items or constant raw pointers
```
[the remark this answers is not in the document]
rkdud007: ops thought const can use as raw pointers like rust i don't think it can. deleted

src/appendix-01-keywords.md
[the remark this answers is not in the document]
enitrat: > i think its just rename import not type assertion

Yes

rkdud007:
quick question : is `for` is usable keyword ? 🤔 @enitrat

src/appendix-01-keywords.md
rkdud007: Is this chapter implies about type can use as associated type??

https://github.com/starkware-libs/cairo/blob/main/adrs/003_simple_trait_system.md#trait-top-level-item

rkdud007: wait sorry I think its Cairo 0. hmmm does Cairo 1 is not supporting?

src/appendix-01-keywords.md
rkdud007: This section is base on original starkware cairo documentation :) currently doesn't seems like its using, can delete 

https://github.com/starkware-libs/cairo/blob/main/docs/reference/src/components/cairo/modules/language_constructs/pages/keywords.adoc#contextual-keywords

src/appendix-01-keywords.md
rkdud007: Thanks for clear explanation ! ye seems like Cairo1 doesn't use `extern keyword as external function use `#[external]` instead. 

I think can say `extern` keyword as  `Function defined at the compiler level using hint available at cairo1 level`

src/appendix-01-keywords.md
rkdud007: Ah alright, also works same as cairo1 right? :) i'll apply that explanation

rkdud007:
Sorry that commit looks quite weird 😅 Base commit that i changed, base on review is 

https://github.com/cairo-book/cairo-book.github.io/pull/143/commits/a21d07435ced8e845ee6b743c1849124ac8a8f1b 

This one!

src/appendix-01-keywords.md
enitrat: > using hint available at cairo1 level with this declaration

this doesn't make a lot of sense. I would rephrase to:
A function defined in the compiler as a Sierra libfunc and that generates CASM code.

src/appendix-01-keywords.md
enitrat: > wait sorry I think its Cairo 0.

It's neither Cairo 0 or valid Cairo1. I think it was just a proposal, but it's not effective yet.

rkdud007:
@enitrat resolved it ! there are some symbol A -> B changes catches as file change tho

enitrat (approved):
🔥

PR #145, opened 2023-05-13
feat: Appendix Derivable Traits

### LINKED ISSUE #134: feat: Derivable traits Appendix
This appendix page explains which traits can be derived, how to do it, when it is required, etc.

- Serde
- Copy
- Drop
- PartialEq

You will also research if other traits can be derived as well.

Estimated time: 1 to 2 hours

zarboq:
Closes #134

### DIFF
--- a/src/appendix-03-derivable-traits.md
+++ b/src/appendix-03-derivable-traits.md
@@ -1,0 +1,153 @@
+# Appendix C: Derivable Traits
+
+In various places in the book, we’ve discussed the `derive` attribute, which you can apply to a struct or enum definition. The `derive` attribute generates code to implement a default trait on the type you’ve annotated with the `derive` syntax.
+
+In this appendix, we provide a comprehensive reference detailing all the traits in the standard library compatible with the `derive` attribute.
+
+These traits listed here are the only ones defined by the core library that can be implemented on your types using `derive`. Other traits defined in the standard library don’t have sensible default behavior, so it’s up to you to implement them in the way that makes sense for what you’re trying to accomplish.
+
+The list of derivable traits provided in this appendix does not encompass all possibilities: external libraries can implement `derive` for their own traits, expanding the list of traits compatible with `derive`.
+
+## PartialEq for equality comparison
+
+The `PartialEq` trait allows for comparison between instances of a type for equality, thereby enabling the == and != operators.
+
+When `PartialEq` is derived on structs, two instances are equal only if all fields are equal, and the instances are not equal if any fields are not equal. When derived on enums, each variant is equal to itself and not equal to the other variants.
+
+Example:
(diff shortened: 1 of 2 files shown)


src/appendix-03-derivable-traits.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Clone and Copy for Duplicating Values
+
+The `Clone` trait provides the functionality to explicitly create a deep copy of a value.
+
+Deriving `Clone` implements the `clone` method, which, in turn, calls clone on each of the type's components. This means all the fields or values in the type must also implement `Clone` to derive `Clone`.
+
+Example:
+
+```Rust
```
enitrat, line 44: doesn't compile for me

src/appendix-03-derivable-traits.md
```diff
@@ -0,0 +1,151 @@
...
+fn main() {
+    let first_struct = A {
+        item: 2
+    };
+    let second_struct = first_struct.clone();
+    assert(second_struct.item == 2, 'Not equal');
+}
+```
+
+The `Copy` trait allows for the duplication of values by copying the bit data stored on the stack only.
```
enitrat, line 61: what is the stack in Cairo? :)

Copy is just a trait that is used to ensure type safety at the sierra-level, so that it's not possible to use two array/dict instances at different places.

src/appendix-03-derivable-traits.md
```diff
@@ -0,0 +1,151 @@
...
+
+The `Copy` trait doesn’t define any methods to prevent programmers from overloading those methods and violating the assumption that no arbitrary code is being run. That way, all programmers can assume that copying a value will be very fast.
```
enitrat, line 63: this is also not really accurate

src/appendix-03-derivable-traits.md
```diff
@@ -0,0 +1,151 @@
...
+    };
+    let second_struct = first_struct;
+    assert(second_struct.item == 2, 'Not equal');
+    assert(first_struct.item == 2, 'Not Equal'); // Copy Trait prevents firs_struct from moving into second_struct
+}
+```
+
+## Serializing with Serde
+
+`Serde` provides trait implementations for `serialize` and `deserialize` functions for data structures defined in your crate, allowing them to be represented conveniently in all of Serde's data formats.
```
enitrat, line 87: > to be represented conveniently in all of Serde's data formats.
 
 I think we can be more specific and say that the type is transformed into an array / deserialized from an array

src/appendix-03-derivable-traits.md
enitrat: works now

enitrat (approved):

Issue #139, opened 2023-05-11
feat: code snippet tester


enitrat:
To make the maintenance process easier, I would like to integrate in the CI a tool that extracts all compilable code snippets and compiles them. This will ensure that code snippets are executable and make it easier to spot breaking changes due to version updates. 

We could annotate "fenced" code blocks with a `does_not_compile` attribute to indicate code snippets that can't be compiled, like chunks of programs or isolated functions. 

@glihm suggested to adapt [this program ](https://github.com/google/comprehensive-rust/tree/main/mdbook-exerciser) for our needs

enitrat:
Would love to :pray:

glihm:
@enitrat and others, I would like to have your opinion on this:

We use fenced code block for cairo examples. Which is nice. Here is my suggestion to hide the implementation details to the reader:

1. Instead of having a filename formatted in HTML or other before the code block, we can add attributes to the tag. The tag is not visible after rendering.
2. The tag can have several attributes, and the syntax color of the code is only determined by the first attribute.
3. Like this, we can then only specify a filename for the fence code block when we want mdbook to generate a cairo program.


````
```rust,file=hello_world.cairo
use debug::PrintTrait;
fn main() {
    'Hello, world!'.print();
}                                                                                                                                        
```
````

Like this, the syntax is preserved, and the parser will know it has to generate a file named `hello_worl.cairo`.

````
```rust
    'Hello, world!'.print();                                                                                                                                    
```
````

Like this, we just want to explain part of the code, no file will be generated.

Sounds good for you guys? Or do you prefer relying on a special syntax, like the filename, that is always preceding the code block?

enitrat:
Sounds good

PR #146, opened 2023-05-14
feat: add Cairo code snippet tester (#139)

glihm:
Related to issue #139.

Add to mdbook a Cairo renderer to extract Cairo programs from code block using markdown parser and associated CI to compile the programs.

Some considerations:

1. To compile a code block, the attribute `file=` must be added to the code block tag. It is expected to be the **LAST** attribute in the list.
   ````
   ```rust,file=hello_world.cairo
   use debug::PrintTrait;
   fn main() {
       'Hello, world!'.print();
   }
   ```
   ````
   
2. The current implementation expects the code block to be **VALID**. There is no extra Cairo code added to the output program.

3. For github pages, it's important to note that now we have 2 backends to render the book (html and cairo). For this reason, the CI was modified to ensure that only the html render is uploaded to github pages.

4. The CI generates a table where the current implementation shows the logs. Maybe it's too verbose and we can only keep the logs for errors.
[example 1](https://github.com/glihm/cairo-book.github.io/actions/runs/4971201782)
[example 2](https://github.com/glihm/cairo-book.github.io/actions/runs/4971193283)


glihm:
@enitrat adding manually the filename is a bit tedious and error prone. After doing some chapters, I don't think it's the good approach.

I propose the following:
1. The mdbook cairo backend automatically names the files based on it's chapter and a simple counter.
2. We can add a rule for the mdbook backend to automatically extract all code blocks where `fn main` is found.
3. To ensure we can add `fn main` is code block without compiling the code, we can add something similar you commented: `does_not_compile` in the attributes list, to have the mdbook backend ignoring the code block.

The only drawback of this approach is that, if a cairo program fails, and let's say we have 10 code blocks compiled from a chapter X, one have to read the entire chapter to find the index of the code block with the `fn main` that is compiled to debug it. Which, in some scenarios, may be a bit tedious and confusing. But if we keep the error logs in the CI summary, it should be fine.

.github/workflows/verify-cairo-programs/action.yml
```diff
@@ -0,0 +1,19 @@
...
+
+description: Verify if all Cairo programs extracted from markdown can be compiled.
+
+runs:
+  using: composite
+  steps:
+    - name: Copy Cairo program verifier script
+      shell: bash
+      run: |
+        cp cairo_programs_verifier.sh book/cairo/cairo-programs/
```
enitrat, line 11: what if `book/cairo/cairo-programs/` doesn't exist?

enitrat (commented):
Looks great!

enitrat:
> @enitrat adding manually the filename is a bit tedious and error prone. After doing some chapters, I don't think it's the good approach.
> 
> I propose the following:
> 
> 1. The mdbook cairo backend automatically names the files based on it's chapter and a simple counter.
> 2. We can add a rule for the mdbook backend to automatically extract all code blocks where `fn main` is found.
> 3. To ensure we can add `fn main` is code block without compiling the code, we can add something similar you commented: `does_not_compile` in the attributes list, to have the mdbook backend ignoring the code block.
> 
> The only drawback of this approach is that, if a cairo program fails, and let's say we have 10 code blocks compiled from a chapter X, one have to read the entire chapter to find the index of the code block with the `fn main` that is compiled to debug it. Which, in some scenarios, may be a bit tedious and confusing. But if we keep the error logs in the CI summary, it should be fine.

I like this approach :) Thank you!

.github/workflows/verify-cairo-programs/action.yml
glihm: It always exists as mdbook rendered creates it in it's main function.
The only reason this folder may not be present is if `mdbook build` command fails. Which will make the CI fail before this step.

glihm:
> > @enitrat adding manually the filename is a bit tedious and error prone. After doing some chapters, I don't think it's the good approach.
> > I propose the following:
> > 
> > 1. The mdbook cairo backend automatically names the files based on it's chapter and a simple counter.
> > 2. We can add a rule for the mdbook backend to automatically extract all code blocks where `fn main` is found.
> > 3. To ensure we can add `fn main` is code block without compiling the code, we can add something similar you commented: `does_not_compile` in the attributes list, to have the mdbook backend ignoring the code block.
> > 
> > The only drawback of this approach is that, if a cairo program fails, and let's say we have 10 code blocks compiled from a chapter X, one have to read the entire chapter to find the index of the code block with the `fn main` that is compiled to debug it. Which, in some scenarios, may be a bit tedious and confusing. But if we keep the error logs in the CI summary, it should be fine.
> 
> I like this approach :) Thank you!

Thank you for the feedback. I'll go that way then and I'll also add any identifying information to help the cairo-book contributor finding the program that does not compile.

glihm:
@enitrat are we expecting the CI to fail is at least one Cairo program is not compiled successfully?

I would say yes, but would be happy to have a confirmation. :+1:

enitrat:
yes!

glihm:
@enitrat the CI is working as expected, I have made some modifications to ensure the correct run and results even on fail.

I have some troubles on code blocks, I would be happy to have your feedback on those. You can find the last CI results [in my fork here](https://github.com/glihm/cairo-book.github.io/actions/runs/4998745516).

But it's functional.

Some notes:
1. To facilitate any contributor work to locate the error, I have used the same nomenclature as your chapter files. And the cairo programs verifier always enumerate `fn main()` functions from top to bottom of the file. Like this, it's easy to locate which function in which chapter.
2. The chapter 09 has two 01 sections, which makes the first one overwritten by the second one in the CI. This is a typo, or I have to adjust?

Thanks a lot for your feedbacks.

enitrat:
@glihm this is great! Is there an easy way to make it run locally with these scripts?

glihm:
> @glihm this is great! Is there an easy way to make it run locally with these scripts?

@enitrat very good point, I just pushed a commit integrating an iso CI script to ensure CI behavior reproducible locally and programs verified before pushing into the repo. :+1:

.github/workflows/verify-cairo-programs/action.yml
```diff
@@ -0,0 +1,24 @@
...
+  steps:
+    - name: Copy Cairo program verifier script
+      shell: bash
+      run: |
+        cp mdbook-cairo/scripts/cairo_programs_verifier.sh book/cairo/cairo-programs/
+
+    - name: Verify Cairo programs
+      uses: addnab/docker-run-action@v3
+      with:
+        image: starknet/cairo:1.0.0-alpha.7
```
enitrat, line 16: can you update this image to `cairo:v1.0.0` ? 

After this, we can merge. Amazing work 🔥

enitrat:
also, can you add to the readme that we have to install `mdbook-cairo`  with `cargo install --path mdbook-cairo`?

glihm:
> 

Nice catch, I'll add this. Thanks for the feedback. I think it's better to add this in the script, with the `--force` to ensure new install anytime it's run to get the modifications, if any.

glihm:
@enitrat all pushed. I also modified the cairo version in the book. Let me know if it's ok or if you want me to remove this last commit.

enitrat (approved):
Amazing PR. TYVM !

enitrat:
Closes #105

enitrat:
Closes #139

PR #157, opened 2023-05-19
Update i18n: Spanish Version es.po

Nadai2010:
I have added new chapters, as well as corrections that have been made, it is very good flow since any changes in `poedit` also indicate these changes to be reviewed.


enitrat (approved):
💯

PR #181, opened 2023-05-31
i18n Update Spanish (Appendix E)

Nadai2010:
Hi team, sorry close the PR #172 as I uploaded commit too much.

As a note, when uploading only the es.po file, I don't know if after the last Merge, I have seen now even more improvement in the syntax, the changes and text are no longer some cut off, from "Poedit" now in some cases self improved.

I have tested in local before uploading and everything is ok but. What was the reason for this? Thanks

#: src/ch00-01-foreword.md:9
msgid ""
"However, Cairo is not only for blockchain developers. As a general purpose "
"programming language, it can be used for any computation that would benefit "
"from being proved on one computer and verified on other machines with lower "
"hardware requirements."

VS

"However, Cairo is not only for blockchain developers. As a general purpose programming language, it can be used for any computation that would benefit from being proved on one "
"computer and verified on other machines with lower hardware requirements."
msgstr ""

Spanish equal

"Sin embargo, Cairo no es solo para desarrolladores de blockchain. Como "
"lenguaje de programación de propósito general, se puede utilizar para "
"cualquier cálculo que se beneficie de ser demostrado en una computadora y "
"verificado en otras máquinas con requisitos de hardware más bajos."

VS

"Sin embargo, Cairo no es solo para desarrolladores de blockchain. Como lenguaje de programación de propósito general, se puede utilizar para cualquier cálculo que se beneficie de ser "
"demostrado en una computadora y verificado en otras máquinas con requisitos de hardware más bajos."


enitrat (approved):
I don't really know what caused this format change 🤔

Nadai2010:
I will check in the next translation, from Poedit these changes were not made, but it is true that there are fragments of documents that were like this, or interpreted like this with separations, but the flow is correct in terms of new changes.

PR #184, opened 2023-06-01
Fix: Some typos in the appendix and solved 1 todo.

cryptonerdcn:
As title.


enitrat (approved):
TYVM, i missed these ones!

PR #198, opened 2023-06-05
Bash for nondeveloper

cryptonerdcn:
Add a bash for non-developer, who just wants to generate a new `.po` file for translation and doesn't want to set up the environment.
This bash script automatically installs only what's necessary to generate a `.po` file from `message.pot` for a new language, and subsequently runs the `msginit` command to generate that file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,11 @@ To work with translations, those are the steps to update the translated content:
 
 The translation work is inspired from [Comprehensive Rust repository](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md).
 
+#### For non-developer translator
+If you can't set up the environment independently (or don't want to do it), and wish to initiate a new translation for your language, consider the following options:
+- Just ask your progammer friend generate a `.po` file of your language for you.
+- Otherwise, execute the command `./new_language.sh xx` (replace `xx` with your language code). This method can generate a `.po` file and almost doesn't need you install anything by yourself. Only for Linux and MacOS right now.
+
 ### Work locally (Cairo programs verification)
 
 The current book has a mdbook backend to extract Cairo programs from the markdown sources. Currently, for each program it test two things: if it compiles and if it adheres to the `cairo-format` coding style. You can run this locally and test if a Cairo program you have written in the book passes these tests.
--- a/new_language.sh
+++ b/new_language.sh
@@ -1,0 +1,41 @@
+LANG=$1
+FILE=po/$LANG.po
+OS=$(uname)
+
+# Stop if no laguage input.
+if [ $# -eq 0 ]
+then
+    echo ""
+    echo "No input language, exit."
+    exit 0
+fi
+
+#Linux
+if [ "$OS" == "Linux" ]; then
+    ARCH=$(uname -m)
+    # Install gettext
+    sudo apt install gettext
+    # Download mdBook and extract it out
+    if [ "$ARCH" == "aarch64" ]; then
+        curl -L https://github.com/rust-lang/mdBook/releases/download/v0.4.30/mdbook-v0.4.30-aarch64-unknown-linux-musl.tar.gz | tar -xz
+    else
+        curl -L https://github.com/rust-lang/mdBook/releases/download/v0.4.30/mdbook-v0.4.30-x86_64-unknown-linux-musl.tar.gz | tar -xz
+    fi
+# MACOS
+elif [ "$OS" == "Darwin" ]; then
+    # Install gettext
+    brew install gettext
+    # Download mdBook and extract it out
+    curl -L https://github.com/rust-lang/mdBook/releases/download/v0.4.30/mdbook-v0.4.30-x86_64-apple-darwin.tar.gz | tar -xz
+else
+    echo "The operating system is not Linux or macOS."
+    exit 0
+fi
+
+# Build a new LANGUAGES. po file if not exist .
+if test -f "$FILE"; then
+    echo ""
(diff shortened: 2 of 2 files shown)


cryptonerdcn:
Inspire by @MathiasTELITSINE. He asked me if there is a way to only generate the `.po` file without all those programmers' stuff.

cryptonerdcn:
Hi @glihm can you review this PR? This bash is based on your `translations.sh`.

glihm:
Thanks for this initiative @cryptonerdcn!

Here are my comments:

1. I think installing stuff is not "developer" related only. Because if we want your script being very generic, we must then install `cargo` too as we need to install the `mdbook-cairo` backend. Without it, even if you download `mdbook`, it is not running. In this case it will not only be related to translations, and it's a script like `rustup` that we need for the `cairo-book` (which can be great).

2. If we take apart this instalation stuff (or if we do a `cairobookup` script), it could be interesting that you add a command like `translation.sh new xx` instead of a new script, to abstract the `msginit`. What do you think?

3. If we are going your way with "on-the-fly" install of `mdbook`, I suggest to put this executable in the `.gitignore` to not confuse the user at the moment of the commit.

4. About the URLs you use for `mdbook`, you can refactor a having a `MDBOOK_URL` at the top, easier to maintain if the version has to change.

This is my take on the technical aspect, let's see @enitrat, @fricoben and/or others opinion on that.

But as you mentioned, it's definitely a good approach to add some abstraction for translators, but in my opinion the cursor may be adjusted and concerns well separated. :+1:

cryptonerdcn:
> Thanks for this initiative @cryptonerdcn!
> 
> Here are my comments:
> 
> 1. I think installing stuff is not "developer" related only. Because if we want your script being very generic, we must then install `cargo` too as we need to install the `mdbook-cairo` backend. Without it, even if you download `mdbook`, it is not running. In this case it will not only be related to translations, and it's a script like `rustup` that we need for the `cairo-book` (which can be great).
Thank you @glihm !  You are right I forgot `cargo` .

> 2. If we take apart this instalation stuff (or if we do a `cairobookup` script), it could be interesting that you add a command like `translation.sh new xx` instead of a new script, to abstract the `msginit`. What do you think?

Yes, it is a better way.  I just want to avoid modifying 'translations.sh' before but now you say we can change it so :p. 

> 4. About the URLs you use for `mdbook`, you can refactor a having a `MDBOOK_URL` at the top, easier to maintain if the version has to change.
Agree.

> 
> This is my take on the technical aspect, let's see @enitrat, @fricoben and/or others opinion on that.
> 
> But as you mentioned, it's definitely a good approach to add some abstraction for translators, but in my opinion the cursor may be adjusted and concerns well separated. 👍

Yeah, let us discuss and then find a good way to accomplish this task.

enitrat:
Here's my take on this: 

While it's great to have non-developer that can contribute to this book by translating content, I think that *at least* some familiarity with Cairo is required - that way, we're sure that the translator fully understands what he's translating. 

In that case, I expect translators to have a minimum of experience in using the CLI, setting up environments, installing dependencies, etc. 

I think it's a great idea to have this script available to make onboarding easier, however developing something like `cairobookup` seems a bit overkill. It's probably easier to just create an issue and ask to generate a translation file - but if the script helps and works, let's include it!

cryptonerdcn:
@glihm @enitrat I finished it, can you review my new update? Thanks!

translations.sh
```diff
@@ -1,28 +1,67 @@
...
+        echo "No input language, stop after build."
+        exit 0
+    fi
+
+    # Build a new LANGUAGE .po file if not exist .
+    if test -f "$FILE"; then
+        echo ""
+        echo "File $FILE already exists. Ensure you're initiating a new translation. Alternatively, use 'translations.sh' to serve an existing one."
+    else
+        msginit -i po/messages.pot -l $LANG -o po/$LANG.po
```
glihm, line 52: Quote variable "$LANG" and po/"$LANG".po

translations.sh
```diff
@@ -1,28 +1,67 @@
...
+
+    # Serve the input language, if available.
+    MDBOOK_BOOK__LANGUAGE="$LANG" mdbook serve -d book/"$LANG"
+}
 
-# Serve the input language, if available.
-MDBOOK_BOOK__LANGUAGE="$LANG" mdbook serve -d book/"$LANG"
+function build_new_language() {
+    LANG=$1
+    FILE=po/$LANG.po
```
glihm, line 37: Quote variable po/"$LANG".po

translations.sh
```diff
@@ -1,28 +1,67 @@
...
+    then
+        echo ""
+        echo "No input language, stop after build."
+        exit 0
+    fi
+
+    # Build a new LANGUAGE .po file if not exist .
+    if test -f "$FILE"; then
+        echo ""
+        echo "File $FILE already exists. Ensure you're initiating a new translation. Alternatively, use 'translations.sh' to serve an existing one."
```
glihm, line 50: we can give the full command as we have the arguments for.

"Alternatively, use './transaction.sh $LANG' to serve the existing one."

glihm (changes requested):
Great work defining functions and reworking this!
I left some comments about English wording + small details on the script, let me know if you have questions/remarks.

cryptonerdcn:
@glihm  Thanks for your review! I fixed them.

glihm:
> @glihm Thanks for your review! I fixed them.

@cryptonerdcn don't hesitate to mark the review items as "resolved". :)

@enitrat look good to me.

glihm (approved):
Thanks!

enitrat (approved):
Looks good. Thanks, @cryptonerdcn!

PR #209, opened 2023-06-07
Feat: backend improvements (Compilation, execution, tests, Starknet contracts and format)

### LINKED ISSUE #189: feat: backend should also run snippets with tests
Currently, our backend only verifies the execution of snippets with a `main` function. We should add a feature that also runs `cairo-test` on testable code.

julio4:
close #189

List of changes:
- Automatically extract Cairo code blocks
- Detect Starknet smart contracts and verify with starknet-compile
- Detect Cairo tests and verify with cairo-test
- Detect runnable Cairo code and verify with cairo-run
- Ensure all other Cairo code is compilable
- Add new tag `does_not_run`
- Add new tag `test_fails`
- Add in-progress tests visualizer for local verifier bash script
- Add clickable path to open the extracted code block (for compatible CLI/IDE)
- Improve tests summary and outputs

Todo:
- [ ] Add README documentation
- [ ] Fix detected errors

### DIFF
--- a/mdbook-cairo/scripts/cairo_local_verify.sh
+++ b/mdbook-cairo/scripts/cairo_local_verify.sh
@@ -33,4 +33,3 @@ sudo docker run --rm \
      --entrypoint sh \
      "$STARKNET_CAIRO_IMAGE" \
      /cairo/cairo_programs_verifier.sh true
-
--- a/mdbook-cairo/src/main.rs
+++ b/mdbook-cairo/src/main.rs
@@ -150,8 +150,10 @@ fn process_chapter(output_dir: &Path, prefix: &str, content: &str) {
                     let is_contract = text.contains(CODE_BLOCK_IS_CONTRACT);
 
                     if is_contract {
-                        // output to contracts directory
-                        write_to_file(&output_dir.join(CONTRACTS_DIR), prefix, &text, program_counter)
+                        if !tag_does_not_compile {
+                            // output to contracts directory
+                            write_to_file(&output_dir.join(CONTRACTS_DIR), prefix, &text, program_counter)
+                        }
                     } else {
                         let should_be_runnable = text.contains(CODE_BLOCK_IS_RUNNABLE);
                         if !tag_does_not_run && should_be_runnable {
--- a/mdbook-cairo/src/main.rs
+++ b/mdbook-cairo/src/main.rs
@@ -11,17 +11,30 @@ use std::path::Path;
 /// The table header expected in book.toml.
 const CAIRO_CONFIG_TABLE_HEADER: &str = "output.cairo";
 
-/// An attribute added to a code block tag to ignore
-/// the code extraction.
-const CODE_BLOCK_DOES_NOT_COMPILE: &str = "does_not_compile";
+/// An attribute added to a code block tag to ignore the cairo-compile check.
+const TAG_DOES_NOT_COMPILE: &str = "does_not_compile";
+/// An attribute added to a code block tag to ignore the cairo-run check.
+const TAG_DOES_NOT_RUN: &str = "does_not_run";
+/// An attribute added to a code block tag to ignore the cairo-format check.
+const TAG_IGNORE_FORMAT: &str = "ignore_format";
+/// An attribute added to a code block tag to ignore the cairo-format check.
+const TAG_FAILING_TESTS: &str = "test_fails";
+// we assume that every rust code block is a cairo block
+const CAIRO_TAG: &str = "rust";
 
-/// An attribute added to a code block tag to ignore
-/// the format check.
-const CODE_BLOCK_IGNORE_FORMAT: &str = "ignore_format";
+/// Expected statement in a code block for it to be runnable.
+const CODE_BLOCK_IS_RUNNABLE: &str = "fn main()";
+/// Expected statement in a code block for it to be runnable.
+const CODE_BLOCK_IS_CONTRACT: &str = "#[contract]";
+/// Expected statement in a code block containing tests.
+const CODE_BLOCK_IS_TESTABLE: &str = "#[test]";
 
-/// Main function expected in a code block to be a candidate
-/// for the code extraction.
-const CODE_BLOCK_MAIN_FUNCTION: &str = "fn main()";
+/// Configurations for the cairo backend.
+const COMPILABLE_DIR: &str = "compilable";
+const RUNNABLE_DIR: &str = "runnable";
+const TESTABLE_DIR: &str = "testable";
+const CONTRACTS_DIR: &str = "contracts";
+const FORMATS_DIR: &str = "formats";
 
 /// Struct mapping fields expected in [output.cairo] from book.toml.
 #[derive(Debug, Default, Serialize, Deserialize)]
@@ -30,7 +43,7 @@ pub struct CairoConfig {
     pub output_dir: String,
 }
 
-// Statically initialize the regex to avoid rebuiling at each loop iteration.
+// Statically initialize the regex to avoid rebuilding at each loop iteration.
 lazy_static! {
     static ref REGEX: Regex =
         Regex::new(r"^ch(\d{2})-(\d{2})-(.*)$").expect("Failed to create regex");
@@ -66,6 +79,15 @@ fn main() {
         .unwrap_or_else(|_| println!("Couldn't clean output directory, skip."));
 
     create_dir_all(output_path).expect("Couldn't create output directory.");
+
+    // Create subdirectories for different types of code blocks
+    let subdirectories = [COMPILABLE_DIR, RUNNABLE_DIR, TESTABLE_DIR, CONTRACTS_DIR, FORMATS_DIR];
+    subdirectories
+        .iter()
+        .for_each(|subdir| {
+            create_dir_all(output_path.join(Path::new(subdir)))
+            .expect("Couldn't create output directory.");
+        });
 
     process_chapters(&ctx.book, &output_path);
 }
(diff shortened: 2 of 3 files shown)


enitrat:
Hey @glihm, could you please take a look at this as well? Thanks!

enitrat:
Ok, now I realize that verifying the compilation of each code snippet is maybe too ambitious 😅

glihm:
About the code, I really like your approach @julio4! Nice to have such organization in the output and it will looks great.

Also, that's nice to have a distinction as you did between compilable and runnable.
Which allow valid cairo code to be checked with the compiler, even if we know that
it will panic.

@enitrat and @julio4 we also have to keep in mind that compiling a code like
```rust
'Hello World'.print();
```
requires extra cairo syntax to be added (a function at least, even if it's not a main), which is fine, but not added in the current PR. This is why most of the tests are failing.

If we add the code to compile even a single line of cairo which is valid, I think we can keep the current behavior: every written cairo code should at least compile, except if annotated `does_not_compile`.

mdbook-cairo/src/main.rs
```diff
@@ -11,17 +11,30 @@ use std::path::Path;
...
-/// Main function expected in a code block to be a candidate
-/// for the code extraction.
-const CODE_BLOCK_MAIN_FUNCTION: &str = "fn main()";
+/// An attribute added to a code block tag to ignore the cairo-compile check.
+const TAG_DOES_NOT_COMPILE: &str = "does_not_compile";
+/// An attribute added to a code block tag to ignore the cairo-run check.
+const TAG_DOES_NOT_RUN: &str = "does_not_run";
+/// An attribute added to a code block tag to ignore the cairo-format check.
+const TAG_IGNORE_FORMAT: &str = "ignore_format";
+/// An attribute added to a code block tag to ignore the cairo-format check.
```
glihm, line 20: Copy paste with edition missing.

mdbook-cairo/src/main.rs
```diff
@@ -103,46 +125,80 @@ fn process_chapter(output_dir: &Path, prefix: &str, content: &str) {
...
-
-                        let file_dir = &output_dir.join(file_name);
-                        let mut file = File::create(file_dir).expect("Failed to create file.");
-
-                        file.write(text.as_bytes()).expect("Can't write to file.");
+                if is_cairo_block {
+                    let is_contract = text.contains(CODE_BLOCK_IS_CONTRACT);
+
+                    if is_contract {
+                        if !tag_does_not_compile {
```
glihm, line 153: A contract may also have the `does_not_run` tag if we know there is a panic at some point.

mdbook-cairo/src/main.rs
```diff
@@ -103,46 +125,80 @@ fn process_chapter(output_dir: &Path, prefix: &str, content: &str) {
...
+                            write_to_file(&output_dir.join(CONTRACTS_DIR), prefix, &text, program_counter)
+                        }
+                    } else {
+                        let should_be_runnable = text.contains(CODE_BLOCK_IS_RUNNABLE);
+                        if !tag_does_not_run && should_be_runnable {
+                            // output to runnable directory
+                            write_to_file(&output_dir.join(RUNNABLE_DIR), prefix, &text, program_counter)
+                        } else if !tag_does_not_compile {
+                            // output to compilable directory
+                            write_to_file(&output_dir.join(COMPILABLE_DIR), prefix, &text, program_counter)
```
glihm, line 164: Here we need an other case.

If the cairo code block `does_not_compile`, and `!should_be_runnable`, we must insert the cairo code into a main function.

Build a string with the following syntax should be enough:

```
fn main() {
    {text}
}
```

We may also consider adding very generic imports like the debug etc..
@enitrat we have to check that part, as we may not be able to auto-import all the required stuff only based on a quick code analysis.

mdbook-cairo/src/main.rs
```diff
@@ -11,17 +11,30 @@ use std::path::Path;
...
-/// for the code extraction.
-const CODE_BLOCK_MAIN_FUNCTION: &str = "fn main()";
+/// An attribute added to a code block tag to ignore the cairo-compile check.
+const TAG_DOES_NOT_COMPILE: &str = "does_not_compile";
+/// An attribute added to a code block tag to ignore the cairo-run check.
+const TAG_DOES_NOT_RUN: &str = "does_not_run";
+/// An attribute added to a code block tag to ignore the cairo-format check.
+const TAG_IGNORE_FORMAT: &str = "ignore_format";
+/// An attribute added to a code block tag to ignore the cairo-format check.
+const TAG_FAILING_TESTS: &str = "test_fails";
```
glihm, line 21: Not sure about this one.

When someone writes test, usually he takes care of using `#[should_panic()]` when the test is expected to fail.

But I may be missing something, what do you have in mind?

mdbook-cairo/scripts/cairo_local_verify.sh
```diff
@@ -32,5 +32,4 @@ sudo docker run --rm \
      -v "$(pwd)"/book/cairo/cairo-programs:/cairo \
      --entrypoint sh \
      "$STARKNET_CAIRO_IMAGE" \
-     /cairo/cairo_programs_verifier.sh true
-
+     /cairo/cairo_programs_verifier.sh true
```
glihm, line 35: Try to always keep a blank line at the end of files. I have perhaps missed some, sorry.

glihm (commented):
Great work @julio4!

We must check with @enitrat about the exact requirement. I don't have yet the sufficient vision about testing single snippets that are not in a function as we may miss some imports.

Please let me know your take on the comments, and let's continue this nice work!

mdbook-cairo/src/main.rs
julio4: Can we test contract execution? I though that `starknet-compile` only compile the contract.

mdbook-cairo/src/main.rs
julio4: That is a great idea! Another potential solution could be to add "hidden" Cairo code that will not show up in the book but can still be extracted. This may require more work from contributors, but we won't have to worry about imports. Check out [hiding-code-lines](https://rust-lang.github.io/mdBook/format/mdbook.html?highlight=hidden#hiding-code-lines).

@enitrat What do you think of this?

mdbook-cairo/src/main.rs
julio4: We can detect `#[should_panic()]` when there are tests and avoid running `cairo-test` on them. However, we may want to keep the `test_fails` tag in the rare case that we want to show test features that don't pass, but don't want to use `#[should_panic()]`.

julio4:
> Maybe we should go with the opposite, actually: assume they're not compilable by default, and mark the ones that should compile correctly with `should_compile`.

We can inject the code into a main function, or manually add the function as hidden lines in the mdbook. Alternatively, we can use a mix of both by adding a "`manual`" tag to not inject the main function and specify which imports are needed in the hidden lines. If we still have too much fails with this, we can go with the `should_compile` tag.

mdbook-cairo/src/main.rs
enitrat: Indeed, you can only compile or test contracts, not run them.

mdbook-cairo/src/main.rs
enitrat: That's an amazing feature, I didn't know about this. Let's use this in the book!

mdbook-cairo/src/main.rs
enitrat: > usually he takes care of using #[should_panic()] when the test is expected to fail.

Not really, we use `#[should_panic()] ` when the _execution_ is expected to fail. But the test itself succeeds.

enitrat:
Let's keep things simple and not too abstracted to keep the onboarding of new contributions simple. I think we should go with the hidden lines way

julio4:
I agree! I will need to check if the hidden lines method really works, and then let's do it that way!

It might be a bit of work to add the missing functions to all current code blocks. Additionally, it will add some complexity for new contributors, so we will need to provide documentation for this.

enitrat:
> It might be a bit of work to add the missing functions to all current code blocks. Additionally, it will add some complexity for new contributors, so we will need to provide documentation for this.

We can do this in multiple PRs one chapter at a time. 
While we're at it, we can probably refactor the structure of the book so that listings are not longer directly in `.md` files, but instead in a separate, listings folder. This helps when working on code examples, as we can format and run these cairo files.

Let's keep this PR open for now until we refactored the structure of the book.

mdbook-cairo/src/main.rs
glihm: Definitely the best approach, excellent finding @julio4!

enitrat:
Now that the book structure has been updated, I think we can re-think how the verifier works. 
Here's my proposal
1. mdbook-cairo is no longer required to extract code snippets, as code snippets are `.cairo` files under the `listings` directory
2. However, we need a way to specify what action to perform on each file: is it `cairo-compile`, `cairo-run`, `cairo-test`? We can indicate this with a special tag at the beginning of each cairo-file as a comment
3. If we have a special tag as comment, we would still need our `mdbook-cairo` backend in order to process these files and remove the tags associated to the verification of programs. In the same way, `mdbook` removes all lines related to `ANCHOR` when building the HTML 

WDYT? :)

julio4:
Following our discussion in Telegram, I think it would be better to completely remove the mdbook-cairo backend. Instead, we can use a bash script or a small Rust CLI app to recursively explore all `.cairo` files under `listings` and run tests according to the metadata/tags (potentially as cairo comments so we don't have to worry about altering the file content before running tests). We can reuse a lot of what has already been done.

> mdbook-cairo is no longer required to extract code snippets
> If we have a special tag as comment, we would still need our mdbook-cairo backend in order to process these files and remove the tags associated to the verification of programs.

The `{{#include ...}}` statement is a mdBook [preprocessor](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html), so the code is included before being sent to the backend. We can follow a similar approach and add our own [custom preprocessor](https://rust-lang.github.io/mdBook/for_developers/preprocessors.html) that only removes tags from cairo code blocks. Hopefully, we can add this preprocessor after the `include` one so that we can easily modify the included text.

glihm:
As you both said the concerns are a bit different now.

1. We can define that the tags are always at the very first line of the file, comma separated. The very first line must always be a comment then, even if no tag is added.

2. We can process all the files in listing easily based on the tags. I like the rust CLI tool @julio4 more than bash. :) as we may extend it as a standalone reusable tool.

3. We can leverage mdbook line range and say to contributor that they always have to skip the first line.

What do you thing @enitrat @julio4?

enitrat:
> We can leverage mdbook line range and say to contributor that they always have to skip the first line.

Not a huge fan of this, ideally the preprocessor only hides the first line if it's a tag about the compilation/run status

I think the basic status should be to run `cairo-format` and  `cairo-compile` or `starknet-compile` instead if it's a contract. If the file contains tests, it should run `cairo-test (--starknet, if required)` as well

Examples:
```rust
//CAIRO: does_not_compile
fn foo(x:felt252, y){
}
```
Only apply the `cairo-format --check` on the above file

```rust
//CAIRO: ignore_fmt
fn main(){
        // a file we don't want to check format, but we still run it
}
```
Only run `cairo-run` on the above file

```rust
#[contract]
mod MyContract{
  struct Storage{}
}
```
Run `starknet-compile` on the above file

```rust
fn foo(x:felt252) -> felt252{
    x
}

#[test]
#[available_gas(200000)]
fn test_foo(){
    assert(foo(1)==1, 'wrong result foo')
}
```
Run both `cairo-compile` and `cairo-test` on this file

julio4:
> Not a huge fan of this, ideally the preprocessor only hides the first line if it's a tag about the compilation/run status
Agreed, it should provide a seamless experience for contributors.

I will begin working on it directly in this PR. Please feel free to add any suggestions.

enitrat:
> I will begin working on it directly in this PR. Please feel free to add any suggestions.

Amazing! Feel free to close this PR and open a new one if it makes things easier. You can follow the spec above

PR #210, opened 2023-06-08
refactor(ch02): move code snippets to cairo file 

enitrat:
Moves code snippets to Cairo files.

This makes maintenance of the book easier, as
- We can use the Cairo formatted on Cairo files instead of writing code in markdown code blocks
- We can hide irrelevant code parts from the displayed code block


enitrat:
@glihm @julio4 following our discussion in #209, here's a first idea of how we can make maintenance simpler by moving code snippets to cairo files. This will make maintenance a lot easier, as we get editor features when working on cairo code.

glihm:
> @glihm @julio4 following our discussion in #209, here's a first idea of how we can make maintenance simpler by moving code snippets to cairo files. This will make maintenance a lot easier, as we get editor features when working on cairo code.

That's amazing @enitrat...! :rocket:
Definitely the best way to go, combining this with hidden lines... It will be awesome.

What's the next move then, what do we change first to continue on this topic with @julio4 and I?

enitrat:
> What's the next move then, what do we change first to continue on this topic with @julio4 and I?

I'm not sure yet, I guess that we can keep running the `cairo-run`, `starknet-compile` and `cairo-test` backends, probably `cairo-format` as well, but `cairo-compile` might cause more harm than good

enitrat:
@glihm @julio4 I'm going to disable format checks for now, because including partial code snippets messes up with the formatter. I think it's less important now that we have Cairo code in files.

fricoben (approved):
lgtm

enitrat:
noooo @fricoben it was not passing the checks 😢

PR #219, opened 2023-06-10
refactor(ch05): code blocks

enitrat:


listings/ch05-enums-and-pattern-matching/no_listing_02_enum_message.cairo
```diff
@@ -0,0 +1,41 @@
...
+            },
+        }
+    }
+}
+// ANCHOR_END: trait_impl
+fn main() {
+    // ANCHOR: main
+    let msg: Message = Message::Quit(());
+    msg.process();
+// ANCHOR_END: main
```
cryptonerdcn, line 37: Maybe need to be formatted in the future.

src/ch05-02-the-match-control-flow-construct.md
```diff
@@ -9,21 +9,7 @@ Think of a match expression as being like a coin-sorting machine: coins slide do
...
-
-fn value_in_cents(coin: Coin) -> felt252 {
-    match coin {
-        Coin::Penny(_) => 1,
-        Coin::Nickel(_) => 5,
-        Coin::Dime(_) => 10,
-        Coin::Quarter(_) => 25,
-    }
-}
+{{#include ../listings/ch05-enums-and-pattern-matching/listing_05_03.cairo:all}}
```
cryptonerdcn, line 12: Just a question: Why do we need an all-anchor here? Because we used some other anchors in the `listing_05_03.cairo`, we can not just use `#include ../listings/ch05-enums-and-pattern-matching/listing_05_03.cairo`?

cryptonerdcn (approved):
LGTM

src/ch05-02-the-match-control-flow-construct.md
enitrat: If we don't use an anchor, the `//ANCHOR` comments are shown in the code block. That's why I define an ANCHOR: all, so that I can include the whole file but without it showing the anchor comments

PR #223, opened 2023-06-12
Fixed dispatcher naming and ownership

tiagofneto:
Currently, the documentation indicates that the `IERC20Dispatcher` struct implements a method called `get_name` (`Listing 99-2`), while `Listing 99-3` suggests that a method called `name` should be implemented instead. The same discrepancy is observed in `LibraryDispatcher` (`Listing 99-4`).

Furthermore, it's expected that the dispatcher structs instances can be used multiple times, indicating that the implemented methods should not take ownership. However, `Listing 99-2`and `Listing 99-4` suggests that these methods take ownership, which would prohibit their multiple uses. To address this, I propose that the methods should be revised to receive a snapshot as a parameter, thereby not taking ownership.

### DIFF
--- a/listings/ch99-starknet-smart-contracts/listing_99_05_dispatcher_trait.cairo
+++ b/listings/ch99-starknet-smart-contracts/listing_99_05_dispatcher_trait.cairo
@@ -2,7 +2,7 @@
 use starknet::{ContractAddress};
 
 trait IERC20DispatcherTrait<T> {
-    fn get_name(self: T) -> felt252;
+    fn name(self: T) -> felt252;
     fn transfer(self: T, recipient: ContractAddress, amount: u256);
 }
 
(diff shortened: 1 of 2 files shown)


tiagofneto:
After looking at the [dispatcher implementation](https://github.com/starkware-libs/cairo/blob/254539541ff976616db22ac87c82b6371ce9fd7d/crates/cairo-lang-starknet/src/plugin/dispatcher.rs#L175), I found that the dispatcher structs derive the `Copy` trait. This means that they inherently do not require a snapshot as a parameter, making the proposed change unnecessary in this context.

Moreover, I noticed that in addition to deriving the `Copy` and `Drop` traits (as already shown in `Listing 99-2` and `Listing 99-4`), the dispatcher structs also derive `storage_access::StorageAccess` and `Serde`. While it might be worth considering the inclusion of these details in the documentation for completeness, I think that this could potentially introduce confusion due to their detailed nature and thus, may be best left as an implementation detail.

The initial issue regarding the naming inconsistency between `get_name` and `name` in `IERC20Dispatcher` and `IERC20LibraryDispatcher` still stands.

enitrat:
Hey, thanks for spotting this! We're trying to make the book a bit more robust and changing some internal structure. I'm going to merge #228 first and then apply your changes.

Regarding `Serde` and `StorageAccess`, they were not derived when this chapter was written - thanks for spotting that. I'm waiting for the next compiler release for a complete refactor of the Starknet contract section, but we can already add them now

PR #227, opened 2023-06-12
refactor: listings ch09 

enitrat:


src/ch09-01-contract-syntaxes.md
cryptonerdcn, line 1: Just want to make sure: This chapter has been moved to https://cairo-book.github.io/ch99-01-02-writing-starknet-contracts.html now. Right?

src/ch09-01-unrecoverable-errors-with-panic.md
```diff
@@ -48,22 +38,18 @@ You can use the `nopanic` notation to indicate that a function will never panic.
...
-fn function_never_panic() nopanic {
-    assert(1 == 1, 'what');
-}
+{{#include ../listings/ch09-error-handling/no_listing_04_nopanic_wrong.cairo}}
 ```
 
 If you write the following function that includes a function that may panic you will get the following error:
 
-```bash
+```console
```
cryptonerdcn, line 52: Why do we need to change `bash` to `console` here?
Also, I do not find mdbook supports `console`, can you give me more details about that?
https://rust-lang.github.io/mdBook/format/theme/syntax-highlighting.html?highlight=bash#supported-languages

src/ch09-01-contract-syntaxes.md
enitrat: yes, it was no longer referenced in the index but the file was still here

cryptonerdcn (approved):
LGTM

PR #228, opened 2023-06-12
refactor: listings ch99

enitrat:


cryptonerdcn (approved):
LGTM

listings/ch99-starknet-smart-contracts/listing_99_03.cairo
cryptonerdcn, line 1: Empty file and has never been used.

Issue #230, opened 2023-06-12
Add support for && and || operators.


enitrat:
Replace usages of `&` and `|` by `&&` and `||`

Add operators to [Appendix 2](https://cairo-book.github.io/appendix-02-operators-and-symbols.html)

Estimated time: 30mins

enitrat:
@0x471 can you finish this over the weekend?

0x471:
> @0x471 can you finish this over the weekend? otherwise I'll take it, it's just one line to add to the appendix

I apologize, but I'm unable to complete it due to my schedule

enitrat:
sure, no problem!

PR #238, opened 2023-06-15
Add support for && and || operators 

0x471:
#230

### DIFF
--- a/src/appendix-02-operators-and-symbols.md
+++ b/src/appendix-02-operators-and-symbols.md
@@ -14,7 +14,7 @@ Table B-1 contains the operators in Cairo, an example of how the operator would 
 | `!=` | `expr != expr` | Non-equality comparison | `PartialEq` |
 | `%` | `expr % expr` | Arithmetic remainder | `Rem` |
 | `%=` | `var %= expr` | Arithmetic remainder and assignment | `RemEq` |
-| `&` | `expr & expr` | Bitwise AND | `BitAnd` |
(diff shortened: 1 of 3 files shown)


src/appendix-02-operators-and-symbols.md
```diff
@@ -14,7 +14,7 @@ Table B-1 contains the operators in Cairo, an example of how the operator would
 | `!=` | `expr != expr` | Non-equality comparison | `PartialEq` |
 | `%` | `expr % expr` | Arithmetic remainder | `Rem` |
 | `%=` | `var %= expr` | Arithmetic remainder and assignment | `RemEq` |
-| `&` | `expr & expr` | Bitwise AND | `BitAnd` |
+| `&&` | `expr && expr` | Bitwise AND | `BitAnd` |
```
enitrat, line 17: This is not a replacement, but rather an addition: we still have the Bitwise AND, OR, XOR operators available (`&, | , ^`). But this new feature is the support for short-circuiting logical AND and OR, it should be a new line.

--


<br class="Apple-interchange-newline">

enitrat (commented):
Can you also replace all usages of `&` and `|` in Cairo files by the newly introduced logical operators instead (where it makes sense)? e.g. in
```
    fn can_hold(self: @Rectangle, other: @Rectangle) -> bool {
        *self.width > *other.width & *self.height > *other.height
    }
```

enitrat:
@0x471, any update?

PR #246, opened 2023-06-22
simplified the trait definition

### LINKED ISSUE #233: simplify the `impl` mechanism with #[generate_trait]
We can now use the `#[generate_trait]` attribute on an implementation to directly generate the trait corresponding to an implementation.
In this issue, you will demonstrate how to use it, with examples. You will cover generic and non-generic traits.
```rust
  #[generate_trait]
  impl AddMethods for AddMethodsTrait {
      fn method(self: TypeWithNewMethods) -> u32 { ... }
  }
  ```
  
  Estimated time: 30mins

ametel01:
Closes #233

### DIFF
--- a/src/ch07-02-traits-in-cairo.md
+++ b/src/ch07-02-traits-in-cairo.md
@@ -2,27 +2,15 @@
 
 Traits specify functionality blueprints that can be implemented. The blueprint specification includes a set of function signatures containing type annotations for the parameters and return value. This sets a standard to implement the specific functionality.
 
-## Defining a Trait
+## Defining and implementing a Trait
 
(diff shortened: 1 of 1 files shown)


enitrat (changes requested):
Hi @ametel01, thanks for the PR! 

`#[generate_trait]` is a useful attribute that allows us to generate the trait directly from an implementation and saves us from writing boilerplate code. However, it does not change how the trait mechanism works - You can still define traits manually, which is useful in some cases. 

Can you edit your PR and create a subsection specifically about how we can use this `#[generate_trait]` attribute to write type implementations and have the compiler define the trait for us? Ideally, you will show the code that the compiler generates.

enitrat:
Closes #233

PR #247, opened 2023-06-22
removed bracket after break statements

### LINKED ISSUE #231: refactor: simplify return and break syntax
> return; and break; are now supported.

We can now replace all `break()` by `break`, same for returns.
We will simplify the syntax used in the book to remove parenthesis now that they're no longer needed.

ametel01:
Closes #231


enitrat (changes requested):
Thank you for the PR!
The convention, in Rust and thus Cairo, is to use explicit returns when you return early from a function (i.e. inside an if/else statement).

For example, in that case:
```rust
if index >= arr.len() {
        return Option::None(());
    }
```
we want to keep the explicit return because we return early from the function.
But for the last statement of the function `return found;` we can remove the `return` keyword, as you modified

PR #249, opened 2023-06-22
Feat: cairo verify tool

### LINKED ISSUE #189: feat: backend should also run snippets with tests

julio4:
Close #189 

Following our discussion in #209, here's a first version of a simple CLI tool (named `cairo-verify`).

### Change to `mdbook-cairo`

The `mdbook-cairo` is no longer a backend. It's now a mdbook preprocessor that only removes the `// TAG` lines in code blocks.

**!! Install it again !!**
```sh
cargo install --path mdbook-cairo
```

### `cairo-verify`

The `cairo-verify` tool is designed to wrap all cairo and starknet plugins for quickly verifying cairo programs.

#### Setup

```sh
cargo run --bin cairo-verify
```

Or full install:
```sh
cargo install --path cairo-verify
```

### Usage

The tool scans for all `*.cairo` files in the specified directory and performs the following actions:

For a Starknet contract:
- `starknet-compile`
- If it has tests: `cairo-test --starknet`

Cairo program:
- If it has a `main` function: `cairo-run`
- Else, `cairo-compile`
- If it has tests: `cairo-test`
- `cairo-fmt`

To specify which tests to run, you can add a comment at the top of your file with the following format:

```rust
// TAG: <tag1>
// TAGS: <tag1>, <tag2>
```

Here is a list of available tags:
- `does_not_compile`: don't run `cairo-compile`/`starknet-compile`
- `does_not_run`: don't run `cairo-run`
- `ignore_fmt`: don't run `cairo-fmt`
- `tests_fails`: don't run `cairo-test`/`cairo-test --starknet`

You can skip and ignore a specific test by adding the corresponding flag:

```sh
$ cairo-verify --help

Usage: cairo-verify [OPTIONS]

Options:
  -p, --path <PATH>    The path to explore for *.cairo files [default: ./listings]
  -v, --verbose        Print more information
  -q, --quiet          Only print final results
  -f, --formats-skip   Skip cairo-format checks
  -s, --starknet-skip  Skip starknet-compile checks
  -c, --compile-skip   Skip cairo-compile checks
  -r, --run-skip       Skip cairo-run checks
  -t, --test-skip      Skip cairo-test checks
      --file <FILE>    Specify file to check
  -h, --help           Print help
  -V, --version        Print version
```

In CI, it's preferable to reduce output, so run `cairo-verify` with the `--quiet` flag.

CI Change:
We have to run `cairo-verify -q`, but it rely both on rust and cairo so we can't use the cairo docker image anymore.


julio4:
@enitrat @glihm, could you please provide some feedback on this?

cairo-verify/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+            format!(
+                "Total errors: {}",
+                total_errors.to_string().red()
+            )
+            .bold()
+        );
+
+        println!("\n{}", "Please review the errors above. Do not hesitate to ask for help by commenting on the issue on Github.".red().italic());
+    } else {
+        println!("\n{}\n", "ALL TESTS PASSING!".green().bold());
```
enitrat, line 77: ALL TESTS PASSED intead?

cairo-verify/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+fn process_file(file_path: &str) {
+    let cfg = &*CFG;
+    let file = File::open(file_path).expect("Failed to open file");
```
enitrat, line 83: we should add the file_path in the error message

cairo-verify/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+    // Parsed tags (if any)
+    let mut tags: HashSet<Tags> = HashSet::new();
+    let mut in_tag_block = true;
+
+    // Program info
+    let mut should_be_runnable = false;
+    let mut should_be_testable = false;
+    let mut is_contract = false;
+
+    // for line in reader.lines() {
```
enitrat, line 95: unused line

cairo-verify/src/utils.rs
```diff
@@ -0,0 +1,64 @@
...
+                cairo_files.push(entry.path().display().to_string());
+            }
+        }
+    }
+
+    cairo_files
+}
+
+/// Will replace the file path contained in the input string with a clickable format for better output
+pub fn clickable(input: &str) -> String {
```
enitrat, line 29: super cool

cairo-verify/src/cmd.rs
```diff
@@ -0,0 +1,50 @@
...
+            Cmd::StarknetTest => "cairo-test",
+            Cmd::StarknetCompile => "starknet-compile",
+        }
+    }
+
+    fn args(&self) -> Vec<&str> {
+        match self {
+            Cmd::CairoFormat => vec!["-c"],
+            Cmd::CairoCompile => vec![],
+            Cmd::CairoRun => vec!["--program"],
```
enitrat, line 28: `cairo-run` doesn't have a `--program` flag

enitrat:
Amazing PR @julio4 🙃

cairo-verify/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+        // This is a contract, it must pass starknet-compile
+        if !tags.contains(&Tags::DoesNotCompile) && !cfg.starknet_skip {
+            match Cmd::StarknetCompile.test(file_path) {
+                Ok(_) => {}
+                Err(e) => handle_error(e, file_path, Cmd::StarknetCompile),
+            }
+        }
+    } else if should_be_runnable {
+        // This is a cairo program, it must pass cairo-run
+        if !tags.contains(&Tags::DoesNotRun) && !cfg.run_skip {
```
glihm, line 131: You should also check for `DoesNotCompile`, because if I have a file with a `main` function `should_be_runnable` is `true`, and if I have a TAG `does_not_compile`, this generates a run error, which is expected as it does not compile, but reported as an error.

Or do we want the user to always write `does_not_run, does_not_compile`?

I have tested this for example on the file `no_listing_21_function_return_values_2.cairo`.

```rust
// TAG: does_not_compile                                                                                                                                                         

use debug::PrintTrait;

fn main() {
    let x = plus_one(5);

    x.print();
}

fn plus_one(x: u32) -> u32 {
    x + 1
}
```

And then running `cairo-verify -p listings/ch02-common-programming-concepts/`.

```bash
Run Errors: 37
 - no_listing_32_array_at.cairo
 - no_listing_26_if_let.cairo
 - no_listing_11_type_casting.cairo
 - no_listing_12_tuple_type.cairo
 - no_listing_35_array_span.cairo
 - no_listing_08_numeric_operations.cairo
 - no_listing_14_tuple_types.cairo
 - no_listing_21_function_return_values_2.cairo <<<<<<< present here
 - ...
```

cairo-verify/src/tags.rs
```diff
@@ -0,0 +1,20 @@
...
+    DoesNotCompile,
+    DoesNotRun,
+    IgnoreFormat,
+    FailingTests,
+}
+
+impl Tags {
+    pub fn from_str(tag: &str) -> Option<Self> {
+        match tag {
+            "does_not_compile" => Some(Tags::DoesNotCompile),
```
glihm, line 13: As mentioned above, `does_not_compile` implies `does_not_run`.

mdbook-cairo/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    fn run(&self, _ctx: &PreprocessorContext, mut book: Book) -> Result<Book, Error> {
+        book.for_each_mut(|item: &mut BookItem| {
+            if let BookItem::Chapter(ref mut chapter) = *item {
+                let mut new_content = String::new();
+                let mut lines = chapter.content.split_terminator('\n').peekable();
+                let mut in_code_block = false;
+
+                while let Some(line) = lines.next() {
+                    in_code_block = in_code_block ^ line.starts_with("```");
```
glihm, line 35: nice bitwise approach, love it.

`in_code_block ^= line.starts_with("```");` would have been more idiomatic. :+1:

mdbook-cairo/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+                    if in_code_block {
+                        if RE.is_match(line) {
```
glihm, line 38: Consider using more explicit names for static variables.

glihm (changes requested):
In general, that's an awesome PR @julio4!

Some feedbacks are in the review items, but here are more general ones:

1. The UX is awesome, the tool feels great in a terminal, very quick and nice to use.

2. You will see in my comments too, but to easily switch between cairo versions, it could be great to be able to pass the binaries folder to the tool. This ensures that users working actively on cairo with several versions installed can easily switch the version with a simple command line argument.

3. Consider applying `cargo fmt` and `cargo clippy --all -- -D warnings` inside `cairo-verify`. :+1:

Thanks again for this massive rework, loving the tool! :heart:

cairo-verify/src/cmd.rs
julio4: When using `cairo-run <program>` I have `cairo-run: error: unrecognized arguments` (cairo-run 0.11.2).
What is the correct way of using it?

cairo-verify/src/main.rs
julio4: Good catch! It's better that way.

cairo-verify/src/tags.rs
julio4: I believe that the `Tags` enum should not provide any information about tag implications. Instead, this logic should be handled in the `main` file. 
Let me know if you think it would be better to handle the tag implication logic here.

cairo-verify/src/cmd.rs
```diff
@@ -0,0 +1,50 @@
+use std::process::{Command};
+
+pub enum Cmd {
```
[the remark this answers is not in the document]
julio4: `cairo-root` is a great idea. I suggest that we first merge this initial version so we can already fix the current issues, and I'll add this soon.

cairo-verify/src/cmd.rs
enitrat: The `cairo-run` on your computer is the one from cairo0! 
You need to delete it from your path and replace it with the cairo1 one.

julio4: Well, I shouldn't be the one making installation instructions 😢 
Fixed now, sorry for that.

cairo-verify/src/tags.rs
enitrat: > Instead, this logic should be handled in the main file.

agreed

glihm: Yes I do agree it was to mention it here too. Definitely in the main. 👍🏻

enitrat:
@julio4 i'm modifying the cairo files to ensure that everything runs accordingly to how it should, and I make some modification to the rust files, don't forget to pull my changes when I push

enitrat:
feature request: if something is marked with `does_not_compile`, or `does_not_run`, then we should still run it and verify that indeed it does not compile / does not run

cairo-verify/src/cmd.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        binary.into()
+    }
+
+    fn args(&self) -> Vec<&str> {
+        match self {
+            Cmd::CairoFormat(_) => vec!["-c"],
+            Cmd::CairoCompile(_) => vec![],
+            Cmd::CairoRun(_) => vec!["--available-gas=20000000"],
+            Cmd::CairoTest(_) => vec!["--starknet"],
```
julio4, line 37: I believe you should remove `--starknet` argument for non-contracts cairo files.

enitrat: It's necessary if we want to test starknet contracts, and it doesn't change anything for pure cairo tests, so let's 
add it by default instead of having another extra condition to take care of.

julio4: In that case we should remove the `StarknetTest` variant in the `Cmd` enum.

enitrat: Ah, right, I didn't see it was there. I'll revert my changes

julio4: If `--starknet` has no side effect for pure cairo, then it's a good idea to keep things simple and remove `StarknetTest`. We just have to replace it by `CairoTest`!

julio4: If there could be additional logic only for Starknet contract in the future, then just remove the flag.

glihm:
> @glihm should we `cargo install` or `cargo run --bin` in our GH action? what's the fastest?

I usually see `cargo install` with the `--locked` as @julio4 did. :+1: It's totally fine IMHO.

cairo-verify/src/cmd.rs
```diff
@@ -1,34 +1,42 @@
...
+            Cmd::CairoFormat(path) => ("cairo-format", path),
+            Cmd::CairoCompile(path) => ("cairo-compile", path),
+            Cmd::CairoRun(path) => ("cairo-run", path),
+            Cmd::CairoTest(path) => ("cairo-test", path),
+            Cmd::StarknetTest(path) => ("cairo-test", path),
+            Cmd::StarknetCompile(path) => ("starknet-compile", path),
+        };
+
+        if let Some(r) = cairo_root {
+            let full_path = r.join(PathBuf::from("target/release")).join(binary);
```
glihm, line 23: I would remove the `target/release` as this takes the hard assumption that the user compiles it and on how the user organizes his files. When it can be a server or GH action with binaries imported/installed directly without compilation.

enitrat (approved):

PR #254, opened 2023-06-23
Removed bracket after break statements

### LINKED ISSUE #231: refactor: simplify return and break syntax

ametel01:
closes #231


enitrat (approved):
LGTM

PR #257, opened 2023-06-23
added scarb test section to docs

### LINKED ISSUE #224: refactor: use `scarb test` in test chapter
With Scarb's release 0.4, we can now use `scarb test` to test scarb projects without relying on a `cairo-project.toml` file.

However, we still need to run `cairo-test`for individual files

ametel01:
closes #224


src/ch01-03-hello-scarb.md
```diff
@@ -151,11 +151,43 @@ Run completed successfully, returning []
 
 Using `scarb run` is a convenient way to run custom shell scripts that can be useful to run files and test your project.
 
+### Running tests
+
+To execute all the unit and integration tests for a given package, use the `scarb test` command. This command requires a specific directory structure, which includes creating a tests directory. Within this directory, a corresponding `lib.cairo` file must be present that lists each test file.
+
+```rust
+tests/
```
enitrat, line 159: isn't it supposed to be under `src/` instead?

src/ch01-03-hello-scarb.md
```diff
@@ -151,11 +151,43 @@ Run completed successfully, returning []
...
+
+
+In order to execute tests for Starknet contracts, it is necessary to append a specific flag to the scarb command, which is `--starknet`. The command would thus appear as `scarb test --starknet.` This action activates the Starknet plugin for the duration of our tests within the software environment.
+
 Let’s recap what we’ve learned so far about Scarb:
 
 - We can create a project using `scarb new`.
 - We can build a project using `scarb build` to generate the compiled Sierra code.
 - We can define custom scripts in `Scarb.toml` and call them with the `scarb run` command.
+- We can run test utilizing either `scarb test` and `scarb test --starknet` commands.
```
enitrat, line 190: `scarb test --starknet` is not valid. There's only `scarb test` afaik (but starknet is activated by default)

src/ch01-03-hello-scarb.md
ametel01: I’m using scarb 0.5 which has Cairo v2 compiler and this works on my repos. Having a tests folder under /src return error for missing ../test/lib.cairo
Maybe I should change the scarb version to 0.5?

src/ch01-03-hello-scarb.md
ametel01: Same as the above with scarb >= 0.5 starknet contracts will not compile without the flag

src/ch01-03-hello-scarb.md
enitrat: Ah, right I'm using 0.4, let me see with 0.5

src/ch01-03-hello-scarb.md
```diff
@@ -31,7 +31,7 @@ As for now, Scarb needs manual installation with the following steps:
 
   ```bash
   $ scarb --version
-  scarb 0.3.0 (182a9019d 2023-05-29)
+  scarb 0.4.0 (182a9019d 2023-06-06)
```
enitrat, line 34: You should change to 0.5 here then

src/ch01-03-hello-scarb.md
ametel01: 0.4 has compiler v1 so no point to merge it into v2 branch 😅

enitrat:
@ametel01 I just updated to `scarb 0.5.0-alpha.2 (7529c2ae0 2023-06-22)` and:
- `scarb test` works, but `scarb test --starknet` doesn't
- Tests written under`tests` instead of `src` are not detected.

Did you change anything in your  `Scarb.toml` files (perhaps you're overridden `cairo-test` and using `protostar`?

ametel01:
This is my Scarb.time: https://github.com/ametel01/NoGame-cairo-1/blob/main/nogame/Scarb.toml

ametel01:
I realised that  my `scarb test` was working differently because i had defined the test folder in my `cairo_project.toml`:
```[crate_roots]
nogame = "src"
mine_tests = "src/mines/tests"
compounds_test = "src/compounds/tests"
fleet_test = "src/fleet/tests"```

Should change the docs paragraph to make it agnostic and use the default `tests` folder under `/src`?

As for the `--starknet` flag i'm not sure why we have a different behaviour but my test would not run without the flag. Here my `Scarb.toml`:

```[package]
name = "nogame"
version = "0.1.0"
cairo-version = ">=2.0.0-rc0"
authors = ["ametel01"]
license = "MIT"

[dependencies]
starknet = ">=2.0.0-rc0"

[[target.starknet-contract]]

[scripts]
test = "cairo-test ."

[lib]
sierra = true  
casm = false   ```

**Comment by enitrat** [comment]
> Should change the docs paragraph to make it agnostic and use the default tests folder under /src?

Yes, ideally we never use `cairo_project.toml` again now that Scarb has all the required features

> As for the --starknet flag i'm not sure why we have a different behaviour but my tests would not run without the flag. Here my Scarb.toml:

I think it's because you overwrote the default `scarb test` command:
```
[scripts]
test = "cairo-test ."
```

In this part, you're explicitly overriding the default `scarb test` behavior to execute `cairo-test .` instead (which reads from your `cairo_project.toml`. I would advise removing this, and `cairo_project.toml` from your repository to not add too much confusion

ametel01:
Removed `cairo_project.toml`  and `scripts` section from `Scarb.toml` as you said and moved everything to `/src/tests` . I have a clearer understanding now. I'll change the paragraph in the docs.

listings/ch07-generic-types-and-traits/no_listing_09_result.cairo
enitrat, line 1: why was this deleted?

theme/fonts/OPEN-SANS-LICENSE.txt
enitrat, line 1: why was this deleted?

enitrat (approved):
All good! Thanks @ametel01 🔥

PR #258, opened 2023-06-23
Added StorageAccess to appendix 3

### LINKED ISSUE #232: feat: derive StorageAccess
We can now use `derive(StorageAccess)` for user-defined structs.
This small addition goes into [Appendix 3](https://cairo-book.github.io/appendix-03-derivable-traits.html)

ametel01:
closes #232

### DIFF
--- a/src/appendix-02-operators-and-symbols.md
+++ b/src/appendix-02-operators-and-symbols.md
@@ -8,39 +8,39 @@ Table B-1 contains the operators in Cairo, an example of how the operator would 
 
 <span class="caption">Table B-1: Operators</span>
 
-| Operator            | Example                                                 | Explanation                              | Overloadable? |
-| ------------------- | ------------------------------------------------------- | ---------------------------------------- | ------------- |
-| `!`                 | `!expr`                                                 | Bitwise or logical complement            | `Not`         |
-| `!=`                | `expr != expr`                                          | Non-equality comparison                  | `PartialEq`   |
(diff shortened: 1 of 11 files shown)


src/appendix-02-operators-and-symbols.md
enitrat, line 1: Formatting broke here

enitrat (changes requested):

src/appendix-03-derivable-traits.md
```diff
@@ -148,6 +148,27 @@ When moving out of scope, variables need to be moved first. This is where the `D
...
+Example:
+
+```rust
+#[derive(Drop, storage_access::StorageAccess)]
+struct A {
+    item_one: felt252,
+    item_two: felt252,
+}
+
+#[storage]
```
enitrat, line 164: The example seems odd here. I don't think it's valid even with the new syntax.

ametel01: Could you elaborate more? Indeed the compiler does not complain about custom structs not deriving StorageAccess. For your reference this is the PR i referred to:
https://github.com/starkware-libs/cairo/pull/3062

enitrat: Sorry, what I meant here was targeted to the `#[storage]` struct being outside of a contract - ideally, we want every code snippet to be compilable. That means that you should wrap this struct inside a contract, as a standalone storage struct doesn't make a lot of sense.

enitrat (approved):
LGTM :100: Thanks @ametel01

PR #272, opened 2023-07-01
:rocket: add macros chapter

### LINKED ISSUE #264: feat: macros
Create a section under "Advanced Features" that explains how to use available macros`consteval_int!`, `array!`, and how they work under the hood.

references: https://github.com/starkware-libs/cairo/commit/35953ba63e05c42c547960749a78255f4832cffe
https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-plugins/src/plugins/inline_macros
https://doc.rust-lang.org/book/ch19-06-macros.html

est. 1 hour

0xEniotna:
> Il est pas frais mon poisson ?

Closes #264.

I'm explaining `array!` and `consteval_int!` macros.


0xEniotna:
I'm not 100% sure about what I say in the under the hood part

src/ch10-02-macros.md
```diff
@@ -0,0 +1,35 @@
+# Macros
+
+The Cairo language has some plugins that allows developers to simplify their code. They are called `inline_macros` and are a way of writing code that writes code. It is easier to understand with an example.
+
+Here is a function that prints the `felt252` passed to it.
```
enitrat, line 5: Before introducing a practical example, I think it makes more sense to introduce motivation behind the macro: 
_Why_ do we need this macro, what use case does it solve? for `array!`, you could mention that we sometimes need to instantiate arrays with values known at compile time, and that the current way of doing this (create array + append n values) is a bit cumbersome. 
Try to be as explicit as possible: A framework that could be apply to write any section in the book is

- Introduction (present a use case that validates this requirement) / Motivation (why are we doing this)
- Solution (present how you can do things)
- Example

I don't think the examples provided here are very meaningful, as they don't really show how practical the macro is when defining "big" arrays

src/ch10-02-macros.md
```diff
@@ -0,0 +1,35 @@
...
+fn print_felt252(message: felt252) {
+    print(array![message]);
+}
+```
+
+In Cairo, there are only two `macros`: `array![]` and `consteval_int!()`.
+
+`array!` creates an array with the elements passed in paramater.
+`consteval_int!` performs the computations passed to it before returning the constant result.
+
```
enitrat, line 27: Same remark as for the array! Macro: motivation/solution/example

src/ch10-02-macros.md
```diff
@@ -0,0 +1,35 @@
...
+And this is the exact same function with a macro `array!` instead.
+
+```rust
+fn print_felt252(message: felt252) {
+    print(array![message]);
+}
+```
+
+In Cairo, there are only two `macros`: `array![]` and `consteval_int!()`.
```
enitrat, line 23: This should come right after the section introduction

enitrat:
> I'm not 100% sure about what I say in the under the hood part
For example, for the array![] macro: 
It creates a variable, and sequentially calls `arr.append()` will all the values passed to the macro
https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-plugins/src/plugins/inline_macros/array.rs

src/ch10-02-macros.md
```diff
@@ -0,0 +1,51 @@
...
+### Let's start by `array!`
+
+Sometimes, we need to create arrays with values that are already known at compile time. The basic way of doing that is redundant. You would first declare the array and then append one by one every values. `array!` creates a simpler way of doing this task by combining the two steps.
+At compile-time, the compiler will create an array and append sequencially every values passed to the `array!` macro.
+
+Without `array!`:
+
+```rust
+let mut input = Default::default();
+input.append(0x0000000000000001);
```
enitrat, line 14: can you create a `.cairo` flie under listings for this? You can check how we can include only specific sections of cairo code from complete, compilable files using the `ANCHOR` system in markdown files in other sections.

src/ch10-02-macros.md
```diff
@@ -0,0 +1,51 @@
...
+input.append(0x0000000000000002);
```
enitrat, line 15: or just 1, 2, 3 ? 😄

enitrat (approved):
LGTM once comments addressed 🔥

src/ch10-02-macros.md
0xEniotna: bruh, sometimes I work with brain off, that's annoying af. Copy pasted from the cairo repo without reading

enitrat:
Merging this when updating to 2.1 release

enitrat (approved):

Issue #263, opened 2023-06-28
feat: extend `#[generate_trait]` part


enitrat:
reference: https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-plugins/src/test_data/generate_trait

est. 2 to 3 hours

rkdud007:
@enitrat i can do this :)

enitrat:
@rkdud007 Cool! Please take your time, explain things clearly and provide comprehensive and meaningful examples 😄 
There is a very limited part about `generate_trait` right now, we want to show what happens under the hood, what code is being generated, how to use it for generics, etc.

rkdud007:
can I know where is generate_trait currently written? I cannot find it 

I'm planning to add it on 7.2 traits in cairo bottom

rkdud007:
actually I suggest to make another chapter or appendix for Plugins.

enitrat:
https://github.com/cairo-book/cairo-book.github.io/blob/cairo-2.0/src/ch07-02-traits-in-cairo.md#implementing-a-trait-without-writing-its-declaration

rkdud007:
oh wait its on cairo2.0 branch, so that means i better push it on cairo2.0 branch right?

enitrat:
yes

PR #274, opened 2023-07-01
generate_trait attribute explanation

rkdud007:
#263

### DIFF
--- a/src/ch07-02-traits-in-cairo.md
+++ b/src/ch07-02-traits-in-cairo.md
@@ -52,6 +52,41 @@ impl RectangleGeometry of RectangleGeometryTrait {
 
 In the aforementioned code, there is no need to manually define the trait. The compiler will automatically handle its definition, dynamically generating and updating it as new functions are introduced.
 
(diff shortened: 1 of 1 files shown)


rkdud007:
sorry commit name is very wrong. should be extend generate_trait 😅

enitrat:
Hi @rkdud007, could you please
- Move all cairo code blocks under `listings/` as it's already done for other chapters
- Refer to specific line of codes using the anchor system https://rust-lang.github.io/mdBook/format/mdbook.html?highlight=ANCHOR#including-portions-of-a-file
- Work on making the explanations clearer? It's very confusing as it is written now. 
- Show what the compiler is producing _exactly_ so that we can understand what kind of magic is happening and how to use the correct naming patterns to use traits when using `generate_trait` on impls

Try to be as explicit as possible: A framework that could be apply to write any section in the book is

- Introduction (present a use case that validates this requirement) / Motivation (why are we doing this)
- Solution (present how you can do things)
- Example

I will review the PR once these are addressed

enitrat:
@rkdud007 you should close this one and reopen one from main

rkdud007:
Alright🫡

PR #280, opened 2023-07-05
Update dictionaries initial value in tables

gaetbout:
As the initial value is zero, it should starts with an initial value of zero or am I missing something? 😅

### DIFF
--- a/src/ch15-00-dictionaries.md
+++ b/src/ch15-00-dictionaries.md
@@ -66,8 +66,8 @@ These instructions would then produce the following list of entries:
 
 |  key  | previous | new |
 | :---: | -------- | --- |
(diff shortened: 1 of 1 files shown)


enitrat:
That's what I thought too at first but after reading the previous Cairo0 docs, that's how it was presented 🤔 

c.f. https://www.cairo-lang.org/docs/hello_cairo/dict.html#dictionaries-maps-in-cairo

enitrat:
@rodrigo-pino, @gaetbout, Shahar just confirmed that the prev_value for the first entry is indeed `0` 😄

enitrat (approved):
TYVM @gaetbout !!

PR #281, opened 2023-07-06
Improvements to the Contract Dispatchers section

martriay:
This PR suggests some readability improvements on top of consistent and standard-conforming contract naming.

### DIFF
--- a/listings/ch99-starknet-smart-contracts/listing_99_06_sample_contract.cairo
+++ b/listings/ch99-starknet-smart-contracts/listing_99_06_sample_contract.cairo
@@ -39,7 +39,7 @@ trait ITokenWrapper<TContractState> {
 //ANCHOR: here
 //**** Specify interface here ****//
 #[starknet::contract]
-mod token_wrapper {
+mod TokenWrapper {
     use super::IERC20DispatcherTrait;
     use super::IERC20Dispatcher;
     use super::ITokenWrapper;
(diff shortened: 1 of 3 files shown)


enitrat:
Hi @martriay, thanks for the improvements - That part will very likely be completely reworked in the near future as I'm not completely satisfied with its quality.

Regarding the module casing, this is something that I think would be worth discussing. In Rust, modules are written in`snake_case`.  In a recent commit in the Cairo repository, modules and contracts casing has been changed from `PascalCase` to `snake_case` - hence the changes in the book.

martriay:
I haven't seen the latest Cairo 1 module names but I'd like to see some rationale behind, otherwise I don't see any reason to use Rust's conventions. Moreover, I perceive some inconsistent criteria, it was very important for Cairo to not be Solidity, so we broke compatibility in places where compatibility/similarity didn't hurt -- we just wanted to be different. And sometimes I see us following Rust conventions, while some others clarifying Cairo is not Rust.

So what is it? What are we basing features and styles on, and why? By which criteria and/or reasons we rename entire codebases? Who make these calls and where are they discussed?

enitrat:
> Who make these calls and where are they discussed?

For now, it mostly seems that SW is leading these calls - but they're not being discussed anywhere. Ideally they should be explicitly specified as SNIPs, as they were in Cairo 0. I tried to start the topic but didn't get any further engagement on that

enitrat:
Let's keep it PascalCase until something clearer emerges. Thanks!

enitrat (approved):

PR #290, opened 2023-07-14
Update ch01-02-hello-world.md to fix the command adding `--single-file`

marciob:
Following the instructions and running only `cargo-run main.cairo` will output an error saying `Error: The given path is a file, but --single-file was not supplied.`. So for that case it's necessary to run `cargo-run --single-file main.cairo`


enitrat:
Hi @marciob thanks for the update, I will merge this one we update the book to support the 2.1 release. For now, we're keeping it compatible with the 2.0 release.

PR #304, opened 2023-07-25
Update ch99-02-01-abis-and-interfaces.md

empea-careercriminal:
just a typo


enitrat (approved):
thanks for spotting all these !

PR #310, opened 2023-07-30
fix: explicitly state that desnap copies a value

### LINKED ISSUE #307: Explicitly state that desnap copies a value
Addition to references and snapshot chapter

julio4:
Close #307

### DIFF
--- a/src/ch03-02-references-and-snapshots.md
+++ b/src/ch03-02-references-and-snapshots.md
@@ -64,9 +64,9 @@ The scope in which the variable `array_snapshot` is valid is the same as any fun
 
 #### Desnap Operator
 
-To converts a snapshot back into a regular value, you can use the `desnap` operator `*`, which serves as the opposite of the `@` operator: it takes a snapshot as input and returns the value it points to.
+To convert a snapshot back into a regular value, you can use the `desnap` operator `*`, which serves as the opposite of the `@` operator: the snapshot value is copied to a new memory cell.
 
-It's important to note that during this conversion process, the value it points to is copied into a new variable. This enables multiple uses of the snapshot without concerns about ownership transfers. This also means that the value pointed to by the snapshot must be copyable and droppable (which is not the case for Arrays, as they don't implement `Copy`).
+It's important to note that during this conversion process, the value it points to is copied into a new variable. This enables multiple uses of the underlying value without concerns about ownership transfers. This also means that the value pointed to by the snapshot must be copyable (which is not the case for Arrays, as they don't implement `Copy`).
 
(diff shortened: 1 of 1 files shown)


src/ch03-02-references-and-snapshots.md
```diff
@@ -62,9 +62,13 @@ fn calculate_length(
...
+#### Desnap Operator
 
-The snapshot type is always copyable and droppable, so that you can use it multiple times without worrying about ownership transfers.
+To converts a snapshot back into a regular value, you can use the `desnap` operator `*`, which serves as the opposite of the `@` operator: it takes a snapshot as input and returns the value it points to.
+
+It's important to note that during this conversion process, the value it points to is copied into a new variable. This enables multiple uses of the snapshot without concerns about ownership transfers. This also means that the value pointed to by the snapshot must be copyable and droppable (which is not the case for Arrays, as they don't implement `Copy`).
```
enitrat, line 69: > must be copyable and droppable

Actually, only `copyable`. (even though in most cases drop is also required - but not specifically by the desnap operation)

src/ch03-02-references-and-snapshots.md
```diff
@@ -62,9 +62,13 @@ fn calculate_length(
...
 
-Snapshots can be converted back into regular values using the `desnap` operator `*`, as long as the value type is copyable (which is not the case for Arrays, as they don't implement `Copy`). In the following example, we want to calculate the area of a rectangle, but we don't want to take ownership of the rectangle in the `calculate_area` function, because we might want to use the rectangle again after the function call. Since our function doesn't mutate the rectangle instance, we can pass the snapshot of the rectangle to the function, and then transform the snapshots back into values using the `desnap` operator `*`.
+#### Desnap Operator
 
-The snapshot type is always copyable and droppable, so that you can use it multiple times without worrying about ownership transfers.
+To converts a snapshot back into a regular value, you can use the `desnap` operator `*`, which serves as the opposite of the `@` operator: it takes a snapshot as input and returns the value it points to.
```
enitrat, line 67: > converts

convert

src/ch03-02-references-and-snapshots.md
enitrat, line 69: > the value it points

I don't really like the term "points" here, which would suppose that the snapshot is some kind of pointer to a value: it's just a feature of the type system that wraps a regular value

You can replace it with "the snapshot value is copied to a new memory cell"

enitrat: > This enables multiple uses of the snapshot

of the underlying value?

enitrat (approved):
@julio4 I reverted the change from memory cell to variable, as actually I don't think it uses a new memory cell - sorry!

PR #315, opened 2023-08-01
refactor: use scarb packages

### LINKED ISSUE #314: refactor: make all listing scarb packages
As part of the migration to use only Scarb binaries, all listings should be packaged into Scarb modules.

The CI will only use `scarb build`, `scarb run` and `scarb test` binaries. Pure Cairo binaries will no longer be needed.

It means that all code inclusion paths will be transformed from `{{#include ../listings/path/to/listing}}` to `{{#include ../listings/path/to/listing/src/lib.cairo}`. All code execution examples will be changed from `cairo-run /file/path.cairo` to `scarb build && scarb cairo-run` (with a gas option if required)

To avoid manually migrating all files, a bash script can be created to automatically move files as required and edit the markdown files.

Estimated: 1hr of work

### LINKED ISSUE #316: refactor: change all `cairo-run` instructions to `scarb cairo-run`
We will make sure that it is clearly indicated (but I remember it should be) that for each code example, we run it inside a Scarb package in the `lib.cairo` file - so it should not cause any trouble. Perhaps it would be nice to emphasize on how code examples can be run in the beginning of the book.

enitrat:
Closes #314 
Closes #316 
Partially covers #320 

Waiting for https://github.com/software-mansion/scarb/issues/526 to optimize scarb cairo-run command

- Moves all code snippets `filename.cairo` to `filename/src/lib.cairo` and recreates the required scarb structure
- All markdown files have been changed accordingly
- Rust cairo-verify script now uses scarb
- CI now uses scarb (faster!)

Overall (structure change + markdown + rust tool + CI + 2.1.0 migration) took ~ 5 hours


enitrat:
Hey @fricoben @julio4 @glihm could you please take a look and make sure everything's OK? (especially rust modifications)

glihm:
Amazing work @enitrat. ⚡️
I will check it asap on my side.

cairo-verify/src/utils.rs
```diff
@@ -5,24 +5,24 @@ use walkdir::WalkDir;
...
+                scarb_manifests.push(entry.path().display().to_string());
             }
         }
     }
 
-    cairo_files
+    scarb_manifests
 }
 
 /// Will replace the file path contained in the input string with a clickable format for better output
```
glihm, line 28: Here in this function, due to the change, we may also update the file path printed
in the error summary. As we have all `lib.cairo`.

2023-08-02-165850_206x148_scrot

This can be done with small changes to print the shortened file's path:

```diff
--- a/cairo-verify/src/utils.rs
+++ b/cairo-verify/src/utils.rs
@@ -34,7 +34,9 @@ pub fn clickable(relative_path: &str) -> String {
         .to_string();
     let mut path_parts: Vec<&str> = full_path.split(|c: char| c == '\\' || c == '/').collect();
 
-    let mut filename: String = path_parts.last().unwrap_or(&"").to_string();
+    let file_listing_path: Vec<&str> = full_path.split("listings").collect();
+    let mut filename: String = file_listing_path.last().unwrap_or(&"")[1..].to_string();
+
     let re = Regex::new(r"([^:]+(:\d+:\d+)?)(:\s|$)").unwrap();
     if let Some(captures) = re.captures(filename.as_str()) {
         filename = captures.get(1).map_or("", |m| m.as_str()).to_string();
@@ -43,7 +45,6 @@ pub fn clickable(relative_path: &str) -> String {
     if let Some(parts) = path_parts.last_mut() {
         *parts = &filename;
     }
-    let full_path = path_parts.join("/");
 
     let clickable_format = format!(
         "\u{1b}]8;;file://{}\u{1b}\\{}\u{1b}]8;;\u{1b}\\",
```

2023-08-02-170528_848x206_scrot

glihm (commented):
Nice job @enitrat!

Some comments:

1. Can you remind me if the tags are supposed to be stripped out? Because I still see `does_not_compile` in the markdown, also in tags comment into cairo files, and the output is like this, which is not expected, right?

2023-08-02-171250_862x206_scrot

2. The switch to scarb is a great move IMHO, the file structure feels a bit heavier, but it's nice for the future.

cairo-verify/src/utils.rs
enitrat: Hey thanks for the proposal for 2.

Regarding 1. it's weird because I don't see any problem when rendering it locally with `mdbook serve`. Is the preprocessor running correctly?

glihm: I'll double check but if you have it nicely rendered locally I may be missing something.

PR #323, opened 2023-08-09
Fixed typo 

trbutler4:
The book read: 

Aside from Scarb you will no have Starkli installed.

Changed to:

Aside from Scarb you will need to have Starkli installed.


enitrat (approved):
thank you :)

PR #326, opened 2023-08-13
ch01-01 minor edit

Utilitycoder:
I am not sure if the error is intentional, but I made a correction to `hello_world.cairo` printing `Hello, World!` instead of `Hello, Scarb!`.


Utilitycoder:
Maybe the `'Hello Scarb!'.print()` is the mistake here. `Hello, World` appears all over the document instead of `Hello, Scarb!`

enitrat (approved):
Great, thanks, must've missed this when refactoring. 💯

PR #329, opened 2023-08-15
updated function mentioned in the paragraph to match the command

trbutler4:
also verified that the command itself is correct, as i am following the voting contract deployment tutorial


enitrat (approved):
nice catch :)

PR #330, opened 2023-08-16
Fix Name Registry contract

satyambnsal:
Name Registry contract external function `store_name` should call internal function `_store_name`.


enitrat (approved):
TYVM!

PR #331, opened 2023-08-17
Fix broken Scarb link

EdgarBarrantes:
Broken: https://docs.swmansion.com/scarb/docs/starknet/contract-target
New: https://docs.swmansion.com/scarb/docs/extensions/starknet/contract-target.html#compiling-external-contracts


enitrat (approved):
Thanks!

PR #346, opened 2023-09-11
Fixed code snippets in `array!` macro chapter

barretodavid:
This is how the `array!` code snippet currently looks like:

Screenshot 2023-09-11 at 4 53 34 PM

This is how it looks like with the fix:

Screenshot 2023-09-11 at 4 53 45 PM


enitrat (approved):
thanks!

PR #347, opened 2023-09-12
improve the description of integer division

orangeryang:
add towards zero


enitrat:
Did you also modify the source files in English? 
We must make sure that there is no mismatch in content between english and translations

orangeryang:
There is 'towards zero' in English/origin description, but translation lacks it.I just added it.

enitrat (approved):
Great 👍

PR #352, opened 2023-09-15
fix: compile error of no_listing_03_enum_option

waheya:


enitrat:
Hey @waheya thanks for the PR! This import is included in the prelude and is no longer required after 2.2.0 (scarb 0.7.0)

PR #359, opened 2023-09-18
Update ch02-99-01-arrays.md

sinsotec:
Span example was not showing


enitrat:
Thanks, fixed it with some other issues in #361!

PR #363, opened 2023-09-18
Fix Dictionary struct member example

satyambnsal:
In Dictionary struct member example code for `UserDatabase`, change method name `get_user`  to `get_balance` for intuitiveness.


enitrat (approved):
Looks good, thank you!

PR #374, opened 2023-09-22
Precisions about Integer/struct/mapping storage

### LINKED ISSUE #367: Enrich ''Storage Variables'' chapter
Hi @enitrat  , 
after completing the Starknet Storage Layout quest yesterday on NG, i spent a few hours facing a problem due to a misunderstood concept.  

from cairo book : 

In mappings, the address of the value at key k_1,...,k_n is h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n) where ℎ is the Pedersen hash and the final value is taken mod2251−256. You can learn more about the contract storage layout in the [Starknet Documentation](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-storage/#storage_variables)

I think this section should be enriched : when reading it, it feels like the nested pederson scheme should only be applied to nested mappings, with 2 different keys to use in the pedesen hash function.

But if you have a single mapping, with a u256 as a key, you should also apply 2 nested pedersen hash, with the only key of the mapping (actually 2 felt252, but with max size U128) considered as 2 separate keys. so basically, a single mapping with a u256 as a key is identical to a 2 nested mappings?

I think i could try to rewrite this section to make it clearer, and maybe add more content about structs storage, and more important about how u256 are stored underneath the hood.

Just let me know if you think this is relevant, but i think it is as i didn't fully understand all this concept when reading cairo book and starknet documentation.

i've been  partially helped with this blog post, but it's all about cairo 0, maybe some fresh content about u256 in cairo book would be great.

https://dev.to/peterblockman/cairo-uint256-ojf

TAdev0:
@enitrat  tell me what you think about these modifications :)

Closes #367

### DIFF
--- a/src/ch02-02-data-types.md
+++ b/src/ch02-02-data-types.md
@@ -60,7 +60,7 @@ As variables are unsigned, they can't contain a negative number. This code will 
 {{#include ../listings/ch02-common-programming-concepts/no_listing_07_integer_types/src/lib.cairo}}
 ```
 
-All integer types previously mentioned fit into a `felt252`, except for `u256` which needs 4 more bits to be stored. Under the hood, `u256` is basically a struct containing 2 variables `(low : felt252, high : felt252)`, which actually only use 128 bits each.
+All integer types previously mentioned fit into a `felt252`, except for `u256` which needs 4 more bits to be stored. Under the hood, `u256` is basically a struct with 2 fields: `u256 {low: u128, high: u128}`
 
 You can write integer literals in any of the forms shown in Table 3-2. Note
 that number literals that can be multiple numeric types allow a type suffix,
--- a/src/ch02-02-data-types.md
+++ b/src/ch02-02-data-types.md
@@ -60,6 +60,8 @@ As variables are unsigned, they can't contain a negative number. This code will 
 {{#include ../listings/ch02-common-programming-concepts/no_listing_07_integer_types/src/lib.cairo}}
 ```
 
+All integer types previously mentioned fit into a `felt252`, except for `u256` which needs 4 more bits to be stored. Under the hood, `u256` is basically a struct containing 2 variables `(low : felt252, high : felt252)`, which actually only use 128 bits each.
+
 You can write integer literals in any of the forms shown in Table 3-2. Note
 that number literals that can be multiple numeric types allow a type suffix,
 such as `57_u8`, to designate the type.
--- a/src/ch99-01-03-01-storage-variables.md
+++ b/src/ch99-01-03-01-storage-variables.md
@@ -22,7 +22,7 @@ In the case of `u256`, the address will be a pointer to `low : felt252`, with `h
 
 ### Storing structs
 
-The compiler knows how to store basic data types, such as unsigned integers (`u8`, `u128`, `u256`...), `felt252`, `bool`, `ContractAddress`, etc. But what if you want to store a custom struct in storage? In that case, you have to explicitly tell the compiler how to store your struct in storage.
+Most types from the core library, such as unsigned integers (`u8`, `u128`, `u256`...), `felt252`, `bool`, `ContractAddress`, etc. implement the `Store` trait and can thus be stored without further action. But what if you wanted to store a struct that you defined yourself? In that case, you have to explicitly tell the compiler how to store your struct.
 In our example, we want to store a `Person` struct in storage, which is possible by implementing the `Store` trait for the `Person` type. This can be achieved by simply adding a `#[derive(starknet::Store)]` attribute on top of our struct definition.
 
 ```rust, noplayground
--- a/src/ch99-01-03-01-storage-variables.md
+++ b/src/ch99-01-03-01-storage-variables.md
@@ -15,8 +15,7 @@ except that it **must** be annotated with `#[storage]` allowing you to store map
 
 ### Storing data types
 
-The base address of a value in storage is `sn_keccak(variable_name)`, with `variable_name` as the ASCII encoding of the variable's name. As the keccak256 hash function produces an output of 256 bits that doesn't fit in a `felt252`, we use `sn_keccak` which is defined as the first 250 bits of the Keccak256 hash.
-In the case of `u256`, the address will be a pointer to `low : felt252`, with `high : felt252` stored contiguously.
+The base address `StorageBaseAddress` of a value in storage is `sn_keccak(variable_name)`, with `variable_name` as the ASCII encoding of the variable's name. As the keccak256 hash function produces an output of 256 bits that doesn't fit in a `felt252`, we use `sn_keccak` which is defined as the first 250 bits of the Keccak256 hash.
 
 ### Storing structs
 
--- a/src/ch99-01-03-01-storage-variables.md
+++ b/src/ch99-01-03-01-storage-variables.md
@@ -17,7 +17,7 @@ except that it **must** be annotated with `#[storage]` allowing you to store map
 
 Basic data types are single values contained in a `felt252`, except for `u256` which is a struct of 2 `felt252`.
 
-The address of such variables in storage is `sn_keccak(variable_name)`, with variable_name as the ASCII encoding of the variable's name. As keccak256 hash doesn't fit into a felt252, we use `sn_keccak` which is defined as the first 250 bits of the Keccak256 hash.
+The base address of a value in storage is `sn_keccak(variable_name)`, with `variable_name` as the ASCII encoding of the variable's name. As the keccak256 hash function produces an output of 256 bits that doesn't fit in a `felt252`, we use `sn_keccak` which is defined as the first 250 bits of the Keccak256 hash.
 In the case of `u256`, the address will be a pointer to `low : felt252`, with `high : felt252` stored contiguously.
 
 ### Storing structs
--- a/src/ch99-01-03-01-storage-variables.md
+++ b/src/ch99-01-03-01-storage-variables.md
@@ -48,7 +48,7 @@ You can also create more complex mappings than that; you can find one in Listing
 
 In mappings, the address of the value at key `k_1,...,k_n` is `h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n)` where ℎ is the Pedersen hash and the final value is taken `mod2251−256`.
 
-If the key of the mapping is a struct, each element of the struct constitue a key. For example, if you have a 2 elements struct as a key, the address will be `h(h(sn_keccak(variable_name),k_1),k_2)`.
+If the key of the mapping is a struct, each element of the struct constitutes a key. For example, if you have struct with two fields, the address will be `h(h(sn_keccak(variable_name),k_1),k_2)` - where `k_1` and `k_2` are the values of the two fields of the struct.
 
 Additionnaly, in the case of a nested mapping such as `LegacyMap((ContractAddress, ContractAddress), u8)`, the address will be computed in the same way : `h(h(sn_keccak(variable_name),k_1),k_2)`.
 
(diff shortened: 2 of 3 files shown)


TAdev0:
btw, i might have one question about storage :

starknet documentation says there are 2^251 slots in storage. Nevertheless, sn_keccak returns 250 bits and can therefore generate 2^250 different values, which is 2x less than the actual total slots. why?

src/ch99-01-03-01-storage-variables.md
```diff
@@ -13,7 +13,25 @@ Storage variables in Starknet contracts are stored in a special struct called `S
...
+### Storing basic data types
+
+Basic data types are single values contained in a `felt252`, except for `u256` which is a struct of 2 `felt252`.
+
+The address of such variables in storage is `sn_keccak(variable_name)`, with variable_name as the ASCII encoding of the variable's name. As keccak256 hash doesn't fit into a felt252, we use `sn_keccak` which is defined as the first 250 bits of the Keccak256 hash.
+In the case of `u256`, the address will be a pointer to `low : felt252`, with `high : felt252` stored contiguously.
+
+### Storing structs
+
+The compiler knows how to store basic data types, such as unsigned integers (`u8`, `u128`, `u256`...), `felt252`, `bool`, `ContractAddress`, etc. But what if you want to store a custom struct in storage? In that case, you have to explicitly tell the compiler how to store your struct in storage.
```
enitrat, line 25: This is not very accurate, actually you can store these types because the `Store` trait is implemented for these types in the core library.

```suggestion
Most types from the core library, such as unsigned integers (`u8`, `u128`, `u256`...), `felt252`, `bool`, `ContractAddress`, etc. implement the `Store` trait and can thus be stored without further action. But what if you wanted to store a struct that you defined yourself? In that case, you have to explicitly tell the compiler how to store your struct.
```

src/ch99-01-03-01-storage-variables.md
```diff
@@ -13,7 +13,25 @@ Storage variables in Starknet contracts are stored in a special struct called `S
...
+In our example, we want to store a `Person` struct in storage, so we have to tell the compiler how to store it in storage by adding a derive attribute of the `starknet::Store` trait to our struct definition.
```
enitrat, line 26:
```suggestion
In our example, we want to store a `Person` struct in storage, which is possible by implementing the `Store` trait for the `Person` type. This can be achieved by simply adding a `#[derive(starknet::Store)]` attribute on top of our struct definition.
```

src/ch99-01-03-01-storage-variables.md
```diff
@@ -13,7 +13,25 @@ Storage variables in Starknet contracts are stored in a special struct called `S
...
+
+```rust, noplayground
+{{#include ../listings/ch99-starknet-smart-contracts/listing_99_03_example_contract/src/lib.cairo:person}}
+```
+
+The address of structs is `sn_keccak(variable_name)`, pointing to the first element of the struct. All other elements contained in the struct are stored contiguously.
```
enitrat, line 32: Might be worth adding an example of how storing a struct containing a struct works when illustrating the value at the base address on top of the suggestion. Feel free to add more details to the suggestion if it's still a bit hard to grasp


```suggestion
The base storage address for structs remains `sn_keccak(variable_name)`, and the value contained at this address is the first primitive type contained by the struct.

Subsequent fields are stored in addresses contiguous to the first elements at addresses `base_address + i`.
```

src/ch99-01-03-01-storage-variables.md
```diff
@@ -28,17 +46,13 @@ You can also create more complex mappings than that; you can find one in Listing
 
 <span class="caption">Listing 99-2bis: Storing mappings</span>
 
-In mappings, the address of the value at key `k_1,...,k_n` is `h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n)` where ℎ
-is the Pedersen hash and the final value is taken `mod2251−256`. You can learn more about the contract storage layout in the [Starknet Documentation](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-storage/#storage_variables)
+In mappings, the address of the value at key `k_1,...,k_n` is `h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n)` where ℎ is the Pedersen hash and the final value is taken `mod2251−256`.
```
enitrat, line 49:
```suggestion
In mappings, the address of the value at key `k_1,...,k_n` is `h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n)` where ℎ is the Pedersen hash and the final value is taken `mod (2^251) − 256`.
```

src/ch99-01-03-01-storage-variables.md
```diff
@@ -28,17 +46,13 @@ You can also create more complex mappings than that; you can find one in Listing
...
 
-### Storing custom structs
+If the key of the mapping is a struct, each element of the struct constitue a key. For example, if you have a 2 elements struct as a key, the address will be `h(h(sn_keccak(variable_name),k_1),k_2)`.
```
enitrat, line 51:
```suggestion
If the key of the mapping is a struct, each element of the struct constitutes a key. For example, if you have struct with two fields, the address will be `h(h(sn_keccak(variable_name),k_1),k_2)` - where `k_1` and `k_2` are the values of the two fields of the struct.
```

src/ch99-01-03-01-storage-variables.md
```diff
@@ -28,17 +46,13 @@ You can also create more complex mappings than that; you can find one in Listing
...
 
-The compiler knows how to store basic data types, such as unsigned integers (`u8`, `u128`, `u256`...), `felt252`, `ContractAddress`, etc. But what if you want to store a custom struct in storage? In that case, you have to explicitly tell the compiler how to store your struct in storage.
-In our example, we want to store a `Person` struct in storage, so we have to tell the compiler how to store it in storage by adding a derive attribute of the `starknet::Store` trait to our struct definition.
+Additionnaly, in the case of a nested mapping such as `LegacyMap((ContractAddress, ContractAddress), u8)`, the address will be computed in the same way : `h(h(sn_keccak(variable_name),k_1),k_2)`.
```
enitrat, line 53:
```suggestion
Similarly, in the case of a nested mapping such as `LegacyMap((ContractAddress, ContractAddress), u8)`, the address will be computed in the same way: `h(h(sn_keccak(variable_name),k_1),k_2)`.
```

enitrat (commented):
Some comments and misunderstandings fixed

enitrat:
> btw, i might have one question about storage :
> 
> starknet documentation says there are 2^251 slots in storage. Nevertheless, sn_keccak returns 250 bits and can therefore generate 2^250 different values, which is 2x less than the actual total slots. why?

good Q, idk

src/ch02-02-data-types.md
```diff
@@ -60,6 +60,8 @@ As variables are unsigned, they can't contain a negative number. This code will
 {{#include ../listings/ch02-common-programming-concepts/no_listing_07_integer_types/src/lib.cairo}}
 ```
 
+All integer types previously mentioned fit into a `felt252`, except for `u256` which needs 4 more bits to be stored. Under the hood, `u256` is basically a struct containing 2 variables `(low : felt252, high : felt252)`, which actually only use 128 bits each.
```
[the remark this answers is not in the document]
TAdev0: really better design ! thank you

src/ch99-01-03-01-storage-variables.md
```diff
@@ -13,7 +13,25 @@ Storage variables in Starknet contracts are stored in a special struct called `S
 The storage struct is a [struct](./ch05-00-using-structs-to-structure-related-data.md) like any other,
 except that it **must** be annotated with `#[storage]` allowing you to store mappings using the `LegacyMap` type.
 
-### Storing Mappings
+### Storing basic data types
+
+Basic data types are single values contained in a `felt252`, except for `u256` which is a struct of 2 `felt252`.
```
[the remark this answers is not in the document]
TAdev0: I don't know, it gives more precision ? 
with ''basic data types'' naming, it feels like we're only talking about types that fit into felt252?

but yeah, it sounds like a repetition of 2.2 Data types, will remove it

src/ch99-01-03-01-storage-variables.md
TAdev0: hoops, some french word in my writing, sorry.

src/ch99-01-03-01-storage-variables.md
TAdev0: really better and more precise than mine. thank you.

src/ch99-01-03-01-storage-variables.md
TAdev0: mmh i didn't modify anything of this. But the suggestion is better, i commit it

src/ch99-01-03-01-storage-variables.md
```diff
@@ -13,7 +13,25 @@ Storage variables in Starknet contracts are stored in a special struct called `S
...
+
+The address of such variables in storage is `sn_keccak(variable_name)`, with variable_name as the ASCII encoding of the variable's name. As keccak256 hash doesn't fit into a felt252, we use `sn_keccak` which is defined as the first 250 bits of the Keccak256 hash.
+In the case of `u256`, the address will be a pointer to `low : felt252`, with `high : felt252` stored contiguously.
```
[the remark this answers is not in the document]
TAdev0: true.

TAdev0:
@enitrat  i think everything is fixed !

src/ch99-01-03-01-storage-variables.md
```diff
@@ -28,17 +45,13 @@ You can also create more complex mappings than that; you can find one in Listing
 
 <span class="caption">Listing 99-2bis: Storing mappings</span>
 
-In mappings, the address of the value at key `k_1,...,k_n` is `h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n)` where ℎ
-is the Pedersen hash and the final value is taken `mod2251−256`. You can learn more about the contract storage layout in the [Starknet Documentation](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-storage/#storage_variables)
+In mappings, the address of the value at key `k_1,...,k_n` is `h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n)` where ℎ is the Pedersen hash and the final value is taken `mod (2^251) − 256`.
 
-### Storing custom structs
+If the key of the mapping is a struct, each element of the struct constitutes a key. For example, if you have struct with two fields, the address will be `h(h(sn_keccak(variable_name),k_1),k_2)` - where `k_1` and `k_2` are the values of the two fields of the struct.
```
enitrat, line 50: This requires the struct to implement the `Hash` trait, which can be derived with the `#[derive(Hash)]` attribute

TAdev0: In NG quest, the struct didn't implement this trait and it worked.  I guess it is because all its members (basic types) already implement it?

TAdev0: If yes, i should precise that we need to use the #[derive(Hash)] if and only if one of the struct members do not implement this trait?

enitrat: > In NG quest, the struct didn't implement this trait and it worked. I guess it is because all its members (basic types) already implement it?

Really? That's odd, it should be the case. This might be a requirement that was added in 2.2.0, as the `Hash` trait wasn't even a thing before

enitrat: Here's the error in 2.2 if you try to compile without it

```shell
error: Trait has no implementation in context: core::hash::LegacyHash::<adder::Contract::Dapp>
 --> /Users/msaug/cairo_projects/adder/src/lib.cairo[contract]:77:47
                    hash::LegacyHash::<MyStruct>::hash(0x1a432f1ecd4c0ac028ebde3a3f78510a21d54087b161590a63080d33b702b8d, key))
                                              ^**^
                                              ```

TAdev0: ok after many tests on my local environment, you're definitely right.
I was wrong because Chest struct was actually not included in a mapping as a key.

I tried with 2.0.0 and 2.2.0 , actually both of them need the Hash trait to be implemented for the struct to be hashed in a .cairo file.

TAdev0:
PR updated!

enitrat (approved):
LGTM !

PR #384, opened 2023-10-02
Dispatcher example code sanity changes

satyambnsal:
`ContractAddress` already brought scope from starknet crate and other places in contract as well simply used as `ContractAddress`


enitrat (approved):
great 💪

PR #387, opened 2023-10-03
refactor: proposition for new `L1-L2` messaging content

glihm:
This PR adds simplicity to the code illustrating how the messaging works + some details on the "how".

Additionally, I've made a complete tutorial with working contracts to work locally with Anvil and Katana.

@enitrat sorry for the delay, I wanted to finish the PR on Katana to support messaging before writing this. :)

I would be very happy to have your feedback on this and from others interested on this topic!

### DIFF
--- a/listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
+++ b/listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
@@ -24,7 +24,7 @@ trait IContractL1<T> {
 
     /// Sends a message to L1 contract with a serialized struct.
     /// To send a struct in a payload of a message, you only have to ensure that
-    /// your structure is serializable implementing the `Serde` traits. Which
+    /// your structure is serializable implementing the `Serde` trait. Which
     /// is automatically done if your structure only contains already serializable members.
     ///
     /// # Arguments
--- a/listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
+++ b/listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
@@ -25,7 +25,7 @@ trait IContractL1<T> {
     /// Sends a message to L1 contract with a serialized struct.
     /// To send a struct in a payload of a message, you only have to ensure that
     /// your structure is serializable implementing the `Serde` trait. Which
-    /// is automatically done if your structure only contains already serializable members.
+    /// can be derived automatically if your struct only contains serializable members.
     ///
     /// # Arguments
     ///
--- a/listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
+++ b/listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
@@ -1,81 +1,83 @@
+//! A simple contract that sends and receives messages from/to
+//! the L1 (Ethereum).
+//!
+//! The reception of the messages is done using the `l1_handler` functions.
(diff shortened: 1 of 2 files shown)


src/ch99-04-00-L1-L2-messaging.md
```diff
@@ -81,54 +105,95 @@ function sendMessageToL2(
     ) external override returns (bytes32);
 ```
 
-In `ex01SendMessageToL2`, we first construct the message (the payload). It is an array of `uint256`. Then we call `StarknetCore.sendMessageToL2`. The first parameter is the L2 contract address. The second is the selector (or the `sn_keccak` hash of the name of the function we want to call on L2), followed by the payload. We add a fee to it as a `msg.value` because we need to pay the transaction on L2.
+The parameters are as follow:
+* `toAddress`: The contract address on L2 that will be called.
+* `selector`: The selector of the function of this contract at `toAddress`. This selector (function) must have the `#[l1_handler]` attribute to be callable.
+* `payload`: The payload is always an array or `felt252` (which are represented by `uint256` in solidity). For this reason we've inserted the input `myFelt` into the array.
```
enitrat, line 111:
```suggestion
* `payload`: The payload is always an array of `felt252` (which are represented by `uint256` in solidity). For this reason we've inserted the input `myFelt` into the array.
```

src/ch99-04-00-L1-L2-messaging.md
```diff
@@ -81,54 +105,95 @@ function sendMessageToL2(
...
+This is why we need to insert the input data into an array.
 
 On the Starknet side, to receive this message, we have:
 
-```rust
-{{#include ../listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo:here}}
+```rust,does_not_compile
```
enitrat, line 116: can you put this inside a file under `listings/`? You can put it inside a template contract and use ANCHOR to only refer to this part. This ensures that it remains compilable as Cairo versions are updated

enitrat (commented):
thanks its awesome!

src/ch99-04-00-L1-L2-messaging.md
glihm: Sorry, you're totally right. Just pushed the update.
I'm not able to remove the extra indent, any tips on that? Or it's not possible for now?

listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
```diff
@@ -1,81 +1,83 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `to_address` - Contract address on L1.
+    /// * `my_felt` - Felt value to be sent in the payload.
+    fn send_message_felt(ref self: T, to_address: starknet::EthAddress, my_felt: felt252);
 
+    /// Sends a message to L1 contract with a serialized struct.
+    /// To send a struct in a payload of a message, you only have to ensure that
+    /// your structure is serializable implementing the `Serde` traits. Which
```
enitrat, line 27:
```suggestion
    /// your structure is serializable implementing the `Serde` trait. Which
```

listings/ch99-starknet-smart-contracts/listing_99_04_L1_L2_messaging/src/lib.cairo
```diff
@@ -1,81 +1,83 @@
...
+    /// is automatically done if your structure only contains already serializable members.
```
enitrat, line 28:
```suggestion
    /// can be derived automatically if your struct only contains serializable members.
```

enitrat (commented):
last changes and then good!

enitrat (approved):

PR #390, opened 2023-10-03
Update ch99-01-04-01-voting-contract.md

OkoliEvans:
Added more details for setting up Account Descriptor, users will be clarified on how to set up signers and account descriptors easily and not run into blocks. Also added texts to guide users who encounter errors with compiler version while declaring a class hash that they can declare without adding the compiler version flag. This was added out of personal experience

### DIFF
--- a/src/ch99-01-04-01-voting-contract.md
+++ b/src/ch99-01-04-01-voting-contract.md
@@ -99,6 +99,7 @@ The Account Descriptor will look like this. You can get the public key and the s
   }
 }
 ```
+Copy the Account Descriptor above and paste in the newly created `account0_account.json file`, then replace "<SMART_WALLET_PUBLIC_KEY>", "<SMART_WALLET_CLASS_HASH>" and "<SMART_WALLET_ADDRESS>" with values from the `katana` account used.
 
 You can retrieve the smart wallet class hash (it will be the same for all your smart wallets) with the following command. Notice the use of the `--rpc` flag and the RPC endpoint provided by `katana`:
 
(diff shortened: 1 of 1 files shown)


src/ch99-01-04-01-voting-contract.md
```diff
@@ -122,6 +123,8 @@ Before deploying, we need to declare the contract. We can do this with the `star
 starkli declare target/dev/starknetbook_chapter_2_Vote.sierra.json --compiler-version 2.0.1 --rpc http://0.0.0.0:5050 --account ~/.starkli-wallets/deployer/account0_account.json --keystore ~/.starkli-wallets/deployer/account0_keystore.json
 ```
 
+If you encounter error while running the command with `--compiler-version 2.0.1`, you can skip the compiler version entirely and run the command without `--compiler-version 2.0.1`.
```
enitrat, line 126: > If you encounter error while running the command with `--compiler-version 2.0.1`, you can skip the compiler version entirely and run the command without `--compiler-version 2.0.1`.

Question: Since the book uses version 2.2.0 of the compiler - is this part still relevant? Should we instead talk about the `--compiler-version 2.2.0` flag?

OkoliEvans: I think yeah, when I tried to run the command  `starkli declare target/dev/YaslonV5_Vote.sierra.json --compiler-version 2.2.0 --rpc http://0.0.0.0:5050 --account acct1.json
 --keystore acct1_keystore.json`

I still got this error: invalid value '2.2.0' for '--compiler-version <COMPILER_VERSION>'
  [possible values: 2.1.0]

  tip: a similar value exists: '2.1.0'

enitrat: Please make sure that it still works with these changes

OkoliEvans: Okay, on it

PR #392, opened 2023-10-04
Fix Vote contract example

satyambnsal:
# Issue

Private method `_get_voting_result_in_percentage` doesn't check if `total_votes`  is `0` due to which `get_vote_status` contract call fails if we call this before any voting.

# Changes

check if `total_votes` is 0 in `_get_voting_result_in_percentage` private method.

other changes: use expression syntax for returning values from method(as other methods do the same).


enitrat (approved):
Great catch!

enitrat (approved):

PR #398, opened 2023-10-09
Update ch06-00-enums-and-pattern-matching.md

Nonnyjoe:
Added a brief introduction to chapter 6 landing page. This page is supposed to contain a brief overview and introduction to the subtopics covered in chapter 6 but is currently empty. I believe this little contribution will help readers get an overhead understanding of what the chapter entails and why the two subtopics were grouped together.

### DIFF
--- a/src/ch06-00-enums-and-pattern-matching.md
+++ b/src/ch06-00-enums-and-pattern-matching.md
@@ -1,1 +1,6 @@
 # Enums and Pattern Matching
+Enums which are also known as enumerations, are a natively supported data type in Cairo. Cairo provides a powerful and expressive way to define enums, making them an integral part of the language. Enums define a set of distinct elements or options, they are used to improve code readability, maintainability and also to make it easier to work with a fixed set of related constants. Enums combined with other constructs like the match countrol flow can be used for various purposes, including creating custom data types and modeling different states and options. 
+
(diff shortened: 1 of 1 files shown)


enitrat (commented):
I am not a fan of this for the following reasons:
- As an introductory text to a chapter, we are looking for _concise_ and _precise_ information, defining the terms used in the chapter. This introduction must be succint and not go into too many details
- Complex and verbose language like "natively supported data type" and "integral part of the language" don't add a lot of value and adds length to sentences for nothing - prefer concise and straight to the point sentences
- The concepts are not well separated: a lot of different concepts all at once - enums, pattern matching, traits, control flow. Prefer introducing the concepts one by one, sequentially, in the order they're presented in the chapter
- The content is too generic. For example `Enums define a set of distinct elements or options,` => there is a specific naming for this called a `variant`

Overall, your text could be improved by following these principles: being concrete, breaking concepts down step-by-step, using active voice, highlighting the most important points up front, and avoid verbose language. Less is better!

For reference, here's an introductory text that I think would suit the chapter well:

In this chapter, we’ll look at *enumerations*, also referred to as *enums*.
Enums allow you to define a type by enumerating its possible *variants*. First
we’ll define and use an enum to show how an enum can encode meaning along with
data. Next, we’ll explore a particularly useful enum, called `Option`, which
expresses that a value can be either something or nothing. Finally, we’ll look at
how pattern matching in the `match` expression makes it easy to run different
code for different values of an enum.

You'll note how short that text is and how much information is passed in only a few lines - Vocabulary, definition, usage, and utility are presented in just a few lines

Nonnyjoe:
Thanks for the tips, I'll definitely apply them to future write-ups.

PR #409, opened 2023-10-18
Fix: Correct parameter type in Coin::Quarter variant

milk-maid:
The "value_in_cents" function encountered an issue where the "Coin::Quarter" variant was expecting (UsState,) instead of UsState. This resulted in a "Method print not found" error.

To resolve this, I updated the Coin::Quarter variant to expect the correct parameter type: ```Coin::Quarter((state))```

This change ensures that the UsState is correctly matched in the "value_in_cents" function, fixing the error related to the print method.


enitrat:
The file you modified doesn't have a problem - the problem is actually in the snippet of code used above, which is not aligned with the one below for some reason.

milk-maid:
Alright, copied the latest as modified in the book not long ago. It is now working unlike before I made the PR. Thanks

PR #412, opened 2023-10-18
fix: Resolve "Variable not dropped" compilation error in `enums` crate

milk-maid:
The commit addresses a compilation error where the compiler screamed `Variable not dropped` indicating a missing drop implementation for the `Option<T>` enum. 

The error was triggered by the use of `enums::Option` in the test module, and it was resolved by adding `#[derive(Drop)]` to the declaration of the `Option<T>` enum.

This change ensures that the `Option<T>` enum implements the necessary drop trait, resolving the "Variable not dropped" error and allowing successful compilation of the `enums` module.


enitrat:
The Option enum is defined in the core library

PR #420, opened 2023-11-01
Updated Chapter 12.3.1

OkoliEvans:
There was no description on how to write  user defined struct to storage, and how to read as well. Added these with examples. @enitrat

### DIFF
--- a/src/ch99-01-03-01-storage-variables.md
+++ b/src/ch99-01-03-01-storage-variables.md
@@ -30,6 +30,11 @@ The base storage address for structs remains `sn_keccak(variable_name)`, and the
 
 Subsequent fields are stored in addresses contiguous to the first elements at addresses `base_address + i`.
 
+## Writing to stored structs
+
+To write custom storage structs to storage you have to first create the struct variant as a member of the default storage struct. The type of the variant is
+the custom storage struct.
+
 ### Storing mappings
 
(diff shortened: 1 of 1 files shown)


OkoliEvans:
> I don't understand why reading/writing structs would be separated from other types. How to read and write variables from storage is already explained here
> 
> https://book.cairo-lang.org/ch99-01-03-01-storage-variables.html#reading-from-storage

I will edit and commit again in a few hours

OkoliEvans:
Updated now, Please check @enitrat

enitrat:
Sorry, what I meant was:

Storing user defined structs is equivalent to storing any other kind of data. Read and writes are performed in the same way as well which is why I do not understand why there would be a section specifically about structs. 

The example provided do not compile and don't make a lot of sense (why is a Person stored under a variable named `name`?) 

For eventual subsequent PRs, you will need to verify that the examples compile first, and then add them to the `listings` directory, as it's currently done for all code snippets

OkoliEvans:
I thought adding the changes will make it clearer to understand how structs annotated with ```#[derive(starknet::Store)]``` are written to. In the book there is no explicit explanation about how this is done, the usual ```self.names.write(value)``` does not work for user defined storage structs.

enitrat:
> the usual self.names.write(value) does not work for user defined storage structs. @enitrat

It does! There is no difference between compiler-defined types, enums, or structs! You just call `self.var.write(value)` !

OkoliEvans:
Oh, I tried that yesterday and I was getting errors. I will confirm and pm you on Telegram. Thank you

PR #423, opened 2023-11-01
Added extensive example for call_contract_syscall

Akashneelesh:
I've completed the example, and it effectively addresses multiple scenarios. The primary focus is on passing the data to a function and forwarding it to the call_contract_syscall through the calldata. Additionally, the example explores various scenarios in which a receiving contract may need to be structured differently to handle the data, providing a comprehensive view of the topic.


enitrat (commented):
I don't think I'm going to merge this for a few reasons:

1. There is no addition to the book. These examples are not referenced anywhere.
2. Why are there deployment scripts? This is way too heavy to simply demonstrate how to use `call_contract_syscall` in an efficient manner. 
3. There is way too much content, no one will ever read this. It's too complex. Especially given that low-level syscalls will not be used in 99% of use cases.

However, there is still one interesting thing we could keep - it's how you used `Serde` to serialize the parameters into a Array<felt252>.

In the existing example, we have 

```rust
fn transfer_token(
            ref self: ContractState,
            address: starknet::ContractAddress,
            selector: felt252,
            calldata: Array<felt252>
        ) -> bool {
            let mut res = starknet::call_contract_syscall(address, selector, calldata.span())
                .unwrap_syscall();
            Serde::<bool>::deserialize(ref res).unwrap()
        }
  ```
  
  If you want to, you can change this example a bit to
  - not take the selector as a function parameter, and use the `selector!` macro instead to call an ERC20 contract
  - not take calldata as a function parameter, use the same as `transferFrom`
  - Serialize the parameters before calling `call_contract_syscall`

**Comment by Akashneelesh** [comment]
> I don't think I'm going to merge this for a few reasons:
> 
> 1. There is no addition to the book. These examples are not referenced anywhere.
> 2. Why are there deployment scripts? This is way too heavy to simply demonstrate how to use `call_contract_syscall` in an efficient manner.
> 3. There is way too much content, no one will ever read this. It's too complex. Especially given that low-level syscalls will not be used in 99% of use cases.
> 
> However, there is still one interesting thing we could keep - it's how you used `Serde` to serialize the parameters into a Array.
> 
> In the existing example, we have
> 
> ```rust
> fn transfer_token(
>             ref self: ContractState,
>             address: starknet::ContractAddress,
>             selector: felt252,
>             calldata: Array<felt252>
>         ) -> bool {
>             let mut res = starknet::call_contract_syscall(address, selector, calldata.span())
>                 .unwrap_syscall();
>             Serde::<bool>::deserialize(ref res).unwrap()
>         }
> ```
> 
> If you want to, you can change this example a bit to
> 
> * not take the selector as a function parameter, and use the `selector!` macro instead to call an ERC20 contract
> * not take calldata as a function parameter, use the same as `transferFrom`
> * Serialize the parameters before calling `call_contract_syscall`

Yeah sure got it, I'll do the change

**Comment by Akashneelesh** [comment]
I have updated the example according to the pointer that you had mentioned.

File: listings/ch99-starknet-smart-contracts/listing_99_09_call_contract_syscall/src/lib.cairo [hunk 1]
```diff
@@ -1,31 +1,43 @@
...
-            address: starknet::ContractAddress,
-            selector: felt252,
-            calldata: Array<felt252>
+            address: ContractAddress,
+            sender: ContractAddress,
+            recipient: ContractAddress,
+            amount: u256
         ) -> bool {
-            let mut res = starknet::call_contract_syscall(address, selector, calldata.span())
+            let mut Calldata: Array<felt252> = ArrayTrait::new();
```
enitrat, line 31: snake_case

src/ch99-02-02-contract-dispatcher-library-dispatcher-and-system-calls.md
```diff
@@ -84,4 +84,4 @@ Using these syscalls can be handy for customized error handling or to get more c
 
 <span class="caption">Listing 99-9: A sample contract using syscalls</span>
 
-To use this syscall, we passed in the contract address, the function selector (which is the `starknet_keccak` hash of the function name), and the calldata (function arguments). At the end, we get returned a serialized value which we'll need to deserialize ourselves!
+To use this syscall, we passed in the contract address, the calldata (function arguments), which then inside the function gets serialized and is passed on to the array which is spanned and passed as the calldata. The function name is already passed in which is transferFrom which we will be calling, At the end, we are returned a serialized value which we'll need to deserialize ourselves!
```
enitrat, line 87:
```suggestion
To use this syscall, we passed in the contract address, the selector of the function we want to call, and the call arguments.

The call arguments must be provided as an array of `felt252`. To build this array, we serialize the expected function parameters into an `Array<felt252>` using the `Serde` trait, and then pass this array as calldata. At the end, we are returned a serialized value which we'll need to deserialize ourselves!
```

listings/ch99-starknet-smart-contracts/listing_99_09_call_contract_syscall/src/lib.cairo
```diff
@@ -1,31 +1,43 @@
+use starknet::ContractAddress;
 #[starknet::interface]
 trait ITokenWrapper<TContractState> {
     fn transfer_token(
         ref self: TContractState,
         address: starknet::ContractAddress,
```
enitrat, line 6: as you imported ContractAddress:
```suggestion
        address: ContractAddress,
```

enitrat (commented):
great!

src/ch99-02-02-contract-dispatcher-library-dispatcher-and-system-calls.md
Akashneelesh: Done

listings/ch99-starknet-smart-contracts/listing_99_09_call_contract_syscall/src/lib.cairo
Akashneelesh: Done

listings/ch99-starknet-smart-contracts/listing_99_09_call_contract_syscall/src/lib.cairo
Akashneelesh: Done

Akashneelesh:
I have acknowledged all the changes and have updated it.

enitrat (approved):
image

Issue #419, opened 2023-10-31
Add a subsection about enum storage in the Storage chapter


enitrat:
It shouldn't take longer than 1 hour for an intermediate cairo dev

OkoliEvans:
Let me make some research on this and work on it @enitrat

enitrat:
Sure!

enitrat:
Just FYI enums can be stored by deriving starknet::Store on them, but what is interesting is to give information about the generated storage layout and how variants are stored/retrieved.

OkoliEvans:
Okay

OkoliEvans:
I have submitted a PR on this, here: https://github.com/cairo-book/cairo-book/pull/438
@enitrat

PR #424, opened 2023-11-02
Updated Chapter 12.3.1 

OkoliEvans:
The chapter did not explicitly explain how user defined structs are written to Storage, I found this might be confusing for newbies so I added some clarifications.
@enitrat

### DIFF
--- a/src/ch99-01-03-01-storage-variables.md
+++ b/src/ch99-01-03-01-storage-variables.md
@@ -30,6 +30,11 @@ The base storage address for structs remains `sn_keccak(variable_name)`, and the
 
 Subsequent fields are stored in addresses contiguous to the first elements at addresses `base_address + i`.
 
+## Writing to stored structs
+
+To write custom storage structs to storage you have to first create the struct variant as a member of the default storage struct. The type of the variant is
+the custom storage struct.
+
 ### Storing mappings
 
 Mappings are a key-value data structure that you can use to store data within a smart contract. They are essentially hash tables that allow you to associate a unique key with a corresponding value. Mappings are also useful to store sets of data, as it's impossible to store arrays in storage.
--- a/src/ch99-01-03-01-storage-variables.md
+++ b/src/ch99-01-03-01-storage-variables.md
@@ -30,10 +30,11 @@ The base storage address for structs remains `sn_keccak(variable_name)`, and the
 
 Subsequent fields are stored in addresses contiguous to the first elements at addresses `base_address + i`.
 
-## Writing custom storage structs to storage
 
-To write custom storage structs to storage you have to first create the struct variant as a member of the default storage struct. The type of the variant is
-the custom storage struct.
+## Writing custom structs to storage
+
+To write custom struct to storage you have to first add the struct variant as a member of the default storage struct. The type of the variant is
+the custom struct.
 
 ```rust, noplayground
     #[storage]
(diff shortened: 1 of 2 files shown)


enitrat:
> The chapter did not explicitly explain how user defined structs are written to Storage

Yes and it's still on purpose, it's not just user defined structs it's all structs, even u256 values, which are structs, are written in the same way to storage.

What I am opposed to is the _specialization_ of the explanation specifically for structs. For all types stored, there is one single common way of reading and writing them, it's not specific to structs. The only specificity for structs is that you need to derive starknet::Store on them

src/ch99-01-03-01-storage-variables.md
```diff
@@ -30,6 +30,21 @@ The base storage address for structs remains `sn_keccak(variable_name)`, and the
 
 Subsequent fields are stored in addresses contiguous to the first elements at addresses `base_address + i`.
 
+
+## Writing user defined structs to storage
```
enitrat, line 34: Therefore you can remove this explanation as it doesn't make sense to have this for structs explicitly.

src/ch99-01-03-01-storage-variables.md
```diff
@@ -30,6 +30,21 @@ The base storage address for structs remains `sn_keccak(variable_name)`, and the
...
+
+To write user defined struct to storage you have to first add the struct variables to the default Storage struct. The variable type is the user defined struct identifier.
```
enitrat, line 36: It's not just structs, it's every single storage variable!

enitrat (changes requested):

src/ch99-01-03-01-storage-variables.md
OkoliEvans: Okay, I get it now

OkoliEvans:
I made the changes, but I'm not sure the wordings are perfect. If there's need to still edit the wordings, some suggestions will really help... like a sample text with the right words / structure. @enitrat

enitrat:
Here are some leads on how to proceed:

- Do not mention storing user-defined structs _at all_. As there is no special treatment for user-defined structs compared to other types in storage, we will not explicitly talk about how they're stored.
- Replace all integrations of code in this page by specific corresponding snippets from `listings/ch99-starknet-smart-contracts/listing_99_03_example_contract`. This ensure that the explanations are in line with the example introduced in the previous chapter.
- If you believe that it's not clear that variables must be added to the storage struct to be considered in storage, you can eventually reword this part. But I do think it's clear enough 🤔 

```rust
Storage variables in Starknet contracts are stored in a special struct called `Storage`:

```rust, noplayground
{{#rustdoc_include ../listings/ch99-starknet-smart-contracts/listing_99_02/src/lib.cairo:here}}
```

<span class="caption">Listing 99-2: A Storage Struct</span>

The storage struct is a [struct](./ch05-00-using-structs-to-structure-related-data.md) like any other,
except that it **must** be annotated with `#[storage]` allowing you to store mappings using the `LegacyMap` type.
```

OkoliEvans:
New changes made @enitrat

enitrat:
I will close this as I don't want to spend too much time on back-and-fourths review. 
I opened another PR with the changes that I would have liked to see in this section, if you wanted to improve the clarity and organization #427

For #419, if you still intend to do it, you can add your part under the `### Structs storage layout`, named `### Enums storage layout`

OkoliEvans:
I will continue on the Enums then.

PR #434, opened 2023-11-13
Book minor edit

Utilitycoder:
Deleted a repeated word in chapter 12.


enitrat:
can you clean your PR and the git history?

Utilitycoder:
Will do. I just found out the mess now.

PR #445, opened 2023-11-21
feat: add typos-check in CI & fix typos

enitrat:


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Scarb
         uses: software-mansion/setup-scarb@v1
         with:
-          scarb-version: "2.3.0-rc1"
+          scarb-version: 2.3.0-rc1
```
julio4, line 19: You can remove this, it will default to the version specified in `.tool-versions`. Easier to upgrade.

xiaolou86:
Nice, it's a good job to add typos-check workflow in CI.

.github/workflows/ci.yml
enitrat: Indeed, nice

Issue #382, opened 2023-09-30
Add a section on the Copy mechanism


enitrat:

enitrat:
@wraitii I heard you would like to take this issue?

wraitii:
![Sans-titre-2023-11-09-1039](https://github.com/cairo-book/cairo-book/assets/1927071/913debcc-e115-4887-821f-f106ecb69da2)

Started an Excalidraw: https://excalidraw.com/#room=e49f594e9f284668b8ea,Eud7vQQefDRFD7GQ-EO0RQ (and see image).

I think the core issue is that Cairo, unlike rust (but like Python / javascript /...) has 'fat-pointer' variables, instead of just being memory-cells.

If you explain that `Copy` in rust refers to the memory being copyable, but in Cairo refers to the variable being copyable, things make a little more sense.

I don't have much time right now, but might try to work on this at Istanbul / later

PR #449, opened 2023-11-23
Rewrite Ch04 - What is Ownership

wraitii:
Addresses #382 

Clarify what the linear type system is for.
Clarify what moving and destroying means.
Clarify the semantics of `Copy`, and specifically introduce a distinction between 'variables' and 'values'.
Clarify the cost of the Copy and desnap operators.
Reorganise a few elements.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/449)

### DIFF
--- a/src/ch02-01-variables-and-mutability.md
+++ b/src/ch02-01-variables-and-mutability.md
@@ -7,8 +7,8 @@ However, the language abstracts this model and gives you the option to make your
 variables mutable. Let’s explore how and why Cairo enforces immutability, and how
 you can make your variables mutable.
 
-When a variable is immutable, once a value is bound to a name, you can’t change
-that value. To illustrate this, generate a new project called _variables_ in
+When a variable is immutable, once a variable is bound to a value, you can’t change
+that variable. To illustrate this, generate a new project called _variables_ in
 your _cairo_projects_ directory by using `scarb new variables`.
 
 Then, in your new _variables_ directory, open _src/lib.cairo_ and replace its
@@ -41,27 +41,23 @@ not a good programmer! Experienced Caironautes still get compiler errors.
 You received the error message `Cannot assign to an immutable variable.`
 because you tried to assign a second value to the immutable `x` variable.
 
-It’s important that we get compile-time errors when we attempt to change a
-value that’s designated as immutable because this specific situation can lead to
-bugs. If one part of our code operates on the assumption that a value will
-never change and another part of our code changes that value, it’s possible
-that the first part of the code won’t do what it was designed to do. The cause
-of this kind of bug can be difficult to track down after the fact, especially
-when the second piece of code changes the value only _sometimes_. The Cairo
-compiler guarantees that when you state that a value won’t change, it really
-won’t change, so you don’t have to keep track of it yourself. Your code is thus
-easier to reason through.
+Cairo, unlike most other languages, has immutable memory. This makes a
+whole class of bugs impossible, because values will never change unexpectedly.
+This makes code easier to reason about.
 
 But mutability can be very useful, and can make code more convenient to write.
 Although variables are immutable by default, you can make them mutable by
 adding `mut` in front of the variable name. Adding `mut` also conveys
 intent to future readers of the code by indicating that other parts of the code
-will be changing this variable’s value.
+will be changing this variable.
 
 However, you might be wondering at this point what exactly happens when a variable
 is declared as `mut`, as we previously mentioned that Cairo's memory is immutable.
-The answer is that Cairo's memory is immutable, but the memory address the variable points
-to can be changed. Upon examining the low-level Cairo Assembly code, it becomes clear that
+The answer is that the _value_ is immutable, but the _variable_ isn't. What value
+the variable points to can be changed. Assigning to a mutable variable in Cairo
+is essentially equivalent to redeclaring it to point to another value in another memory cell,
+but the compiler handles that for you, and the keyword `mut` makes it explicit.
+Upon examining the low-level Cairo Assembly code, it becomes clear that
 variable mutation is implemented as syntactic sugar, which translates mutation operations
 into a series of steps equivalent to variable shadowing. The only difference is that at the Cairo
 level, the variable is not redeclared so its type cannot change.
--- a/src/ch02-01-variables-and-mutability.md
+++ b/src/ch02-01-variables-and-mutability.md
@@ -41,6 +41,14 @@ not a good programmer! Experienced Caironautes still get compiler errors.
 You received the error message `Cannot assign to an immutable variable.`
 because you tried to assign a second value to the immutable `x` variable.
 
+It’s important that we get compile-time errors when we attempt to change a
+value that’s designated as immutable because this specific situation can lead to
+bugs. If one part of our code operates on the assumption that a value will
+never change and another part of our code changes that value, it’s possible
+that the first part of the code won’t do what it was designed to do. The cause
+of this kind of bug can be difficult to track down after the fact, especially
+when the second piece of code changes the value only _sometimes_.
+
 Cairo, unlike most other languages, has immutable memory. This makes a
 whole class of bugs impossible, because values will never change unexpectedly.
 This makes code easier to reason about.
@@ -49,13 +57,15 @@ But mutability can be very useful, and can make code more convenient to write.
 Although variables are immutable by default, you can make them mutable by
 adding `mut` in front of the variable name. Adding `mut` also conveys
 intent to future readers of the code by indicating that other parts of the code
-will be changing this variable.
+will be changing the value associated to this variable.
+
+<!-- TODO: add an illustration of this -->
 
 However, you might be wondering at this point what exactly happens when a variable
 is declared as `mut`, as we previously mentioned that Cairo's memory is immutable.
 The answer is that the _value_ is immutable, but the _variable_ isn't. What value
 the variable points to can be changed. Assigning to a mutable variable in Cairo
-is essentially equivalent to redeclaring it to point to another value in another memory cell,
+is essentially equivalent to redeclaring it to refer to another value in another memory cell,
 but the compiler handles that for you, and the keyword `mut` makes it explicit.
 Upon examining the low-level Cairo Assembly code, it becomes clear that
 variable mutation is implemented as syntactic sugar, which translates mutation operations
--- a/src/ch04-01-what-is-ownership.md
+++ b/src/ch04-01-what-is-ownership.md
@@ -3,17 +3,19 @@
 Cairo uses a linear type system. In such a type system, any value (a basic type, a struct, an enum) must be used and must only be used once. 'Used' here means that the value is either _destroyed_ or _moved_.
 
 _Destruction_ can happen in several ways:
- - a variable goes out of scope
- - a struct is destructured
- - explicit destruction using destruct()
+
+- a variable goes out of scope
+- a struct is destructured
+- explicit destruction using destruct()
 
 _Moving_ a value simply means passing that value to another function.
 
 This results in somewhat similar constraints to the Rust ownership model, but there are some differences.
 In particular, the rust ownership model exists (in part) to avoid data races and concurrent mutable access to a memory value. This is obviously impossible in Cairo since the memory is immutable.
 Instead, Cairo leverages its linear type system for two main purposes:
- - Ensuring that all code is provable and thus verifiable.
- - Abstracting away the immutable memory of the Cairo VM.
+
+- Ensuring that all code is provable and thus verifiable.
+- Abstracting away the immutable memory of the Cairo VM.
 
 #### Ownership
 
@@ -22,9 +24,10 @@ Variables however can be mutable, so the compiler must ensure that constant vari
 This makes it possible to talk about ownership of a variable: the owner is the code that can read (and write if mutable) the variable.
 
 This means that variables (not values) follow similar rules to Rust values:
- - Each variable in Cairo has an owner.
- - There can only be one owner at a time.
- - When the owner goes out of scope, the variable is destroyed.
+
+- Each variable in Cairo has an owner.
+- There can only be one owner at a time.
+- When the owner goes out of scope, the variable is destroyed.
 
 Now that we’re past basic Cairo syntax, we won’t include all the `fn main() {`
 examples inside a `main` function manually. As a result, our examples will be a
@@ -118,6 +121,7 @@ For example, the following code will not compile, because the struct `A` is not 
 ```rust,does_not_compile
 {{#include ../listings/ch04-understanding-ownership/no_listing_04_no_drop_derive_fails/src/lib.cairo}}
 ```
+
 However, types that implement the `Drop` trait are automatically destroyed when going out of scope. This destruction does nothing, it is a no-op - simply a hint to the compiler that this type can safely be destroyed once it's no longer useful. We call this "dropping" a value.
 
 At the moment, the `Drop` implementation can be derived for all types, allowing them to be dropped when going out of scope, except for dictionaries (`Felt252Dict`) and types containing dictionaries.
@@ -128,6 +132,7 @@ For example, the following code compiles:
 ```
 
 #### Destruction with a side-effect: the Destruct trait
+
 When a value is destroyed, the compiler first tries to call the `drop` method on that type. If it doesn't exist, then the compiler tries to call `destruct` instead. This method is provided by the `Destruct` trait.
 
 As said earlier, dictionaries in Cairo are types that must be "squashed" when destructed, so that the sequence of access can be proven. This is easy for developers to forget, so instead dictionaries implement the `Destruct` trait to ensure that all dictionaries are _squashed_ when going out of scope.
@@ -163,9 +168,9 @@ Here’s an example of the `clone` method in action.
 ```rust
 {{#include ../listings/ch04-understanding-ownership/no_listing_08_array_clone/src/lib.cairo}}
 ```
-When you see a call to `clone`, you know that some arbitrary code is being executed and that code may be expensive. It’s a visual indicator that something different is going on.
-In this case, both the _variable_ and the _value_ are being copied, resulting in new memory cells being used.
 
+When you see a call to `clone`, you know that some arbitrary code is being executed and that code may be expensive. It’s a visual indicator that something different is going on.
+In this case, _value_ is being copied, resulting in new memory cells being used, and the a new _variable_ is created, referring to the new, copied value.
 
 ### Return Values and Scope
 
--- a/src/ch04-01-what-is-ownership.md
+++ b/src/ch04-01-what-is-ownership.md
@@ -1,10 +1,6 @@
 ## Ownership Using a Linear Type System
 
-Cairo uses a linear type system for two main reasons:
- - Ensuring that all code is provable and thus verifiable.
- - Abstracting away the immutable memory of the Cairo VM.
-
-In a linear type system, any value (a basic type, a struct, an enum) must be used and must only be used once. 'Used' here means that the value is either _destroyed_ or _moved_.
+Cairo uses a linear type system. In such a type system, any value (a basic type, a struct, an enum) must be used and must only be used once. 'Used' here means that the value is either _destroyed_ or _moved_.
 
 _Destruction_ can happen in several ways:
  - a variable goes out of scope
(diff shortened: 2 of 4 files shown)


src/ch04-01-what-is-ownership.md
```diff
@@ -1,90 +1,86 @@
...
 
 ```rust
 {{#rustdoc_include ../listings/ch04-understanding-ownership/no_listing_01_array/src/lib.cairo:2:4}}
 ```
 
-So, how does the ownership system ensure that each cell is never written to more than once?
-Consider the following code, where we try to pass the same instance of an array in two consecutive
-function calls:
+How does the type system ensure that the Cairo program never tries to write to the same memory cell twice?
+Consider the following code, where we try to remove the front of the array twice:
```
wraitii, line 75: Seemed more obvious to me but YMMY

src/ch04-01-what-is-ownership.md
```diff
@@ -162,49 +156,27 @@ When A goes out of scope, it can't be dropped as it implements neither the `Drop
...
-{{#include ../listings/ch04-understanding-ownership/listing_03_03/src/lib.cairo}}
-```
-
-<span class="caption">Listing 4-3: Functions with ownership and scope
-annotated</span>
-
-If we tried to use `my_struct` after the call to `takes_ownership`, Cairo would throw a
-compile-time error. These static checks protect us from mistakes. Try adding
-code to `main` that uses `my_struct` and `x` to see where you can use them and where
-the ownership rules prevent you from doing so.
```
wraitii, line 203: I removed this 'caused it seemed pretty un-necessary, but can be kept

src/ch04-01-what-is-ownership.md
```diff
@@ -162,49 +156,27 @@ When A goes out of scope, it can't be dropped as it implements neither the `Drop
...
 
 ### Return Values and Scope
 
-Returning values can also transfer ownership. Listing 4-4 shows an example of a
+Returning values is equivalent to _moving_ them. Listing 4-4 shows an example of a
```
wraitii, line 179: I didn't mention ownership in my initial draft but since I kept the lines in the end the OG formulation is maybe cleaner

src/ch02-01-variables-and-mutability.md
```diff
@@ -41,27 +41,23 @@ not a good programmer! Experienced Caironautes still get compiler errors.
...
+Cairo, unlike most other languages, has immutable memory. This makes a
+whole class of bugs impossible, because values will never change unexpectedly.
+This makes code easier to reason about.
 
 But mutability can be very useful, and can make code more convenient to write.
 Although variables are immutable by default, you can make them mutable by
 adding `mut` in front of the variable name. Adding `mut` also conveys
 intent to future readers of the code by indicating that other parts of the code
-will be changing this variable’s value.
+will be changing this variable.
```
enitrat, line 52: idk about this - variables are still aliases to values, so i'd keep something like 
```suggestion
will be changing the value associated to this variable.
```

src/ch02-01-variables-and-mutability.md
```diff
@@ -41,27 +41,23 @@ not a good programmer! Experienced Caironautes still get compiler errors.
...
 
 However, you might be wondering at this point what exactly happens when a variable
 is declared as `mut`, as we previously mentioned that Cairo's memory is immutable.
-The answer is that Cairo's memory is immutable, but the memory address the variable points
-to can be changed. Upon examining the low-level Cairo Assembly code, it becomes clear that
+The answer is that the _value_ is immutable, but the _variable_ isn't. What value
+the variable points to can be changed. Assigning to a mutable variable in Cairo
+is essentially equivalent to redeclaring it to point to another value in another memory cell,
```
enitrat, line 58:
```suggestion
is essentially equivalent to redeclaring it to refer to to another value in another memory cell,
```

I'd be careful with the word `point` as it's often associated with pointers - which is not really the case

src/ch02-01-variables-and-mutability.md
```diff
@@ -41,27 +41,23 @@ not a good programmer! Experienced Caironautes still get compiler errors.
...
+but the compiler handles that for you, and the keyword `mut` makes it explicit.
+Upon examining the low-level Cairo Assembly code, it becomes clear that
```
enitrat, line 60: maybe we should add an illustration? 
I can do it later if you want - just add a TODO line & create an issue

src/ch02-01-variables-and-mutability.md
```diff
@@ -41,27 +41,23 @@ not a good programmer! Experienced Caironautes still get compiler errors.
...
-never change and another part of our code changes that value, it’s possible
-that the first part of the code won’t do what it was designed to do. The cause
-of this kind of bug can be difficult to track down after the fact, especially
-when the second piece of code changes the value only _sometimes_. The Cairo
-compiler guarantees that when you state that a value won’t change, it really
-won’t change, so you don’t have to keep track of it yourself. Your code is thus
-easier to reason through.
+Cairo, unlike most other languages, has immutable memory. This makes a
+whole class of bugs impossible, because values will never change unexpectedly.
+This makes code easier to reason about.
```
enitrat, line 46: While I agree with this I think the previous paragraph still had value in this context. Even if Cairo has an immutable memory, the Cairo language totally abstracts this concept. Perhaps you can place the previous paragraph it after the paragraph on mutability? wdyt

src/ch04-01-what-is-ownership.md
```diff
@@ -1,90 +1,86 @@
-## What Is Ownership?
+## Ownership Using a Linear Type System
 
-Cairo implements an ownership system to ensure the safety and correctness of its compiled code.
-The ownership mechanism complements the linear type system, which enforces that objects are used exactly once.
-This helps prevent common operations that can produce runtime errors, such as illegal memory address
-references or multiple writes to the same memory address, and ensures the soundness of Cairo programs
-by checking at compile time that all the dictionaries are squashed.
+Cairo uses a linear type system for two main reasons:
```
enitrat, line 3: You need to succinctly introduce what a linear type system is. Perhaps add some reasoning about programming paradigms?

src/ch04-01-what-is-ownership.md
```diff
@@ -1,90 +1,86 @@
...
+ - explicit destruction using destruct()
+
+_Moving_ a value simply means passing that value to another function.
+
+This results in somewhat similar constraints to the Rust ownership model, but there are some differences.
+In particular, the rust ownership model exists (in part) to avoid data races and concurrent mutable access to a memory value. This is obviously impossible in Cairo since the memory is immutable.
+
+#### Ownership
+
+In Cairo, ownership applies to _variables_ and not to _values_. A value can safely be pointed to by many different variables (even if they are mutable variables), as the value itself is always immutable.
```
enitrat, line 21:
```suggestion
In Cairo, ownership applies to _variables_ and not to _values_. A value can safely be referred to by many different variables (even if they are mutable variables), as the value itself is always immutable.
```

src/ch04-01-what-is-ownership.md
```diff
@@ -1,90 +1,86 @@
...
 {{#rustdoc_include ../listings/ch04-understanding-ownership/no_listing_01_array/src/lib.cairo:2:4}}
 ```
 
-So, how does the ownership system ensure that each cell is never written to more than once?
-Consider the following code, where we try to pass the same instance of an array in two consecutive
-function calls:
+How does the type system ensure that the Cairo program never tries to write to the same memory cell twice?
+Consider the following code, where we try to remove the front of the array twice:
 
+//// CHANGEME
```
enitrat, line 77: still needs to be changed?

src/ch04-01-what-is-ownership.md
enitrat: I think this would need some more details about the inner construction of the array type. If you know that it's a tuple of pointers, then as discussed yesterday you can easily figure out that arrays referred to in two variables are not synced if arrays can be 'copy'

src/ch04-01-what-is-ownership.md
```diff
@@ -1,90 +1,86 @@
...
-In this case, we try to pass the same array instance `arr` by value to the functions `foo` and `bar`, which means
-that the parameter used in both function calls is the same instance of the array. If you append a value to the array
-in `foo`, and then try to append another value to the same array in `bar`, what would happen is that
-you would attempt to try to write to the same memory cell twice, which is not allowed in Cairo.
-To prevent this, the ownership of the `arr` variable moves from the `main` function to the `foo` function. When trying
-to call `bar` with `arr` as a parameter, the ownership of `arr` was already moved to the first call. The ownership
-system thus prevents us from using the same instance of `arr` in `foo`.
-
-Running the code above will result in a compile-time error:
+In this case, we try to pass the same value (the array in `arr`) to both function calls. This means our code tries to remove the first element twice, which would try to write to the same memory cell twice - which is forbidden by the Cairo VM, leading to a runtime error.
```
enitrat, line 82:
```suggestion
In this case, we try to pass the same value (the array in the `arr` variable) to both function calls. This means our code tries to remove the first element twice, which would try to write to the same memory cell twice - which is forbidden by the Cairo VM, leading to a runtime error.
```

src/ch04-01-what-is-ownership.md
```diff
@@ -1,90 +1,86 @@
...
+Thankfully, this code does not actually compile. Once we have passed the array, the variable is no longer usable. We get this compile-time error, telling us that we would need Array to implement the Copy Trait:
```
enitrat, line 83:
```suggestion
Thankfully, this code does not actually compile. Once we have passed the array to the `foo` function, the `arr` variable is no longer usable. We get this compile-time error, telling us that we would need Array to implement the Copy Trait:
```

src/ch04-01-what-is-ownership.md
```diff
@@ -93,53 +89,51 @@ error: Variable was previously moved. Trait has no implementation in context: co
...
-If a type implements the `Copy` trait, passing it to a function will not move the ownership of the value to the function called, but will instead pass a copy of the value.
 You can implement the `Copy` trait on your type by adding the `#[derive(Copy)]` annotation to your type definition. However, Cairo won't allow a type to be annotated with Copy if the type itself or any of its components don't implement the Copy trait.
 While Arrays and Dictionaries can't be copied, custom types that don't contain either of them can be.
 
 ```rust,ignore_format
 {{#include ../listings/ch04-understanding-ownership/no_listing_03_copy_trait/src/lib.cairo}}
 ```
 
-In this example, we can pass `p1` twice to the foo function because the `Point` type implements the `Copy` trait. This means that when we pass `p1` to `foo`, we are actually passing a copy of `p1`, and the ownership of `p1` remains with the main function.
+In this example, we can pass `p1` twice to the foo function because the `Point` type implements the `Copy` trait. This means that when we pass `p1` to `foo`, we are actually passing a copy of `p1`, so `p1` remains valid.
```
enitrat, line 104: should we really remove all mentions of ownership here?

src/ch04-01-what-is-ownership.md
```diff
@@ -93,53 +89,51 @@ error: Variable was previously moved. Trait has no implementation in context: co
...
 If you remove the `Copy` trait derivation from the `Point` type, you will get a compile-time error when trying to compile the code.
 
 _Don't worry about the `Struct` keyword. We will introduce this in [Chapter 5](ch05-00-using-structs-to-structure-related-data.md)._
 
-### The `Drop` Trait
+### Destroying values - example with FeltDict
+
+The other way linear types can be _used_ is by being destroyed. Destruction must ensure that the 'resource' is now correctly released. In rust for example, this could be closing the access to a file, or locking a mutex.
+In Cairo, one type that has such behaviour is felt252Dict. For provability, dicts must be 'squashed' when they are destructed.
```
enitrat, line 112:
```suggestion
In Cairo, one type that has such behaviour is `Felt252Dict`. For provability, dicts must be 'squashed' when they are destructed.
```

src/ch04-01-what-is-ownership.md
```diff
@@ -93,53 +89,51 @@ error: Variable was previously moved. Trait has no implementation in context: co
...
 
 ```rust
 {{#include ../listings/ch04-understanding-ownership/no_listing_05_drop_derive_compiles/src/lib.cairo}}
 ```
 
-### The `Destruct` Trait
-
-Manually calling the `squash` method on a dictionary is not very convenient, and it is easy to forget to do so. To make it easier to use dictionaries, Cairo provides the `Destruct` trait, which allows you to specify the behavior of a type when it goes out of scope. While Dictionaries don't implement the `Drop` trait, they do implement the `Destruct` trait, which allows them to automatically be `squashed` when they go out of scope. This means that you can use dictionaries without having to manually call the `squash` method.
+#### Destruction with a side-effect: the Destruct trait
+When a value is destroyed, the compiler calls the `destruct` method on that type. This method is provided by the `Destruct` trait. This is actually how `Drop` is implemented at the moment, by providing a `Destruct` impl that does nothing.
```
enitrat, line 133: > This is actually how `Drop` is implemented at the moment, by providing a `Destruct` impl that does nothing.

really?

src/ch04-01-what-is-ownership.md
```diff
@@ -93,53 +89,51 @@ error: Variable was previously moved. Trait has no implementation in context: co
...
 
-Consider the following example, in which we define a custom type that contains a dictionary:
+As said earlier, dictionaries in Cairo are types that must be "squashed" when destructed, so that the sequence of access can be proven. This is easy for developers to forget, so instead dictionaries implement the Destruct Trait.
```
enitrat, line 135:
```suggestion
As said earlier, dictionaries in Cairo are types that must be "squashed" when destructed, so that the sequence of access can be proven. This is easy for developers to forget, so instead dictionaries implement the `Destruct` trait to ensure that all dictionaries are _squashed_ when going out of scope.
```

src/ch04-01-what-is-ownership.md
```diff
@@ -162,49 +156,27 @@ When A goes out of scope, it can't be dropped as it implements neither the `Drop
 
 Now, when `A` goes out of scope, its dictionary will be automatically `squashed`, and the program will compile.
 
+#### Copy vs Destruct
+
+You may notice that these two traits are somewhat incompatible. With a `Copy` type, we could have two variables pointing to the same value, which would lead to a double-destruction when both variables go out of scope. This is obviously quite broken for any type with a non-trivial `Destruct` implementation.
```
enitrat, line 161: I didn't follow on this

src/ch04-02-references-and-snapshots.md
```diff
@@ -7,12 +7,15 @@ call to `calculate_length`, because the `Array` was moved into
...
-appending values to arrays. However, it's not very convenient. Let's see how we can retain ownership
-of the value in the calling function using snapshots.
+In the previous chapter, we talked about how Cairo's ownership system prevents
+us from using a variable after we've moved it, protecting us from potentially
+writing twice to the same memory cell. However, it's not very convenient.
+Let's see how we can retain ownership of the variable in the calling function using snapshots.
+
+In Cairo, a snapshot is an immutable view of a value at a certain point in time.
+Recall that memory is immutable, so modifying a value actually creates a new memory cell.
+The old memory cell still exists, and snapshots are variables that point to that "old" value.
```
enitrat, line 17:
```suggestion
The old memory cell still exists, and snapshots are variables that refer to that "old" value.
```

enitrat (commented):
Great improvement - it's very clear. 
Just a few comments.

However, I think we should keep talking about ownership. It's a good way of representing which values belongs to what and talking about ownership when talking about moving variables to different "owners" would make sense

src/ch04-01-what-is-ownership.md
wraitii: yup, though there's a TODO to remove that. Check https://github.com/starkware-libs/cairo/blob/main/corelib/src/traits.cairo#L138

My understanding is that this might be changed in the future, but IMO it's clearer to always talk about destroying the value, with drop as a special no-op case

enitrat: I think this is only an impl that allows you to derive `Destruct` on a type that has Drop. But it doesn't inform us about what the compiler is actually doing. 

I think the compiler first tries to drop, then if it's not droppable tries to destruct.

see https://github.com/starkware-libs/cairo/blob/13f4c6aa249143e5cca51450b0744f2a646a5bf0/crates/cairo-lang-lowering/src/destructs.rs#L54-L94

wraitii: Mh, then the TODO is probably redundant at this point.

But it doesn't change much - the fact is that conceptually drop and destroy are the same operation, it's just "destroying" the value. This can have side-effects of not.

enitrat: I don't agree - Destruct is just a drop with extra steps. You squash a dict, you move the dict to the squash funciton, you receive a SquashedDict, which is droppable and goes out of scope seamlessly

wraitii: hm, sure, but that's not the cairo semantics. Would have been neat though.

src/ch04-01-what-is-ownership.md
wraitii: I think the previous version had the same problem though, maybe reversed?
I'm not sure which is most obviously broken

src/ch04-01-what-is-ownership.md
wraitii: `Destruct` is at the variable scope, not the value-scope. So if you have a `Copy` type, there can be two variables, and thus two `Destruct` runs for the same 'value'.
This is pretty weird and IMO a massive code smell, but not technically broken so I guess I'll change the wording.

src/ch04-01-what-is-ownership.md
enitrat: yes, I think we can just remove this sentence. 

Or you can add precision: The compiler calls `drop` on that type. If the type doesn't implement the `Drop` trait, it calls the `Destruct` trait

src/ch04-01-what-is-ownership.md
enitrat: ah ok, yes the wording makes it hard to understand

enitrat: also the title `Copy vs Destruct` is not so much adapted

src/ch02-01-variables-and-mutability.md
wraitii: Added the early part back, so now it just emphasizes a little more.

src/ch02-01-variables-and-mutability.md
wraitii: There's my graph on excalidraw, but I'll do the TODO thing for now

src/ch04-01-what-is-ownership.md
wraitii: Reworked a little as I had an explanation after this

src/ch04-01-what-is-ownership.md
wraitii: Added back

enitrat (changes requested):
> Reviewed 1 of 4 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/449)* status: all files reviewed, 6 unresolved discussions (waiting on @wraitii)

___
*[`src/ch04-01-what-is-ownership.md` line 167 at r2](https://reviewable.io/reviews/cairo-book/cairo-book/449#-NlE2Rrl8jDnJuSIlKvN:-NlE2Rrl8jDnJuSIlKvO:b4rr478) ([raw file](https://github.com/cairo-book/cairo-book/blob/7c6a72a462bd1125aaa0c46bb73cd65817c23612/src/ch04-01-what-is-ownership.md#L167)):*
> ```Markdown
> ```
> When you see a call to `clone`, you know that some arbitrary code is being executed and that code may be expensive. It’s a visual indicator that something different is going on.
> In this case, both the _variable_ and the _value_ are being copied, resulting in new memory cells being used.
> ```

cloned instead of copied?



_Suggestion:_
```Markdown
In this case, both the _variable_ and the _value_ are being cloned, resulting in new memory cells being used.
```

enitrat (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/449#-:-NlE3t_Y4XSeNorpoorS:bnfp4nl)* status: 3 of 4 files reviewed, 6 unresolved discussions (waiting on @wraitii)

PR #455, opened 2023-11-30
Link to starknet documentation storage variables was broken

Eckmoule:
Hi, 
Looks like the link was broken. 
I didn't regenerate the corresponding translation files (don't know if I should).


enitrat (approved):

enitrat:
Thanks :)

PR #460, opened 2023-12-04
Update zh-cn.po Translated Storage Mapping

shepf:
Translated Storage Mapping

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/460)


enitrat:
@shepf ?

PR #464, opened 2023-12-07
add README-CN.md

shepf:


shepf:
> Why did you set the english README in Chinese?

original English README still exists.  it provides flexibility for readers who are more comfortable with different languages.

enitrat:
yes but you moved the english to README-CN and README.md is now in chinese

shepf:
> yes but you moved the english to README-CN and README.md is now in chinese

Oh, I'm sorry. It seems like I got it mixed up. Please wait a moment while I fix it.

PR #465, opened 2023-12-07
fix add README-CN.md

shepf:
fix add README-CN.md

### DIFF
--- a/README-CN.md
+++ b/README-CN.md
@@ -1,4 +1,4 @@
-English | [简体中文](README-CN.md)
+[English](README.md) | 简体中文
 
 <div align="center">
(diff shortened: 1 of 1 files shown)


shepf:
@enitrat
please check, thanks

enitrat:
>  I believe that the README-CN.md file should be at the same level as the README.md file.

I want top-level files to only be for source content, in english. Translations are not part of the main code and thus should be in subfolders. please move to `translations/README-cn.md

shepf:
> > I believe that the README-CN.md file should be at the same level as the README.md file.
> 
> I want top-level files to only be for source content, in english. Translations are not part of the main code and thus should be in subfolders. please move to `translations/README-cn.md`

ok~

PR #468, opened 2023-12-07
Update ch01-02-hello-world.md

thejdavid:
fix old version resulting in error 

error: Identifier not found.
error: Method `print` not found on type "?0". Did you import the correct trait and impl?

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/468)

### DIFF
--- a/src/ch01-02-hello-world.md
+++ b/src/ch01-02-hello-world.md
@@ -65,7 +65,7 @@ Open _Scarb.toml_ in your text editor of choice. It should look similar to the c
 ```toml
 [package]
 name = "hello_world"
-version = "0.1.0"
+version = "2.4.0"
 
 # See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest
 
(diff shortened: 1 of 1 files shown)


src/ch01-02-hello-world.md
```diff
@@ -65,7 +65,7 @@ Open _Scarb.toml_ in your text editor of choice. It should look similar to the c
 ```toml
 [package]
 name = "hello_world"
-version = "0.1.0"
+version = "2.4.0"
```
enitrat, line 68: this field indicates the version of the package, not the version of Cairo to use

enitrat (commented):
I don't understand. What was your issue? The book is meant to be used with the version specified in the introduction and I don't see what the problem could have been here.

enitrat:
> got it fixed by aligning my versions with the ones installed locally so i was assuming the issue was the difference between the latest scarb / cairo version installed and the ones from the package since that got the issue fixed.

Hmm, unless you were running a super old version of Cairo, this shouldn't have happened 🤔 It's weird 🤔

Issue #453, opened 2023-11-25
feat: 2.3.0


enitrat:
## Cairo

- [x] Anonymous generic impl parameters](https://community.starknet.io/t/cairo-v2-3-0-release-candidate/101137#anonymous-generic-impl-parameters-3)
- [x] #638
- [x] String literals! prepare for 2.4 release with display/debug traits
 
```rust
use debug::PrintTrait;
fn main() {
    let x: ByteArray = "this is a string literal";
    x[0].print();
}
```
- [ ]

## Starknet

- [x] [embeddable impls](https://community.starknet.io/t/cairo-v2-3-0-release-candidate/101137#impl-alias-4)
- [x] #462
- [x] #461
- [x] Relaxed event enum names (no longer required to match enum variant name)
- [x] dep_component{_mut}! macros - for #452

PR #469, opened 2023-12-08
Feat: anonymous generic impl parameters

julio4:
Related to  #453

Explanation + example of anonymous generic implementation parameters in generic chapter.
Enforce usage of the anonymous operator `+` throughout the book's listings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/469)


enitrat (approved):
amazing 
image

enitrat (commented):
for next time don't worry about having the proper numerotation for listings, you can just add a `bis` of just add another number or no number at all if it's descriptive enough

PR #473, opened 2023-12-09
Refactor imports to be consistent with `edition = "2023_10"` 

TAdev0:
while doing this refactoring, i just noticed one thing :  i'm wondering if it is the normal behaviour that `zero()` function is not `inline always` in Zeroable trait, contrary to `is_zero` and `is_non_zero` functions? 
This means that when importing `use core::zeroable::Zeroable;`, calling `is_zero` and `is_non_zero`  is straightforward while `zero()` call needs `Zeroable::zero()`` syntax?

Also, it seems that both syntax compile in the cairo book example situation : 
```use core::zeroable::Zeroable;```
or 
``` use core::zeroable;``` 

Is this normal?

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/473)

### CODE REVIEW

**Comment by enitrat** [comment]
> it is the normal behaviour that zero() function is not inline always in Zeroable trait, contrary to is_zero and is_non_zero functions?

Apparently that's handled automatically by the compiler so there's no need to inline

> is_zeroandis_non_zero is straightforward whilezero()call needsZeroable::zero()`` syntax?

The thing is, you call is_zero and non_zero on a certain type passed with the `self` keyword as parameter. But how can you call `zero()`, which doesn't expect any parameter, otherwise?

> Also, it seems that both syntax compile :
```use core::zeroable::Zeroable;````
or
 use core::zeroable;
 
hm, I don't know!

**Comment by TAdev0** [comment]
yeah i see what you mean for `zero()` function.

One last thing : 
for CH04 , no listing 08 , example was initially : 
```
use clone::Clone;
use array::ArrayTCloneImpl;
fn main() {
    let arr1 = ArrayTrait::<u128>::new();
    let arr2 = arr1.clone();
    ```

i modified it for : 
```
fn main() {
    let arr1 = ArrayTrait::<u128>::new();
    let arr2 = arr1.clone();
    ```
    Indeed, `use clone::Clone` is in the prelude so no need to import. But `ArrayTCloneImpl` is not in the prelude, and i think it is not necessary. I guess Clone trait is enough as `clone` method uses T as parameter, and T can be an array of T ?
    
   I feel like  `impl ArrayTCloneImpl<T, +Clone<T>, +Drop<T>> of Clone<Array<T>>` in array corelib might redondant with Clone generic trait?

**Comment by TAdev0** [comment]
Just to make sure everything is good - In my initial issue , you said: "A restricted version, which requires importing content from the prelude manually with use core::path::to:object (edition 2023_10)" 

Actually this is erroneous because edition 2023_10 actually doesn't need manual imports for traits in the restricted prelude ? But indeed path must be more explicit with "core::path::trait" for all other traits

**Comment by enitrat** [comment]
> Also , I should probably modify some part of text where it's written to import traits that we don't need to import with edition 2023-10. What do you think? And maybe add some infos in appendix E to explain the difference between preludes 2023-01 and 2023-10 ?

Yes to everything

> edition 2023_10 actually doesn't need manual imports for traits in the restricted prelude ? 

Interesting, I didn't know

> But ArrayTCloneImpl is not in the prelude, and i think it is not necessary. I guess Clone trait is enough as clone method uses T as parameter, and T can be an array of T ?

When the chapter was written it was necessary, nice to know it's not anymore!

**Comment by TAdev0** [comment]
In the issue you created with the prelude , you wrote : edit all Scarb.toml files to be edition = "2023_10" based.
For these items, remove all of their explicit imports inside cairo files.

So that was correct , items from prelude can me removed from imports. That's why I asked.

**Comment by TAdev0** [comment]
@enitrat  last commit should be good. I modified all sections in the book which required to be updated. 

There is just this `ArrayTCloneImpl` vs `TCopyClone` (why not `TCopyCloneImpl` in corelib? ) that i dont feel comfortable with. It might need an additional update in Appendix C Derivable traits. Indeed, the book says :
"This means all the fields or values in the type must also implement Clone to derive Clone."

But Clone<T> trait in corelib is :

```
pub trait Clone<T> {
    fn clone(self: @T) -> T;
}

impl TCopyClone<T, +Copy<T>> of Clone<T> {
    fn clone(self: @T) -> T {
        *self
    }
}
```
Doesn't this mean that all fields or values should implement Copy instead of Clone? Or that the overall array needs to implement Copy and not Clone?  I think i need some help to understand key differences between  `TCopyClone` and `ArrayTCloneImpl`.

**Comment by enitrat** [comment]
> impl TCopyClone<T, +Copy<T>> of Clone<T> {
    fn clone(self: @T) -> T {
        *self
    }
}

This is just the implementation of the `Clone` trait for all types that implement copy. 
As `Array` doesn't implement copy, you need to define a specific implementation of Clone, just for the Array type

File: listings/ch04-understanding-ownership/listing_03_03/src/lib.cairo
Comment (enitrat) [inline listings/ch04-understanding-ownership/listing_03_03/src/lib.cairo:1]: could you revert the formatting on this one please

File: listings/ch04-understanding-ownership/listing_03_04/src/lib.cairo
Comment (enitrat) [inline listings/ch04-understanding-ownership/listing_03_04/src/lib.cairo:1]: same

File: listings/ch11-advanced-features/no_listing_04_hash_poseidon/src/lib.cairo [hunk 1]
```diff
@@ -1,7 +1,7 @@
 //ANCHOR: import
-use pedersen::PedersenTrait;
-use poseidon::PoseidonTrait;
-use hash::{HashStateTrait, HashStateExTrait};
+use core::pedersen::PedersenTrait;
```
enitrat, line 2: do pedersen and poseidon need to be imported?

listings/ch11-advanced-features/no_listing_05_advanced_hash/src/lib.cairo
```diff
@@ -1,7 +1,7 @@
 //ANCHOR: import
-use poseidon::PoseidonTrait;
-use poseidon::poseidon_hash_span;
-use hash::{HashStateTrait, HashStateExTrait};
+use core::poseidon::PoseidonTrait;
```
enitrat, line 2: same - still required?

src/appendix-05-common-types-and-traits-and-cairo-prelude.md
```diff
@@ -22,6 +22,11 @@ imports. This prevents repetition and provides a better devX. This is what
 allows you to use `ArrayTrait::append()` or the `Default` trait without bringing
 them explicitly into scope.
 
+You can choose which prelude to use. For example, adding `edition = "2023_10"` in `scarb.toml` configuration file will provide you november 2023 version of the prelude.
```
enitrat, line 25:
```suggestion
You can choose which prelude to use. For example, adding `edition = "2023_10"` in `Scarb.toml` configuration file will load the prelude from October 2023, which is more restricted than the one from January 2023.
```

src/ch02-02-data-types.md
```diff
@@ -119,7 +119,7 @@ Here are some examples of declaring values by putting them between single quotes
 
 ### Type casting
 
-In Cairo, you can convert types scalar types from one type to another by using the `try_into` and `into` methods provided by the `TryInto` and `Into` traits, respectively.
+In Cairo, you can convert scalar types from one type to another by using the `try_into` and `into` methods provided by the `TryInto` and `Into` traits, respectively. These traits are included in the prelude and don't need to be explicitly imported.
```
enitrat, line 122:
```suggestion
In Cairo, you can convert scalar types from one type to another by using the `try_into` and `into` methods provided by the `TryInto` and `Into` traits from the core library.
```

imo: no need to specify unless it requires a specific import

src/ch03-01-arrays.md
```diff
@@ -1,6 +1,6 @@
 ## Arrays
 
-An array is a collection of elements of the same type. You can create and use array methods by importing the `array::ArrayTrait` trait.
+An array is a collection of elements of the same type. You can create and use array methods without explicitly importing the `array::ArrayTrait` trait, as it is included in the prelude.
```
enitrat, line 3:
```suggestion
An array is a collection of elements of the same type. You can create and use array methods by using the `ArrayTrait` trait from the core library.
```

src/ch03-02-dictionaries.md
```diff
@@ -19,7 +19,7 @@ These functions allow us to manipulate dictionaries like in any other language.
 {{#include ../listings/ch03-common-collections/no_listing_07_intro/src/lib.cairo}}
 ```
 
-The first thing we do is import `Felt252DictTrait` which brings to scope all the methods we need to interact with the dictionary. Next, we create a new instance of `Felt252Dict<u64>` by using the `default` method of the `Default` trait and added two individuals, each one with their own balance, using the `insert` method. Finally, we checked the balance of our users with the `get` method.
+`Felt252DictTrait` trait doesn't need to be explicitly imported as it is included in the prelude. Therefore, all the methods we need to interact with the dictionary are by default in the scope. We can create a new instance of `Felt252Dict<u64>` by using the `default` method of the `Default` trait and add two individuals, each one with their own balance, using the `insert` method. Finally, we check the balance of our users with the `get` method.
```
enitrat, line 22:
```suggestion
We can create a new instance of `Felt252Dict<u64>` by using the `default` method of the `Default` trait and add two individuals, each one with their own balance, using the `insert` method. Finally, we check the balance of our users with the `get` method. These methods are defined in the `Felt252DictTrait` trait in the core library.
```

src/ch03-02-dictionaries.md
```diff
@@ -172,7 +172,7 @@ As a finalizing note, these two methods are implemented in a similar way to how
 ### Dictionaries of types not supported natively
 
 One restriction of `Felt252Dict<T>` that we haven't talked about is the trait `Felt252DictValue<T>`.
-This trait defines the `zero_default` method which is the one that gets called when a value does not exist in the dictionary.
+This trait, included in the prelude, defines the `zero_default` method which is the one that gets called when a value does not exist in the dictionary.
```
enitrat, line 175:
```suggestion
This trait defines the `zero_default` method which is the one that gets called when a value does not exist in the dictionary.
```

src/ch06-02-the-match-control-flow-construct.md
```diff
@@ -60,7 +60,7 @@ If we were to call `value_in_cents(Coin::Quarter(UsState::Alaska))`, `coin` woul
 
 ## Matching with Options
 
-In the previous section, we wanted to get the inner `T` value out of the `Some` case when using `Option<T>`; we can also handle `Option<T>` using `match`, as we did with the `Coin` enum! Instead of comparing coins, we’ll compare the variants of `Option<T>`, but the way the `match` expression works remains the same. You can use Options by importing the `option::OptionTrait` trait.
+In the previous section, we wanted to get the inner `T` value out of the `Some` case when using `Option<T>`; we can also handle `Option<T>` using `match`, as we did with the `Coin` enum! Instead of comparing coins, we’ll compare the variants of `Option<T>`, but the way the `match` expression works remains the same. You can use Options without explicitly importing the `option::OptionTrait` trait, as it is included in the prelude.
```
enitrat, line 63:
```suggestion
In the previous section, we wanted to get the inner `T` value out of the `Some` case when using `Option<T>`; we can also handle `Option<T>` using `match`, as we did with the `Coin` enum! Instead of comparing coins, we’ll compare the variants of `Option<T>`, but the way the `match` expression works remains the same.
```

TAdev0:
yes, i understand arrays need a specific implementation as they dont implement copy trait. 

But how is that possible that the following code will compile, without any import? 
```
fn main() {
    let arr1 = ArrayTrait::<u128>::new();
    let arr2 = arr1.clone();
}
```
`ArrayTCloneImpl` is not in the prelude 2023_10 , while clone is. Is it because it is an array of u128 which are copyable, allowing the array to be copyable?

enitrat:
> ArrayTCloneImpl is not in the prelude 2023_10 , while clone is. Is it because it is an array of u128 which are copyable, allowing the array to be copyable?

Since the `array` module is I guess the implementation of the trait is found by the compiler automatically

enitrat (approved):

PR #476, opened 2023-12-11
i18n: Add Indonesian Language

Megumiiiiii:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/476)

PR #477, opened 2023-12-12
[feat]Let wasm-cairo run in each languages' path.

cryptonerdcn:
I'm sorry for forgetting this for a long time. 
Basically before this PR, you can only run the code in the English version of CairoBook -- Because the js files are just under the `root/js`, so it can't be used in the subpath(such as `root/zh-cn/`).

Now all language versions of Cairobook can run the code!
You can test it here: 
https://starknetastro.github.io/cairo-book.github.io/zh-cn/ch01-02-hello-world.html

BTW: Because of Webworker's relative path problem I had to use an ugly way to solve this. https://github.com/parcel-bundler/parcel/issues/5430. I will update it when I find a good way.
@enitrat Can you review this PR? Thanks!

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/477)

### DIFF
--- a/theme/index.hbs
+++ b/theme/index.hbs
@@ -340,6 +340,7 @@
         {{else}}
         <script src="{{ path_to_root }}../js/index.mjs"></script>
         {{/if}}
+        {{/if}}
 
         {{#if search_js}}
         <script src="{{ path_to_root }}elasticlunr.min.js"></script>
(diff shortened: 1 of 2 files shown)


cryptonerdcn:
Maybe this is a piece of good news for translators too.
 @Nadai2010 @glihm  @FlorianLgrs @MathiasTELITSINE 


BTW: @Megumiiiiii After your PR is merged, please let me know. I will add the path for `id` too.

Megumiiiiii:
@cryptonerdcn Merged ✅
https://github.com/cairo-book/cairo-book/pull/476

enitrat (approved):
great 🔥 

also, could you work on 2.4 support?

cryptonerdcn:
@enitrat Of course, I am working on it now. Will release it next week.

PR #483, opened 2023-12-22
Fix typo: replace address with balance

mananbordia:
- In the example where term `address` is being used the Wallet struct doesn't have that field rather there is a `balance` field.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/483)


mananbordia:
@enitrat Can you please tal ?

enitrat (approved):
great, nice catch

PR #487, opened 2023-12-24
Update macros after 2.4.0 release

### LINKED ISSUE #482: feat: update 11.2 Macros
While recently working with `println!` (cairo macro) and `selector!` (starknet-specific macro), i noticed that Macros section in Cairo Book only describes `array!` and `consteval_int!` macros.  
I guess there are some other inline macros but i didn't find where to see all available macros. 

I think Macros section should add `println!` macro short description (which is now used everywhere in the book instead of  print() function of core::debug::PrintTrait). 
`selector!` could also be described (as mentioned in #425), though it is specific to starknet contracts.

If there is a way to find all available macros, i would be happy to describe all (or some) of them in the book as well.

TAdev0:
@enitrat  here is a first draft.

I didn't modify `assert` function call everywhere, especially in the `how to write tests` section where i only mentioned the possible usage of the macro, because there is a whole part about assert function and i feel like it should stay as it is (both function and macro are still usable, I guess the book should keep `assert` at least in one chapter?).

I didn't integrate `writeln!` macro in macros section, as i can't find information to really understand the difference with `write!` macro (anyway, it seems both of them are not used outside debug traits, `format!` is  more convenient).

You also might want to modify the way i reference macros in the macros section, with '' please refer to the .. page'' . I write it 3 times in a row, it feels kind of robotic, i don't know. 

Closes #482 

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/487)


TAdev0:
@enitrat  i can't find a way to make tests pass for this PR. I tried to modify .tool-versions file without hope and indeed it didn't work. I guess i added a mistake somewhere, but each listing file I modified compiles well..

src/ch09-01-how-to-write-tests.md
```diff
@@ -101,6 +101,8 @@ Now that you’ve seen what the test results look like in different scenarios, l
 
 The `assert` function, provided by Cairo, is useful when you want to ensure that some condition in a test evaluates to `true`. We give the `assert` function a first argument that evaluates to a Boolean. If the value is `true`, nothing happens and the test passes. If the value is `false`, the assert function calls `panic()` to cause the test to fail with a message we defined as the second argument of the `assert` function. Using the `assert` function helps us check that our code is functioning in the way we intend.
 
+Since cairo version 2.4.0, `assert!` macro is available, allowing to return an error that doesn't fit into a `felt252`.
```
enitrat, line 104: I would remove this and only mention the assert! macro 🤔  If the assert function is not used in the code it comes from nowhere, therefore I think we can just rework the `assert function` section and adapt it to `assert macro`

src/ch10-01-unrecoverable-errors-with-panic.md
```diff
@@ -31,6 +31,25 @@ Let's consider an example:
 
 Executing this program will yield the same error message as before. In that case, if there is no need for an array and multiple values to be returned within the error, so `panic_with_felt252` is a more succinct alternative.
 
+## `panic!` macro
+
+`panic!` macro has been introduced with cairo version 2.4.0. The previous example returning the error code 2 shows how convenient `panic!` macro is. There is no need to create an array and pass it as argument like `panic` function.
```
enitrat, line 36: Remove all mentions of the compiler version, I think it can only confuse readers - if the feature is available, we simply mention it, otherwise we don't

src/ch10-01-unrecoverable-errors-with-panic.md
```diff
@@ -31,6 +31,25 @@ Let's consider an example:
...
+
+```rust
```
enitrat, line 38: This should be in a `listings` subdirectory, not directly in the markdown

src/ch10-01-unrecoverable-errors-with-panic.md
```diff
@@ -31,6 +31,25 @@ Let's consider an example:
...
+fn main() {
+    if true == true {
+        panic!("2");
+    }
+    println!("This line isn't reached");
+}
+```
+
+Unlike `panic_with_felt252` function, using `panic!` allows the input, which is ultimately the panic error, to be a literal longer than 31 bytes. This is because `panic!` takes a formatted byte array as parameter. For example, the following line of code will successfully compile:
```
enitrat, line 47:
```suggestion
Unlike the `panic_with_felt252` function, using `panic!` allows the input, which is ultimately the panic error, to be a literal longer than 31 bytes. This is because `panic!` takes a string as parameter. For example, the following line of code will successfully compile: 
```

lets keep it simple

src/ch10-01-unrecoverable-errors-with-panic.md
```diff
@@ -85,3 +104,32 @@ Here is an example of its usage:
 We are asserting in main that `my_number` is not zero to ensure that we're not performing a division by 0.
 In this example, `my_number` is zero so the assertion will fail, and the program will panic
 with the string 'number is zero' (as a felt252) and the division will not be reached.
+
+## `assert!` and `assert_eq` macros
+
+Cairo version 2.4.0 introduced `assert!` and `assert_eq` macros. Similarly to `panic!` macro, they both allow to use a panic error string longer than 31 bytes which can be formatted. It is also possible to omit the panic error.
```
enitrat, line 110: same, remove version mention

enitrat, line 110: You need to mention that `assert_eq!` can __only be used in tests__ (and the types compared need to implement `Debug`

src/ch10-01-unrecoverable-errors-with-panic.md
```diff
@@ -85,3 +104,32 @@ Here is an example of its usage:
...
+
+`assert!` macro checks whether a condition holds and panics otherwise: 
+
+```rust
+let var1 = 1;
```
enitrat, line 115: should be a .cairo file under listings/

src/ch10-01-unrecoverable-errors-with-panic.md
```diff
@@ -85,3 +104,32 @@ Here is an example of its usage:
...
+let var2 = 2;
+assert!(var1 != var2);
+assert!(var1 != var2, "values should not be equal");
+assert!(var1 != var2, "{},{} should not be equal", var1, var2);
+```
+
+`assert_eq!` macro checks whether 2 values are equal and panics otherwise:
+
+```rust
```
enitrat, line 124: should be a .cairo file under listings/

src/ch11-02-macros.md
```diff
@@ -1,32 +1,52 @@
...
 
-In some situations, a developer might need to declare a constant that is the result of a computation of integers. To compute a constant expression and use its result at compile time, it is required to use the `consteval_int!` macro.
+### `array!` macro
 
-Here is an example of `consteval_int!`:
+Please refer to the [Arrays](./ch03-01-arrays.md) page.
+
+### `panic!`, `assert!` and `assert_eq!` macros macro
+
+Please refer to the [Unrecoverable Errors with panic](./ch10-01-unrecoverable-errors-with-panic.md) page.
```
enitrat, line 27:
```suggestion
See [Unrecoverable Errors with panic](./ch10-01-unrecoverable-errors-with-panic.md) page.
```

if you think it's too redundant

enitrat (commented):
Great, just a few things to fix. Will check later for the CI error

.tool-versions
```diff
@@ -1 +1 @@
-scarb 2.4.0
+scarb nightly-2023-12-14
```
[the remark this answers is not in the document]
TAdev0: sure.

listings/ch03-common-collections/no_listing_11_dict_of_complex/src/lib.cairo
```diff
@@ -27,14 +23,14 @@ fn main() {
 
     // Search the value and assert it is not null
     let span = match match_nullable(val) {
-        FromNullableResult::Null(()) => panic_with_felt252('No value found'),
+        FromNullableResult::Null(()) => panic!("No value found"),
```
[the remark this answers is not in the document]
TAdev0: i confirm this is correct, it compiles with your correction !

src/ch01-02-hello-world.md
```diff
@@ -177,7 +177,15 @@ screen. There are four important details to notice here.
 
 First, Cairo style is to indent with four spaces, not a tab.
 
-Second, `println!` calls a Cairo macro. If it had called a function instead, it would be entered as `println` (without the `!`). We’ll discuss Cairo macros in more detail in [Chapter Macros](./ch11-02-macros.md). For now, you just need to know that using a `!` means that you’re calling a macro instead of a normal function and that macros don’t always follow the same rules as functions.
+Second, `println!` calls a Cairo macro. If it had called a function instead, it would be entered as `println` (without the `!`). Two macros are available for printing values:
+- `println!` which prints on a new line
```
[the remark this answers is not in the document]
TAdev0: sure

src/ch09-01-how-to-write-tests.md
TAdev0: @enitrat  in ''How to write tests'' section, `assert` function is used in all snippets, i.e. in 9 listings. 
Should i change all of them, to have this section using only assert macro, or should i leave it as it is in my PR, with just a mention to the possible usage of the macro instead ?

TAdev0:
@enitrat  i used ''rust'' in md file for the 3 listings i created. I see sometimes ''rust,noplayground",  which one should i use?

src/ch03-01-arrays.md
```diff
@@ -76,6 +76,23 @@ To determine the number of elements in an array, use the `len()` method. The ret
 
 If you want to check if an array is empty or not, you can use the `is_empty()` method, which returns `true` if the array is empty and `false` otherwise.
 
+### `array!` macro
+
+Sometimes, we need to create arrays with values that are already known at compile time. The basic way of doing that is redundant. You would first declare the array and then append each value one by one. `array!` is a simpler way of doing this task by combining the two steps.
+At compile-time, the compiler will create an array and append all values passed to the `array!` macro sequentially.
```
enitrat, line 82:
```suggestion
At compile-time, the compiler will create an array and append all values passed to the `array!` macro sequentially.
```
Actually this is not totally accurate - At compile time, the compiler will expand the macro to generate the code that appends the items sequentially for us.

src/ch09-01-how-to-write-tests.md
enitrat: Yes, I think we can change all of them to the `assert!` - especially in tests for now.
In the future, we will also rework the test section to use `assert_eq!` for equality testing, but for now it's going to be problematic as there are quite a lot of missing implementations of the Debug trait

src/ch03-01-arrays.md
TAdev0: thank you! will modify this.

TAdev0:
@enitrat  test section updated. Have you figured out why CI always fail?

enitrat:
@TAdev0 two errors
- assert_eq! was used outside of a test
- should_panic test was still expecting a felt252 panic (single quotes) instead of a bytearray (double quotes)

enitrat (approved):
lfg

TAdev0:
@enitrat oh yes sorry for these mistakes. Thanks for the quick fix !

TAdev0:
@enitrat listing 9.9 still uses single quote as well 

´´´
    #[test]
    #[should_panic(expected: ('Guess must be <= 100',))]
    fn greater_than_100() {
        GuessTrait::new(200);
    }
´´´

Will push a small fix to replace with double quote asap

enitrat:
hmm weird, should've been caught by CI 🤔

TAdev0:
As it is just a snippet , maybe it is not compiled in CI ?

enitrat:
probably!

PR #488, opened 2023-12-25
migration from #[external(v0)] to #[abi(embed_v0)] 

### LINKED ISSUE #461: migration from `#[external(v0)]` to `#[abi(embed_v0)]`
Convert all `impl` block marked as `#[external(v0)]` to either `#[abi(embed_v0)]` or `#[abi(per_item)]`, depending on what's the most appropriate one to use (rule of thumb: explicit interface -> `#[abi(embed_v0)]`, generated trait -> `#[abi(per_item)]` )

### LINKED ISSUE #462: `#[abi(embed_v0)]` and `#[abi(per_item)]` annotations
goes with #461 

@faytey

Add explanations about these two annotations, how / when they should be used

Reference: [2.3.0 release notes](https://community.starknet.io/t/cairo-v2-3-0-release-candidate/101137) [Logic in impl blocks](https://community.starknet.io/t/cairo-1-contract-syntax-is-evolving/94794?u=feedthefed)

What needs to be modified: https://book.cairo-lang.org/ch99-01-03-02-contract-functions.html
- Rework the `public functions` section to explain that the public logic of the contract can be grouped inside an impl block annotated with `#[abi(embed_v0)]`
- Add a subsection for the `#[abi(per_item)]` annotation - displaying how a single impl block can share constructor/l1_handle/public/internal functions

TAdev0:
Closes #461 
Closes #462

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/488)


src/ch99-01-03-02-contract-functions.md
```diff
@@ -59,3 +59,41 @@ At this point, you might still be wondering if all of this is really necessary i
 ```rust,noplayground
 {{#include ../listings/ch99-starknet-smart-contracts/listing_99_03_example_contract/src/lib.cairo:stateless_internal}}
 ```
+
+### 4. [abi(per_item)] attribute
+
+Before cairo version 2.3.0, it was not possible to annotate specific functions inside an implementation block. This means that constructor or l1_handler functions could not be included inside a trait implementation, and needed to be annotated individually. Moreover, internal functions needed to be implemented in a different impl than public functions.
```
enitrat, line 65:
```suggestion
Before cairo version 2.3.0, it was not possible to annotate specific functions inside an implementation block. This means that `constructor` or `l1_handler` functions could not be included inside a trait implementation, and needed to be annotated individually. Moreover, internal functions needed to be implemented in a different impl than public functions.
```

src/ch99-01-03-02-contract-functions.md
```diff
@@ -59,3 +59,41 @@ At this point, you might still be wondering if all of this is really necessary i
...
+
+It is now possible to annotate function individually inside an impl, hence allowing different types of entrypoints. `#[abi(per_item)]` attribute is always used with `#[generate_trait]` attribute. In this case, impl and trait name will not be part of the ABI. Note that when using `#[abi(per_item)]` attribute, public functions  need to be annotated with `#[external(v0)]` attribute.
```
enitrat, line 67: > #[abi(per_item)]` attribute is always used with `#[generate_trait]` attribute.

Is that the case? I think you can impl an explicit trait with abi(per_item)

src/ch99-01-03-02-contract-functions.md
```diff
@@ -59,3 +59,41 @@ At this point, you might still be wondering if all of this is really necessary i
...
+Here is a short example:
+
+```rust
+#[abi(per_item)]
```
enitrat, line 71: can you create a Scarb package inside `listings/ch99-...`, add this code, and reference it in the markdown like it's done for other snippets?

enitrat (commented):
Great work

src/ch99-01-03-02-contract-functions.md
TAdev0: mmmh indeed you are gith. I tried two things inside a contract : 
-using `abi(per_item)` won't work when a trait is explicitly defined as an interface. seems logic.
-using `abi(per_item)` works with a trait that is not an interface. I guess it is therefore possible to use it for a trait containing constructor, l1-handler and internal functions. This pattern doesn't seem optimized as it requires 2 traits : one for public functions exposed in the interface and another one for the rest. Using `[generate_trait]` seems better as it is straightforward, allowing a single trait to include everything and generate the correct ABI.

enitrat: agreed - but it's *technically* possible, so we should just outline _why_ it's better.

src/ch99-01-03-02-contract-functions.md
TAdev0: @enitrat  i just did it. But now when i try to git add , i'm having this error : 

```
tristan@Tristans-MacBook-Pro cairo-book.github.io % git add .
error: 'listings/ch99-starknet-smart-contracts/listing_99_14_abi_per_item_attribute/' does not have a commit checked out
fatal: adding files failed
```

any idea how i can fix this ? thank you.

enitrat: > @enitrat  i just did it. But now when i try to git add , i'm having this error : 
> 
> ```
> tristan@Tristans-MacBook-Pro cairo-book.github.io % git add .
> error: 'listings/ch99-starknet-smart-contracts/listing_99_14_abi_per_item_attribute/' does not have a commit checked out
> fatal: adding files failed
> ```
> 
> any idea how i can fix this ? thank you.

Make sure that the newly created folder doesn't have a `.git` folder that could've been automatically created - if it has delete it

What I often end up doing is copy-paste an already existing package and just renaming it's content

TAdev0: @enitrat  indeed, there was one. now it works, thanks !!

TAdev0:
last commit should be good :)

src/ch99-01-03-02-contract-functions.md
```diff
@@ -59,3 +59,17 @@ At this point, you might still be wondering if all of this is really necessary i
 ```rust,noplayground
 {{#include ../listings/ch99-starknet-smart-contracts/listing_99_03_example_contract/src/lib.cairo:stateless_internal}}
 ```
+
+### 4. [abi(per_item)] attribute
+
+Before cairo version 2.3.0, it was not possible to annotate specific functions inside an implementation block. This means that `constructor` or `l1_handler` functions could not be included inside a trait implementation, and needed to be annotated individually. Moreover, internal functions needed to be implemented in a different impl than public functions.
```
enitrat, line 65: Can you remove this sentence?

src/ch99-01-03-02-contract-functions.md
```diff
@@ -59,3 +59,17 @@ At this point, you might still be wondering if all of this is really necessary i
...
+
+It is now possible to annotate function individually inside an impl, hence allowing different types of entrypoints. `#[abi(per_item)]` attribute is often used with `#[generate_trait]` attribute. In this case, impl and trait names will not be part of the ABI. Note that when using `#[abi(per_item)]` attribute, public functions need to be annotated with `#[external(v0)]` attribute.
```
enitrat, line 67:
```suggestion
You can also define the entrypoint type of a function individually inside an impl using the`#[abi(per_item)]` attribute on top of your impl. It is often used with the `#[generate_trait]` attribute, as it allows you to define entrypoints without an explicit interface. In this case, the functions will not be grouped under an impl in the abi. Note that when using `#[abi(per_item)]` attribute, public functions need to be annotated with `#[external(v0)]` attribute - otherwise, they will not be exposed.
```

enitrat (commented):
last thing to fix and all good!

TAdev0:
@enitrat should be good!

enitrat (approved):
:lgtm:

> Reviewed 9 of 10 files at r1, 4 of 5 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/488#-:-NmjoJlv9p-Kk3uHx5EA:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @TAdev0)

PR #489, opened 2023-12-25
Anonymous vs Named Generic Implementation Parameter

remybar:
In the chapter 8.1 "Generic Functions", anonymous generic implementation parameters are described at the end of the chapter (in "Anonymous Generic Implementation Parameter (+ operator)" but they are used in the second and third code snippets, at the beginning of the chapter.

Then, the chapter "Constraints for Generic Types" mentions a `TDrop` implementation but in the code snippet, the drop implementation is anonymous.

To be more consistent with the chapter structure, this PR uses named generic implementation parameters in the first code snippets, until anonymous generic implementation parameter are described.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/489)


enitrat (approved):
Thank you very much!

PR #490, opened 2023-12-25
Issue in wallet field types

remybar:
At the end of the chapter 8.1, 2 `Wallet` are instanciated with the following code:

```
    let w1 = Wallet { balance: true, address: 10 };
    let w2 = Wallet { balance: 32, address: 100 };
```

Then, a paragraph explains that we create 2 instances: one of `Wallet<bool, u128>` and the other of `Wallet<felt252, u8>`.

I think that it's not correct as the compiler will use `felt252` to store `10`, `32` and `100` values by default.

To be sure to have an instance of `Wallet<bool, u128>` and another of `Wallet<felt252, u8>`, we have to use `10_u128` and `100_u8`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/490)

### DIFF
--- a/listings/ch08-generic-types-and-traits/no_listing_14_compiling/src/lib.cairo
+++ b/listings/ch08-generic-types-and-traits/no_listing_14_compiling/src/lib.cairo
@@ -23,7 +23,7 @@ impl WalletMixImpl<T1, +Drop<T1>, U1, +Drop<U1>> of WalletMixTrait<T1, U1> {
 // ANCHOR: main
 fn main() {
     let w1: Wallet<bool, u128> = Wallet { balance: true, address: 10 };
-    let w2 = Wallet { balance: 32, address: 100_u8 };
+    let w2: Wallet<felt252, u8> = Wallet { balance: 32, address: 100 };
 
     let w3 = w1.mixup(w2);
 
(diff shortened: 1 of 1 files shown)


listings/ch08-generic-types-and-traits/no_listing_14_compiling/src/lib.cairo
```diff
@@ -22,8 +22,8 @@ impl WalletMixImpl<T1, +Drop<T1>, U1, +Drop<U1>> of WalletMixTrait<T1, U1> {
 
 // ANCHOR: main
 fn main() {
-    let w1 = Wallet { balance: true, address: 10 };
-    let w2 = Wallet { balance: 32, address: 100 };
+    let w1 = Wallet { balance: true, address: 10_u128 };
```
enitrat, line 25:
```suggestion
    let w1: Wallet<bool, u128> = Wallet { balance: true, address: 10_u128 };
```

listings/ch08-generic-types-and-traits/no_listing_14_compiling/src/lib.cairo
```diff
@@ -22,8 +22,8 @@ impl WalletMixImpl<T1, +Drop<T1>, U1, +Drop<U1>> of WalletMixTrait<T1, U1> {
...
+    let w2 = Wallet { balance: 32, address: 100_u8 };
```
enitrat, line 26:
```suggestion
    let w2: Wallet<bool, u8> = Wallet { balance: 32, address: 100_u8 };
```

enitrat (commented):
Great, nice catch. However I'd rather have explicit type annotation, for clarity, rather than relying on automatic type inference. It makes it more obvious to the reader as well.

Thank you for your thorough reviews!

remybar:
> Great, nice catch. However I'd rather have explicit type annotation, for clarity, rather than relying on automatic type inference. It makes it more obvious to the reader as well.
> 
> Thank you for your thorough reviews!

Yes, I agree 👍

enitrat (approved):

PR #494, opened 2023-12-29
is_err takes a snapshot, not a mutable reference.

VensonLiou:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/494)


enitrat (approved):
indeed !

PR #496, opened 2023-12-30
Example for pedersen hash of two felts

### LINKED ISSUE #495: Missing example for `pedersen` function usage with 2 felts
While doing Security Challenges Factory for Starknet (https://starknet-challenges.vercel.app/) , i've been facing an issue related to `pedersen` hash function. 

In one of their CTFs, i was supposed to calculate the pedersen hash of two `felt252`.  I first used the method described in cairo book : 
```
    let hash = PedersenTrait::new(0).update_with(struct_to_hash).finalize();
```
As my solution didn't work, i searched in corelib and found that pedersen hash of 2 `felt252` is simply calculated as follows : 
```
            let hash = core::pedersen::pedersen(param1, param2);
```
It worked this way. I guess that using the `struct_to_hash` with `update_with` method doesn't return the same hash, even with a struct containing 2 felts, as each element of the struct is hashed one after the other, using hash state accumulators.  This means that in my example, the hash will be the result of `pedersen(pedersen(0, param1), param2)`, instead of `pedersen(param1, param2)`.

I'm not sure this is relevant as `pedersen` seems to be less and less used, with `poseidon` being more STARK friendly.  But I think this is a missing detail that might be important. What do you think @enitrat ?

TAdev0:
Closes #495 

This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/496)


enitrat:
I think you should modify the existing example of Pedersen hashing rather than create a new one. `0` was just a hardcoded value but it can be anything else, so you can reflect this by giving a meaningful variable name

enitrat:
The confusion comes from the fact that Pedersen is instantiated with a base state while poseidon is not.
So if you want to simply hash the struct members together you would need to instantiated the state with the first member of the struct, and then update it with the remaining members, and then finalize. Whereas for poseidon you can simply update the default empty state with the struct to hash.

I'll verify later if there isn't a more elegant way of just hashing a structs fields together - and if there isnt one we should ask for it

TAdev0:
Understood!

listings/ch11-advanced-features/no_listing_04_hash_pedersen/src/lib.cairo
```diff
@@ -14,11 +14,18 @@ struct StructForHash {
...
+fn main() -> (felt252, felt252) {
+    let base_value_to_hash: felt252 = 0;
+    let value_to_update_state: felt252 = 1;
+
     let struct_to_hash = StructForHash { first: 0, second: 1, third: (1, 2), last: false };
 
-    let hash = PedersenTrait::new(0).update_with(struct_to_hash).finalize();
-    hash
+    // hash1 is the result of hashing two felt252
+    let hash1 = PedersenTrait::new(base_value_to_hash).update(value_to_update_state).finalize();
```
enitrat, line 24: Not sure this is very useful

enitrat: I think I would keep the example the same , but add some explanations in the markdown about the base state

listings/ch11-advanced-features/no_listing_04_hash_pedersen/src/lib.cairo
```diff
@@ -14,11 +14,18 @@ struct StructForHash {
...
+    // hash2 is the result of hashing a struct with a base state of 0
+    let hash2 = PedersenTrait::new(0).update_with(struct_to_hash).finalize();
```
enitrat, line 26: actually: keep hash2 (maybe rename to hash1)

and introduce hash2: the result of hashing _only_ the fields of the struct. To do so:
- Use `Serde` to serialize the struct into an array
- Instantiate let state = PedersenTrait::new(struct_serialized[0])
- loop over the serialized struct's fields and call state.update(next_field)

(variable names and correct syntax to adapt)

src/ch11-03-hash.md
```diff
@@ -38,19 +38,20 @@ The first step is to initialize the hash with either `PoseidonTrait::new() -> Ha
...
 
-As our struct derives the trait HashTrait, we can call the function as follow for Poseidon hashing :
+As our struct derives the trait HashTrait, we can call the function as follows for Poseidon hashing :
 
 ```rust
 {{#rustdoc_include ../listings/ch11-advanced-features/no_listing_04_hash_poseidon/src/lib.cairo:main}}
 ```
 
-And as follow for Pedersen hashing :
+And as follows for Pedersen hashing :
```
enitrat, line 47: Edit to something like:
Pedersen is different from Poseidon, as it starts with a base state. This base state must be of `felt252` type, which forces us to either hash the struct with an arbitrary base state using the `update_with` method, or serialize the struct into an array to loop thorugh all of its fields and hash its elements together

TAdev0:
@enitrat  pedersen example updated

enitrat (approved):
lgtm

PR #498, opened 2024-01-03
implementation -> module?

VensonLiou:
A little bit confused here, not sure if it means module originally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/498)


enitrat:
No, implementation was the right word here

PR #500, opened 2024-01-04
Match catch-all operator

remybar:
The chapter about `match` catch-all operator is not so clear as it talks about `enums` and `felt252` while it is only supported for `felt252` in Cairo 2.4.

As `match` will be strongly improved in Cairo 2.5 according to the roadmap, I think it would be better to just mention how it works for `felt252` for the moment, and then deeply update this chapter for Cairo 2.5.

As I'm a Cairo newbie, I may have missed some information about `match` and so this PR may be not fully correct.


enitrat (approved):
This is perfect 💯 thank you for the initiative, feel free to add improvements in any section you feel that should be improved!

enitrat:
can you `cd` into `no_listing_06_match_zero` and run `scarb fmt` ?

remybar:
> This is perfect 💯 thank you for the initiative, feel free to add improvements in any section you feel that should be improved!

ok, thanks ! Yes, if I see something not easy to understand, I will do a PR ;-)
And thanks for the rewards on OnlyDust ❤️

PR #506, opened 2024-01-09
Updating the shell in the examples ch02-05

LeandroCarvajal:
Update the examples mentioned in the issue with the corresponding command to scarb

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/506)


src/ch02-05-control-flow.md
```diff
@@ -30,15 +31,18 @@ Let’s try changing the value of `number` to a value that makes the condition `
...
+Run completed successfully, returning []
+Remaining gas: 19780390
 ```
 
 It’s also worth noting that the condition in this code must be a bool. If the condition isn’t a bool, we’ll get an error.
 
 ```shell
-$ cairo-run main.cairo
-thread 'main' panicked at 'Failed to specialize: `enum_match<felt252>`. Error: Could not specialize libfunc `enum_match` with generic_args: [Type(ConcreteTypeId { id: 1, debug_name: None })]. Error: Provided generic argument is unsupported.', crates/cairo-lang-sierra-generator/src/utils.rs:256:9
+$ scarb cairo-run --available-gas=20000000
```
enitrat, line 43: what did you run exactly for this?

src/ch02-05-control-flow.md
```diff
@@ -19,8 +19,9 @@ Optionally, we can also include an `else` expression, which we chose to do here,
 Try running this code; you should see the following output:
 
 ```shell
-$ cairo-run main.cairo
-[DEBUG]	condition was false
+$ scarb cairo-run --available-gas=20000000
+Run completed successfully, returning []
```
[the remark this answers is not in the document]
LeandroCarvajal: using the same example of the book, I agree that it should be reflected in the false

src/ch02-05-control-flow.md
LeandroCarvajal: Use the panicked example from the book
```
{{#include ../listings/ch10-error-handling/no_listing_01_panic/src/lib.cairo}}
```
What would be more convenient for this example?

enitrat: This is not really relevant. What was described here was that you _had_ to use a `bool` as a condition, otherwise your program would not compile. At the time it was written, the error was the one written in the book when you tried to run the example. But now, I think it just won't even let you compile. Can you revert this specific change, i'll open an issue specifically for this

enitrat (approved):

PR #509, opened 2024-01-12
Internal functions in a smart contract

remybar:
By reading the explanation about how to write internal functions in a Cairo smart contract, it's not clear if these internal functions must be in an implementation block to access to `ContractState` or not, as the free function `get_owner_storage_address` uses the ContractState and is not in an implementation block.

After discussion on the Starknet discord, it appears that internal functions just have to be outside the implementation block tagged with `#[abi(embed_v0)]`. That means, they can be grouped in an implementation block if it makes the code more readable/easier to use, or just be added as free functions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/509)


enitrat (approved):
awesome 💯

PR #517, opened 2024-01-15
update: ch99-01-04-01-voting-contract.md

hel-kame:
added notes on the versions to be used for the example to work properly, updated all redirection links.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/517)


enitrat (approved):
Thanks for the careful update 🖖

PR #521, opened 2024-01-17
fix: adjust messaging content after feedbacks from users

glihm:
@enitrat I was wrong about messaging on the wording. We had a fix on Katana and Devnet which corrected how messaging fields were extracted from blockifier.

This PR aims at giving a more accurate information to the reader.

Also from some feedback, users are trying to directly call the starknet core contract. I added some rewording to avoid this too.

Sorry again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/521)

### DIFF
--- a/src/ch99-04-00-L1-L2-messaging.md
+++ b/src/ch99-04-00-L1-L2-messaging.md
@@ -158,6 +158,8 @@ function consumeMessageFelt(
 
 As you can see, in this context we don't have to verify which contract from L2 is sending the message (as we do on the L2 to verify which contract from L1 is sending the message). But we are actually using the `consumeMessageFromL2` of the starknet core contract to validate the inputs (the contract address on L2 and the payload) to ensure we are only consuming valid messages.
 
+As an additional note on `consumeMessageFromL2`, this function of the starknet core contract is expected to be called from a solidity contract, and not directly on the starknet core contract. The reason of that is because the starknet core contract is using `msg.sender` to actually compute the hash of the message. And this `msg.sender` must correspond to the `to_address` field that is given to the function `send_message_to_l1_syscall` that is called on Starknet.  
+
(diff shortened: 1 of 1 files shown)


enitrat:
You cannot consume a message without being in a solidity contract?

glihm:
> You cannot consume a message without being in a solidity contract?

Usually not because you've the `to_address` field that you've to give at the syscall, which is a solidity contract address. So unless you're sending a message to starknet core contract (which I don't this there are much use cases here), you've to give a valid solidity contract address.

Also, the starknet core contract code uses `msg.sender` to compute the hash.
```js
    function consumeMessageFromL2(uint256 fromAddress, uint256[] calldata payload)
        external
        override
        returns (bytes32)
    {
        bytes32 msgHash = keccak256(
            // *** Here, the starknet core contract uses the caller contract address ***
            abi.encodePacked(fromAddress, uint256(uint160(msg.sender)), payload.length, payload)
        );

        require(l2ToL1Messages()[msgHash] > 0, "INVALID_MESSAGE_TO_CONSUME");
        emit ConsumedMessageToL1(fromAddress, msg.sender, payload);
        l2ToL1Messages()[msgHash] -= 1;
        return msgHash;
    }
```

enitrat:
> So unless you're sending a message to starknet core contract (which I don't this there are much use cases here), you've to give a valid solidity contract address.

can you perhaps make it more obvious?

glihm:
> > So unless you're sending a message to starknet core contract (which I don't this there are much use cases here), you've to give a valid solidity contract address.
> 
> can you perhaps make it more obvious in the text?

Totally, that's a good point. Sorry for the delay, just added a little note on that.
Thank you @enitrat.

enitrat (approved):

Issue #523, opened 2024-01-22
feat: 2.5.0


TAdev0:
- [x] match improvements, exhaustive  arms not required anymore / matching over integers available
- [x] while loops now available
- [x] storing ByteArray (ie strings) in contracts
- [x] pub and pub(crate) keywords
- [x] #652

PR #524, opened 2024-01-23
feat: While loops

julio4:
#523 While loops

Scarb 2.4.0 -> Scarb 2.5.0 (For cairo 2.5.0)
Modified ch02/05-control-flow/Loops to introduce `while` loops construct.
Moved most of the loops explanation of `loop` to `while` to introduce `while` loops first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/524)


enitrat (changes requested):
Can you modify the order in which things are presented (and how they're presented) to match the rust book? 
https://doc.rust-lang.org/book/ch03-05-control-flow.html

julio4:
> Can you modify the order in which things are presented (and how they're presented) to match the rust book? https://doc.rust-lang.org/book/ch03-05-control-flow.html

Didn't noticed it was from rust book. I'll update PR

enitrat:
yeah, it was a convenient, proof-tested and fast way of starting the cairo book given the similarities between the two languages

julio4:
> yeah, it was a convenient, proof-tested and fast way of starting the cairo book given the similarities between the two languages

I reverted most of my changes and keep the same order of sections from the rustbook. It's indeed more simple and straightforward!

src/ch02-05-control-flow.md
```diff
@@ -119,22 +119,17 @@ The `number` variable will be bound to a value based on the outcome of the `if`
 
 ### Repetition with Loops
 
-It’s often useful to execute a block of code more than once. For this task, Cairo provides a simple loop syntax, which will run through the code inside the loop body to the end and then start immediately back at the beginning. To experiment with loops, let’s create a new project called loops.
```
enitrat, line 122: > To experiment with loops, let’s create a new project called *loops*.

Bring this back

src/ch02-05-control-flow.md
```diff
@@ -119,22 +119,17 @@ The `number` variable will be bound to a value based on the outcome of the `if`
...
 
-Cairo only has one kind of loop for now: `loop`.
+Cairo has two kinds of loops: `loop` and `while`.
 
 #### Repeating Code with `loop`
 
-The `loop` keyword tells Cairo to execute a block of code over and over again
-forever or until you explicitly tell it to stop.
-
-As an example, change the _src/lib.cairo_ file in your _loops_ directory to look
```
enitrat, line 131: >As an example, change the _src/lib.cairo_ file in your _loops_ directory to look
like this:
<span class="filename">Filename: src/lib.cairo</span>

bring this back

src/ch02-05-control-flow.md
```diff
@@ -119,22 +119,17 @@ The `number` variable will be bound to a value based on the outcome of the `if`
...
-like this:
-
-<span class="filename">Filename: src/lib.cairo</span>
+The `loop` keyword tells Cairo to execute a block of code over and over again forever or until you explicitly tell it to stop.
 
 ```rust
-{{#include ../listings/ch02-common-programming-concepts/no_listing_27_loop/src/lib.cairo}}
+{{#include ../listings/ch02-common-programming-concepts/no_listing_27_infinite_loop/src/lib.cairo}}
```
enitrat, line 132: The code snippet is:
```rust
fn main() {
    let mut i: usize = 0;
    loop {
        if i > 10 {
            break;
        }
        println!("i = {}", i);
    }
}
```

Can we change it to 

```rust
fn main() {
    loop {
        println!("again!");
    }
}
``` 

Without an explicit condition, as an infinite loop should have?

src/ch02-05-control-flow.md
```diff
@@ -119,22 +119,17 @@ The `number` variable will be bound to a value based on the outcome of the `if`
...
 ```
 
 When we run this program, we’ll see `i = 0` printed over and over continuously
```
enitrat, line 135: and thus adapt the sentence below to

> When we run this program, we’ll see `again!` printed over and over continuously until either the program runs out of gas we stop the program manually. Most terminals support the keyboard shortcut ctrl-c to interrupt a program that is stuck in a continual loop. Give it a try:

- adapt code snippet here - 

Then, before introducing the concept of gas, add this line.

> Now, try running the same program again, but this time with the `--available-gas` flag set to `100000000000000000`

enitrat (commented):
lgtm overall

enitrat (approved):

PR #537, opened 2024-01-29
maintain: Chapter-2-2

### LINKED ISSUE #534: maintain: Chapter 2-2
- [x] introduce signed integer i8 -> i128 which are missing. Use the rust book as reference.
- [x] felt section: "an overflow occurs" => an overflow (or underflow) occurs
- [x] integer section: "such as overflow checks" => "such as overflow and underflow checks"
- [x] introduce `_` as a visual separator 
- [x] Tuple examples including type declaration should use the tuple variable instead of the destructured result : `let tup: (felt252, felt252)`
- [x] precise that bool cannot be declared using 0 or 1 for false and true, only the literals false and true

TAdev0:
Resolve #534

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/537)


TAdev0:
@enitrat  tell me what you think of the ''Compound'' section, with references for arrays and dict , to have a clear chapter with : 
- 1 -> Scalar types
- 2 -> Compound types

enitrat:
I didn't really know the difference, so after a quick search:

The difference between a compound type and a collection in programming lies in their structure and intended use:

1. **Compound Type**: A compound type, also known as a composite type, is a data type that is composed of two or more primitive or other compound types. These types are structured and have a fixed schema. Examples include:

   - **Structures (Structs)**: Combine different types of elements (e.g., integers, strings) into a single entity. Each element of a struct is called a field.
   - **Arrays**: Hold a fixed number of elements of the same type and are indexed.
   - **Tuples**: Contain a fixed set of elements that may be of different types. Each element in a tuple can be of a different type, and the tuple has a fixed size.
   - **Classes** (in object-oriented programming): Define objects that encapsulate data and behavior together. A class is a blueprint for creating instances (objects), each with its own state and behavior.

   Compound types are typically used to represent more complex data structures and are integral to the data modeling aspect of programming.

2. **Collection**: A collection, on the other hand, is a more general term that refers to any grouping of multiple items, often referred to as elements or members. Collections can grow or shrink dynamically and are primarily used to store, retrieve, and manipulate data. Common types of collections include:

   - **Lists**: Ordered collections that can contain duplicate elements and whose size can change dynamically.
   - **Sets**: Unordered collections that do not contain duplicate elements.
   - **Maps or Dictionaries**: Collections of key-value pairs, where each key is unique and is used to retrieve its corresponding value.
   - **Queues**: Collections that operate on a first-in, first-out (FIFO) principle.
   - **Stacks**: Collections that operate on a last-in, first-out (LIFO) principle.

   Collections are more about managing groups of data and providing mechanisms for dynamically handling these groups (like adding, removing, and searching elements).

In summary, compound types are structured and more rigid in their composition, combining multiple data types into one, while collections are more about handling groups of data, often with flexible sizes and dynamic capabilities. Compound types are a way to structure and relate data, whereas collections are a way to manage and manipulate groups of data.



From this definition: 
- A span is a sort of compound type
- A tuple is a compound type

- An array is a collection (a mutable data structure, non-fixed size of elements)
- A dictionnary is a collection


I think there is a plan to support later on __const arrays__ which would be basically arrays of constant size known and declared at compile time - which would fit in the compound type section. 

So, TLDR: let's keep only tuples there for now.

TAdev0:
very clear, thanks for this.
will modify and go back to tuple only then

TAdev0:
@enitrat  updated

enitrat (approved):

PR #539, opened 2024-01-30
fix: Code example inconsistencies and Pedersen hash not updating

b-j-roberts:
Noticed the following typos/errors in the Cairo book: 

In ch09, panics testing :
1. Code example for `should_panic` with expected `ByteArray` doesn't contain updated `GuessImpl` code in listing 9-8.
2. The next code example, which is supposed to introduce a bug to demonstrate errors, does not introduce said bug.

In ch11, on hashes : 
1. Imports refer to `Pederson` hashes but use `Poseidon` hashes in code.
2. `Pederson` hashing example does not update `state` on `hash2` calculation, so no accumulation is done.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/539)


TAdev0:
@b-j-roberts  thank you for the PR.

- CH 09-1 : Nice catch for the non updated GuessImpl
- CH09 - 2 and CH10 - 1 : those errors were spotted and about to be fixed
- CH10-2 : the following block actually creates a new state for hash2, updates it multiple times, before finalizing in the last line, so it should be ok: 

```
    let mut serialized_struct: Array<felt252> = ArrayTrait::new();
    Serde::serialize(@struct_to_hash, ref serialized_struct);
    let first_element = serialized_struct.pop_front().unwrap();
    let mut state = PedersenTrait::new(first_element);
    loop {
        match serialized_struct.pop_front() {
            Option::Some(value) => state.update(value),
            Option::None => { break; }
        };
    };
    
        let hash2 = state.finalize();
```

b-j-roberts:
> * CH10-2 : the following block actually creates a new state for hash2, updates it multiple times, before finalizing in the last line, so it should be ok:
> 
> ```
>     let mut serialized_struct: Array<felt252> = ArrayTrait::new();
>     Serde::serialize(@struct_to_hash, ref serialized_struct);
>     let first_element = serialized_struct.pop_front().unwrap();
>     let mut state = PedersenTrait::new(first_element);
>     loop {
>         match serialized_struct.pop_front() {
>             Option::Some(value) => state.update(value),
>             Option::None => { break; }
>         };
>     };
>     
>         let hash2 = state.finalize();
> ```

When I run the code like this and do `println!("hash2: {}", hash2);` after finalizing, I get `hash2: 3` as output.
This is from using this struct : `let struct_to_hash = StructForHash { first: 3, second: 1, third: (1, 2), last: false };`

From what I can tell the new `state` doesn't accumulate values from `update` calls unless reassigning like `state = state.update(value);`.
Ie I think `state.update(val)` doesn't mutate `state`.

b-j-roberts:
I think it can be seen [here](https://github.com/starkware-libs/cairo/blob/v2.5.0/corelib/src/pedersen.cairo) that the `HashState` is not passed by reference.

enitrat:
> Ie I think state.update(val) doesn't mutate state.

Indeed, the purpose is that each update returns the new state instate of mutating the old one, so that you can chain state updates like
```rust
let mut hash = state.update(1).update(2).update(3); 
```
in a single line

TAdev0:
@b-j-roberts i didn't take enough time to check your commit. Indeed, `state` variable was not updated in the loop, my bad. Great job !
You need to rebase  for the PR to be merged :)

enitrat (dismissed):

PR #542, opened 2024-01-30
maintain: Chapter 2-3

### LINKED ISSUE #541: maintain: Chapter 2-3
- [x] “Calling a function is an expression” could be completed, explaining that if there is no return value for the function, the unit type is the expression returned
- [x] remove ‘’Compiling this code produces an error, as follows:” as it is redundant
- [x] update shell result and println! content
- [x] remove -available-gas flag and remaining gas

TAdev0:
Resolves #541

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/542)


src/ch02-03-functions.md
```diff
@@ -175,9 +174,13 @@ Expressions evaluate to a value and make up most of the rest of the code that
 you’ll write in Cairo. Consider a math operation, such as `5 + 6`, which is an
 expression that evaluates to the value `11`. Expressions can be part of
 statements: in Listing 2-1, the `6` in the statement `let y = 6;` is an
-expression that evaluates to the value `6`. Calling a function is an
-expression. A new scope block created with
-curly brackets is an expression, for example:
+expression that evaluates to the value `6`.
+
+Calling a function is an expression. Indeed, it can either return:
```
enitrat, line 179: Everything in Cairo is an expression, and expressions that return "nothing" actually return (). I think that would be clearer if we don't have the "either return", but more something like 

"Calling a function is an expression since it always evaluates to a value: the function's explicit return value, if specified, or the 'unit' type () otherwise."

TAdev0: i didnt feel comfortable with how i wrote that
this is definitely better

PR #557, opened 2024-02-03
Maintain Chapter 3-2

### LINKED ISSUE #556: Maintain Chapter 3-2 Dictionaries
- [x] remove residual unnecessary imports 
- [x] value should be key in " A key field that identifies the value for this key-value pair of the dictionary."
- [x] nopanic is used twice, but never defined, add a short reference
- [x] `add_user` function implementation is wrong: `users_amount` is incremented whenever it is called, while it shouldn’t be incremented if we call `insert` on the same name key

TAdev0:
Resolves #556

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/557)


TAdev0:
@enitrat  should be good
- listing with compile error updated
- Dict as struct members moved to Custom data structures

just one thing: i ran the script to reorder the listings as i moved a section with listings at the end of Custom Data Structures which also contains listings
but nothing happened
The last listing used is still the number 12, with 13 and 14 being used before it in Custom Data Structures section

listings/ch03-common-collections/no_listing_07_intro/src/lib.cairo
```diff
@@ -5,8 +5,8 @@ fn main() {
     balances.insert('Maria', 200);
 
     let alex_balance = balances.get('Alex');
-    assert(alex_balance == 100, 'Balance is not 100');
+    assert!(alex_balance == 100, "Balance is not 100");
```
enitrat, line 8: assert_eq when checking for equality

src/ch03-03-custom-data-structures.md
```diff
@@ -143,6 +143,57 @@ can use in your code, can be found in the community-maintained
 [Alexandria](https://github.com/keep-starknet-strange/alexandria/tree/main/src/data_structures)
 library, in the "data_structures" crate.
 
+## Dictionaries as Struct Members
```
enitrat, line 146: I think this should be the first example of this chapter.

enitrat:
>  i ran the script to reorder the listings as i moved a section with listings at the end of Custom Data Structures which also contains listings

Yeah because they're not listings, they're not being caption with `Listing X-Y` - and the folder name is `no_listing_xxx`. In general we want to caption and use numbered listings for __important__ code snippets we make a reference to, and otherwise no captions, and no_listing.

For example: You introduce a new concept, how to create a rust program that prints Hello World -> Listing
Then, you illustrate each part of the program (the function signature, the body, the brackets) in isolated snippets -> Not a listing

It has not been respected everywhere

TAdev0:
oh great now i get it thanks!

listings/ch03-common-collections/no_listing_07_intro/src/lib.cairo
TAdev0: shouldn't we use assert_eq only in tests?

TAdev0: i just tried and it doesnt compile outside of tests

enitrat: sorry! my bad. read too fast

enitrat (approved):

PR #583, opened 2024-02-09
Maintain Chapter 9 and 10 : Error Handling and Testing Cairo Programs

### LINKED ISSUE #566: Maintain Chapter 9 Testing cairo programs
- [x]  Error Handling (10) should be moved just before Testing cairo programs (9), as `assert!` is widely used in testing chapter without being introduced 
- [x] 9-1: ‘’For now … on the function’’ should be removed as we explain the first line just after
- [ ] 9-1: #[cfg(test)] attribute never described in this section, we should do it , at least a small reference to the next section (testing organization) 
- [x] 9-1: In should_panic section, the last snippet of code is supposed to introduce a bug via swapping bodies but this is not done
- [ ] 9-1: print() call in benchmarking the gas usage example should be replaced by println!
- [ ] 9-2: Introduction phrase should be modified as public/private functions are now possible with pub attribute
- [ ] 9-2: `cfg(tests)` should be `cfg(test)`

### LINKED ISSUE #567: Maintain Chapter 10 Error handling
- [x] 10-1: just add that #[cfg(test)] will be explained in the next section (Tests), after moving this section before Tests
- [x] 10-2: blank space to delete before ‘’it is always easier to understand with examples”
- [x] Shortly introduce the should_panic attribute (used once here) classically used in tests as we move tests section after error handling
- [x] typos

TAdev0:
Resolves #566 
Resolves #567

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/583)


TAdev0:
@enitrat  no need to check before tomorrow.

Basically i am moving Error handling before Testing. so i change all names manually for listings and everything, so that everything will be really clear

Thats why i grouped these 2 issues in the same PR

TAdev0:
will wait for Maintain Chapter 8 to be merged to continue, i get weird things because of all deleted and re-created files

TAdev0:
@enitrat  i think i'm finished with it.

About testing organization, we explain this organization : 

```
adder
├── Scarb.toml
├── src
 │   ├── lib.cairo
 │   ├── tests
 │    │   └── integration_test.cairo
 │   └── tests.cairo
 ```
 
 but this is also possible :
 
 ```
 adder
├── Scarb.toml
├── src
 │   ├── lib.cairo
├── tests
 │    │   └─integration_test.cairo
 │   └── lib.cairo
````

Should we add it ?

enitrat:
No, let's keep the tests inside `src`

enitrat:
@TAdev0 can you check the conflict

enitrat:
also ideally: don't move files AND make changes inside them, as this makes it impossible to review what the new additions are in the git diff. Prefer changing and moving in two different PRs

enitrat (approved):

PR #588, opened 2024-02-12
Fix typos and display errors

TAdev0:
Fixes many small typos, as well as display issues (TAG visible, alignment , etc)

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/588)


enitrat (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/588)* status: 0 of 41 files reviewed, 1 unresolved discussion (waiting on @TAdev0)

___
*[`listings/ch02-common-programming-concepts/no_listing_01_variables_are_immutable/src/lib.cairo` line 2 at r2](https://reviewable.io/reviews/cairo-book/cairo-book/588#-NqWWB_96UoZhdARJTMM:-NqWWB_96UoZhdARJTMN:b-axqxik) ([raw file](https://github.com/cairo-book/cairo-book/blob/87a0bb876b5829c789a8a2e1af78b7105395d775/listings/ch02-common-programming-concepts/no_listing_01_variables_are_immutable/src/lib.cairo#L2)):*
> ```Cairo
> //TAG: does_not_compile
> // ANCHOR: main
> ```


The tag is not shown in the image - I don't think it's necessary then. If you think the extra line at the top is an issue I think we can probably fix this in the preprocessor

TAdev0:
Im not sure to understand . What do you mean ?

enitrat:
The //TAG line is correctly eliminated by the preprocessor, thus we don't need to wrap the entire file between //ANCHOR lines when refering to them in the corpus of the book.

If the problem is that there's an empty line at the beginning of code snippets, we can fix this by modifying the preprocessor to skip empty lines following a TAG line

TAdev0:
Mmh if you check the book , you will see all these TAGs. They don't seem to be automatically removed.

Issue #454, opened 2023-11-25
feat: 2.4.0 


enitrat:
## Cairo

- [x] refactor `#[should_panic(expected: 'error')]` to use double quotes instead of single (less confusing)
- [x] match on felt252 continuous from 0 and on values
- [x] Felt252Dict<bool> fixed
- [ ] Dicts of non-copy nullable type possible: access items using `entry` rather than `get` #421
- [x] Prelude editions - Add info about this in the first chapter where we dissect `Scarb.toml`
    Currently 2 editions exist: 2023_01 (default) and 2023_10.
- [x] Remove `#[available_gas(X)]` attributes on test functions (simplify)
- [x] `assert_eq!` for testing - in tests only
- [x] `assert!` and `panic!` macros to panic with a format string instead of a single felt252 non-dynamic value
- [x] `print!` and `println!` macros - replace all existing uses of `val.print()` by `println!("{}", val)`
- [x] format string (bytearray) creation with `format!` macro
- [x] `write!` and `writeln!` - TBD
- [x] Display and Debug traits

enitrat:
Hi @faytey can you link me some of your previous work?

faytey:
okay do you mean personal projects or contributions i've made because all is on my repo github.com/faytey

TAdev0:
@enitrat  i'm available to work on some of 2.4 features if needed :)

PR #621, opened 2024-02-21
add a printing page

remybar:
Related to https://github.com/cairo-book/cairo-book/issues/454.

With Cairo 2.4.0 appear `Display` trait and some macros like `println!`, so I suggest adding a page dedicated to how to print some data to the console. This page describes how to print standard types with `println!` and `print`, how to format with `format!` and how to write custom data types using `Display` trait and `write!`/`writeln!`. I think it would be easier to find for a new Cairo developer than in the macro page.

I'm not sure about where to add this Printing page because it is not really an advanced feature, but the reader should know what is a trait before seeing how to implement the `Display` trait 🤔 

Then, I think that the "Macros" page should just be a list of Cairo macros in the "Appendix" section. Then, these macros should directly appear in the pages where they need to be used (for example, `assert!` in the page about tests, `array!` to the array page, ...).

I didn't move the "Macros" page yet, I prefer having your point of view :-)

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/621)


enitrat (dismissed):
great initiative! I'm not sure where to store it, so let's keep it here for now

enitrat:
needs a rebase and a conflictsolve

enitrat (approved):

PR #625, opened 2024-02-22
chore(component): add precision on `get_dep_component` macro

LucasLvy:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/625)

### DIFF
--- a/src/ch99-01-05-02-component-dependencies.md
+++ b/src/ch99-01-05-02-component-dependencies.md
@@ -51,8 +51,9 @@ mutate it to transfer the ownership.
 ## `get_dep_component!` macro
 
 The `get_dep_component!` and `get_dep_component_mut!` are two macros which are
-used to get a component give a contract state that has it.
+used to get a component given a contract state that has it. It goes up to the
+contract state and down again to the requested component state
 This macro takes 2 arguments:
 
-- a contract state (`@self` for `get_dep_component!` and `ref self` for `get_dep_component!`)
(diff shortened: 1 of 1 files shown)


src/ch99-01-05-02-component-dependencies.md
```diff
@@ -51,8 +51,9 @@ mutate it to transfer the ownership.
 ## `get_dep_component!` macro
 
 The `get_dep_component!` and `get_dep_component_mut!` are two macros which are
-used to get a component give a contract state that has it.
+used to get a component given a contract state that has it. It goes up to the
+contract state and down again to the requested component state
```
TAdev0, line 55: It seems there is a dot missing at the end. 

Shouldn't it be :  ''two macros which are used to get a component given a **component** state that has it ? 
I mean, these macros are used by components, to inherit other component's storage, events and functions, right?  i'm not sure it is clear as currently explained

enitrat: > I mean, these macros are used by components, to inherit other component's storage

yes but all of it is done through the main contract that embeds these compnents. So what happens when you get compoonent C from B is that you go through A

B -> Contract A -> C

LucasLvy: no these macros allow you to get another component from a component and then call the fonction of the other component

enitrat (approved):

PR #627, opened 2024-02-22
chore: improve component dependencies section

enitrat:
@TAdev0 please TAL

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/627)

### DIFF
--- a/src/ch99-01-05-02-component-dependencies.md
+++ b/src/ch99-01-05-02-component-dependencies.md
@@ -1,11 +1,12 @@
 # Component dependencies
 
-A component can use another component as a dependency. It's for example possible to create
-the `OwnableCounter` component that will use the `Ownable` component as a dependency.
+Working with components becomes more complex when we try to use one component inside another. As mentioned earlier, a component can only be embedded within a contract, meaning that it's not possible to embed a component within another component. However, this doesn't mean that we can't use one component inside another. In this section, we will see how to use a component as a dependency of another component.
 
-The component structure is very similar to a regular component.
+Consider a component called `OwnableCounter` whose purpose is to create a counter that can only be incremented by its owner. This component can be embedded in any contract, so that any contract that uses it will have a counter that can only be incremented by its owner.
 
-Here is the full implementation that we'll break down right after
+The first way to implement this is to create a single component that contains both counter and ownership features from within a single component. However, this approach is not recommended: our goal is to minimize the amount of code duplication and take advantage of component reusability. Instead, we can create a new component that _depends_ on the `Ownable' component for the ownership features, and internally defines the logic for the counter.
+
+Here's the complete implementation, which we'll break down right after:
 
 ```rust
 {{#include ../listings/ch99-starknet-smart-contracts/components/listing_01_component_dep/src/counter.cairo:full}}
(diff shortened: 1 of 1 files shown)


TAdev0 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/627)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @enitrat)

___
*[`src/ch99-01-05-02-component-dependencies.md` line 7 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/627#-NrGpTRP9KGgUIeKbZVg:-NrGpTRP9KGgUIeKbZVh:b-kxzt5p) ([raw file](https://github.com/cairo-book/cairo-book/blob/bd76d532758d49350ea4ada85fe1ce0a0876777a/src/ch99-01-05-02-component-dependencies.md#L7)):*
> ```Markdown
> Consider a component called `OwnableCounter` whose purpose is to create a counter that can only be incremented by its owner. This component can be embedded in any contract, so that any contract that uses it will have a counter that can only be incremented by its owner.
> 
> The first way to implement this is to create a single component that contains both counter and ownership features from within a single component. However, this approach is not recommended: our goal is to minimize the amount of code duplication and take advantage of component reusability. Instead, we can create a new component that _depends_ on the `Ownable' component for the ownership features, and internally defines the logic for the counter.
> ```

i think you wrote `Ownable' instead of `Ownable` , wrong parenthesis
___
*[`src/ch99-01-05-02-component-dependencies.md` line 23 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/627#-NrGswDC7B_ThNJtIvIg:-NrGswDC7B_ThNJtIvIh:b-p5s0yz) ([raw file](https://github.com/cairo-book/cairo-book/blob/bd76d532758d49350ea4ada85fe1ce0a0876777a/src/ch99-01-05-02-component-dependencies.md#L23)):*
> ```Markdown
> ```
> 
> In [Chapter 8](./ch08-02-traits-in-cairo.md), we introduced trait bounds, which are used to specify that a generic type must implement a certain trait. In the same way, we can specify that a component depends on another component by restricting the `impl` block to be available only for contracts that contain the required component. 
> ```

actually there is strictly no example of trait bounds in 08-02, it is just mentioned at the beginnning, but trait bounds are introduced in 08-01

so (./ch08-01-generic-data-types.md)
___
*[`src/ch99-01-05-02-component-dependencies.md` line 35 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/627#-NrGu9RT5zoeeCzWst3C:-NrGu9RT5zoeeCzWst3D:b-lbvnep) ([raw file](https://github.com/cairo-book/cairo-book/blob/bd76d532758d49350ea4ada85fe1ce0a0876777a/src/ch99-01-05-02-component-dependencies.md#L35)):*
> ```Markdown
> Now that we have made our `impl` depend on the `Ownable` component, we can access its functions, storage, and events within the implementation block. To bring the `Ownable` component into scope, we have two choices, depending on whether we intend to mutate the state of the `Ownable` component or not.
> If we want to access the state of the `Ownable` component without mutating it, we use the `get_dep_component!` macro. If we want to mutate the state of the `Ownable` component (for example, change the current owner), we use the `get_dep_component_mut!` macro.
> Both macros take two arguments: the first is `self`, representing the state of the component using the dependency, and the second is the component to access.
> ```

mmh  its self (ref self) for get_dep_component_mut! but @self for get_dep_component!

one macro passes by reference and the other one by snapshot

LucasLvy (approved):
Very well written consider written an actual book in addition to selling a printed version of this one 💋

Issue #568, opened 2024-02-03
Maintain Chapter 11 Advanced features


TAdev0:
- [x] should be renamed Advance Cairo (features)
- [x] Move custom data structures in this section
- [x] 11-1: maybe add precisions that there is no need to use the #[generate_trait] attribute here as it is just one implementation of Add<T> trait method of corelib
- [x] 11-2: allows -> allow
- [x] 11-3: update_with is a part of the extension trait for hash state accumulators called HashStateExTrait, this should be shortly mentioned with other traits, especially because update_with is used multiple times in the section
- [x] 11-3: A few errors with imports (pedersen instead of poseidon at some place)
- [x] 11-3: Serde and copy attributes are not necessary for StructToHash
- [x]  11-3: Serialization could be shortly introduced as it is used for the poseidon struct example 
- [x] 11-3: Advanced hashing : example of hashing a **struct** (instead of function) that contains a span<felt252>
- [x] 11-3: following trait**s** and function

kwkr:
Hello @enitrat and @TAdev0 ! I allowed myself to fix some of the points that I could find from the list.
The points I wasn't able to fix/find:

11-1: maybe add precisions that there is no need to use the #[generate_trait] attribute here as it is just one implementation of Add trait method of corelib
 11-3: A few errors with imports (pedersen instead of poseidon at some place)
 11-3: Serde and copy attributes are not necessary for StructToHash

They might have been already fixed/moved.

PR #644, opened 2024-02-29
ch11 add serialization explanation and minor additions

kwkr:
Fix some issues mentioned in #568

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/644)

### DIFF
--- a/listings/ch11-advanced-features/no_listing_03_hash_trait/src/lib.cairo
+++ b/listings/ch11-advanced-features/no_listing_03_hash_trait/src/lib.cairo
@@ -6,6 +6,12 @@ trait HashStateTrait<S> {
     fn finalize(self: S) -> felt252;
 }
 
+/// Extension trait for hash state accumulators.
+trait HashStateExTrait<S, T> {
+    /// Updates the hash state with the given value.
+    fn update_with(self: S, value: T) -> S;
+}
+
 /// A trait for values that can be hashed.
 trait Hash<T, S, +HashStateTrait<S>> {
     /// Updates the hash state with the given value.
--- a/src/appendix-03-derivable-traits.md
+++ b/src/appendix-03-derivable-traits.md
@@ -69,6 +69,8 @@ Here is an example:
 
 `Serde` provides trait implementations for `serialize` and `deserialize` functions for data structures defined in your crate. It allows you to transform your structure into an array (or the opposite).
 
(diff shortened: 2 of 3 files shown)


TAdev0:
@kwkr  thanks for the add of HashStateExTrait !

I'm not sure that the wikipedia link is accurate pour serialization. @enitrat  What do you think of adding a short explanation of what serialization is here?

enitrat (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/644)* status: all files reviewed, 1 unresolved discussion (waiting on @kwkr)

___
*[`src/ch11-05-hash.md` line 49 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/644#-NrtCxNHAIGxZI5uYBFT:-NrtCxNHAIGxZI5uYBFU:b-4pbdi2) ([raw file](https://github.com/cairo-book/cairo-book/blob/bf9dedfabc61077f88ecef5ea83716e36ca1ef8f/src/ch11-05-hash.md#L49)):*
> ```Markdown
> Pedersen is different from Poseidon, as it starts with a base state. This base state must be of `felt252` type, which forces us to either hash the struct with an arbitrary base state using the `update_with` method, or serialize the struct into an array to loop through all of its fields and hash its elements together. 
> 
> > **[Serialization](https://en.wikipedia.org/wiki/Serialization)** is a process of transforming data structures into a format that can be easily stored or transmitted. Let's say you are running a program and would like to persist its state to be able to resume it later. To do this, you could take each of the objects your program is using and save information about them, for example in a file. This exactly is a simplified version of serialization. Now if you want to resume your program with this saved state, you would perform **deserialization** which means you would load the state of the objects from some source.
> ```

i think this would be better here https://book.cairo-lang.org/appendix-03-derivable-traits.html?highlight=serialize#serializing-with-serde

kwkr:
@enitrat I moved the serialization note.

enitrat (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/644)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kwkr)

PR #648, opened 2024-03-02
add unit type explanation

kwkr:


enitrat (changes requested):
ty!

> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/648#-:-NryqYBmCWJA9zVKyr1G:b2nidcn)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @kwkr)

___
*[`src/ch02-02-data-types.md` line 206 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/648#-NryoYQ98-15vXY2wk4U:-NryoYQ98-15vXY2wk4V:byleuim) ([raw file](https://github.com/cairo-book/cairo-book/blob/38db6a4224c38251712dad1eff3cac9151fd3ff0/src/ch02-02-data-types.md#L206)):*
> ```Markdown
> Its size is always zero, and it is guaranteed to not exist in the compiled code.
> 
> You might be wondering why you would even need a unit type? In Cairo, functions must always return some value. Even if your function is not meant to return anything, this will be its default return type. You don't have to explicitly specify it when declaring your functions, but it's good to be clear about why it is even there. You will learn more about functions later in the book.
> ```

You might be wondering why you would even need a unit type? In Cairo, everything is an expression, and an expression that returns nothing actually returns `()` implicitly.



_Suggestion:_
```Markdown
You might be wondering why you would even need a unit type? In Cairo, functions must always return some value. Even if your expression is not meant to return anything, this will be its default return type.
```

src/ch02-02-data-types.md
```diff
@@ -203,5 +203,7 @@ A _unit type_ is a type which has only one value `()`.
 It is represented by a tuple with no elements.
 Its size is always zero, and it is guaranteed to not exist in the compiled code.
 
+You might be wondering why you would even need a unit type? In Cairo, functions must always return some value. Even if your expression is not meant to return anything, this will be its default return type. You don't have to explicitly specify it when declaring your functions, but it's good to be clear about why it is even there. You will learn more about functions later in the book.
```
enitrat, line 206:
```suggestion
You might be wondering why you would even need a unit type? In Cairo, everything is an expression, and an expression that returns nothing actually returns `()`  implicitly.
```

enitrat (approved):
sry i messed up the suggestion -all good

PR #649, opened 2024-03-02
add references + stylistics

kwkr:
Added some links to the mentioned concepts + some minor stylistic fixes.

Two things I'm not sure about:

[Here](https://book.cairo-lang.org/ch07-02-defining-modules-to-control-scope.html#grouping-related-code-in-modules) it states that code within modules is private by default but [here](https://book.cairo-lang.org/ch07-05-separating-modules-into-different-files.html#summary) it's public by default. I corrected it to private, but please have a closer look. 

[Here](https://book.cairo-lang.org/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#using-external-packages-in-cairo-with-scarb) git branch imports are mentioned as deprecated but I couldn't find a reference to it in the current [Scarb docs](https://docs.swmansion.com/scarb/docs/guides/dependencies.html). I can fix it if it's needed + maybe add the reference.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/649)

### DIFF
--- a/src/ch05-03-method-syntax.md
+++ b/src/ch05-03-method-syntax.md
@@ -66,7 +66,7 @@ Let's write a new method `scale` which resizes a rectangle of a `factor` given a
 
 It is also possible to define a method which takes ownership of the instance by using just `self` as the first parameter but it is rare. This technique is usually used when the method transforms `self` into something else and you want to prevent the caller from using the original instance after the transformation.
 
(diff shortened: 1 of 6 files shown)


enitrat (approved):
Great improvements! indeed, module code _used to_ be public by default, recent changes made it private by default, we missed that one.

PR #665, opened 2024-03-06
feat: scarb version

suiwater:
add scarb version
add Sierra language

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/665)

### DIFF
--- a/src/ch01-01-installation.md
+++ b/src/ch01-01-installation.md
@@ -42,18 +42,25 @@ and set a global version:
 asdf global scarb 2.5.3
 ```
 
+- You can verify installation by running the following command in a new terminal session, it should print both Scarb, Cairo and Sierra language versions, e.g:
+
+```bash
+scarb 2.5.3 (46d5d5cea 2024-02-01)
+cairo: 2.5.3 (https://crates.io/crates/cairo-lang-compiler/2.5.3)
+sierra: 1.4.0
+```
+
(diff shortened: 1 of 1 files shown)


TAdev0:
hey @suiwater. I'm not sure this modification is useful, you are just writing twice the same thing. It seems better to explain how to install with ASDF,  then with Scarb, an finally show the  output version which will be the same with both methods.

But scarb 2.6.0 was just released yesterday, therefore you can modify all mentions of scarb version 2.5.3 in the book with 2.6.0 if you want :)

src/ch01-01-installation.md
```diff
@@ -42,18 +42,26 @@ and set a global version:
...
+```
+
 Otherwise, you can simply run the following command in your terminal, and follow the onscreen instructions. This will install the latest stable release of Scarb.
 
 ```bash
 curl --proto '=https' --tlsv1.2 -sSf https://docs.swmansion.com/scarb/install.sh | sh
 ```
-
-- In both cases, you can verify installation by running the following command in a new terminal session, it should print both Scarb and Cairo language versions, e.g:
+- You can verify installation by running the following command in a new terminal session, it should print both Scarb, Cairo and Sierra language versions, e.g:
```
enitrat, line 59: this is a repetition from the text above

suiwater: yes

suiwater:
https://github.com/cairo-book/cairo-book/issues/666
close it.

PR #671, opened 2024-03-06
fix option error

suiwater:
https://github.com/starkware-libs/cairo/blob/a14705ac42bfc25e7ec2f87ec1d0c4ec781b5820/corelib/src/option.cairo#L3-L6

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/671)

### DIFF
--- a/src/ch06-01-enums.md
+++ b/src/ch06-01-enums.md
@@ -66,12 +66,12 @@ Running this code would print `quitting`.
 
 ## The `Option` Enum and Its Advantages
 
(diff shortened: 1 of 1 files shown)


TAdev0:
@suiwater  this is not an error and therefore there is no fix to do. 

We are describing the Option enum, highlighting the fact that Option::None returns the unit type. 
But indeed, when declaring an enum containing a variant without associated value, we can either be explicit with `()` or just omit it

PR #674, opened 2024-03-11
add ref in self: ContractState

suiwater:
maybe it's better this way

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/674)


enitrat (commented):
I don't really agree with this change, the "generic" argument is `self: ContractState` and we use modifiers to specify if passed by ref or @. I think the sentence already conveys this idea well

PR #678, opened 2024-03-14
Add div_mod example

feltroidprime:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/678)


TAdev0:
Hi @feltroidprime  , thanks for the PR.

I think we shouldn't mention this as this section describes native functionalities of cairo, and `div_rem` is imported form the core library. @enitrat  what do you think?

enitrat:
agreed as discussed on telegram

PR #682, opened 2024-03-15
dev: add scarb similar cargo

suiwater:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/682)


enitrat:
It's a répétition of the paragraph right below

PR #686, opened 2024-03-19
Fix public function description

Uniblake:
Technically there is no view function in Starknet that ensures no state change happens during an external call. Calling a view function in an untrusted contract does not prevent the callee from reentering the main contract and triggering functions that change the state.

Hence I rephrased the public function description and fully removed the view function description to avoid any misunderstanding.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/686)

### DIFF
--- a/src/ch14-02-contract-functions.md
+++ b/src/ch14-02-contract-functions.md
@@ -18,21 +18,22 @@ Some important rules to note:
 
 ## 2. Public functions
 
-As stated previously, public functions are accessible from outside of the contract. They are usually defined inside an implementation block annotated with the `#[abi(embed_v0)]` attribute. This attribute means that all functions embedded inside it are implementations of the Starknet interface, and therefore entry points of the contract. It only affects the visibility (public vs private/internal), but it doesn't inform us on the ability of these functions to modify the state of the contract. 
+As stated previously, public functions are accessible from outside of the contract. They are usually defined inside an implementation block annotated with the `#[abi(embed_v0)]` attribute. This attribute means that all functions embedded inside it are implementations of the Starknet interface, and therefore entry points of the contract. It only affects the visibility (public vs private/internal), but it doesn't inform us on the ability of these functions to modify the state of the contract.
 
 ```rust,noplayground
 {{#include ../listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo:impl_public}}
 ```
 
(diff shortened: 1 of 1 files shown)


TAdev0:
Hi @Uniblake ,

We pushed a PR a few days ago that introduced this subtlety in the previous chapter (anatomy of a simple contract):

```
A view function is a public function that is typically read-only and cannot mutate the state of the contract. However, this limitation is only enforced by the compiler, and not by Starknet itself. We will discuss the implications of this in a later section. get is a view function.
```

This section in chapter 14 needs to be adjusted as well.
I think it is better to keep the `view function` section, @enitrat  what do you think?

enitrat:
I like the changes made in this PR - very clear. This was a debate that I had a few months back about the terminology of entrypoints in starknet.

My initial proposal was that we get rid of these confusing terms inherited from solidity. 

What we need is
- A term that refers to functions that you can query with `starknet_call` - typically known as `view` functions
- A term that refers to functions mutating state that you can _only_ invoke with `starknet_invoke` - typically known as `external` functions.
- A mention that these restrictions are _not_ enforced by the network.

The reason why I want to keep these terms is that it's how they're defined by the language themselves when compiling your contracts - consider this ABI for example

```json
{
      "type": "interface",
      "name": "tempor::IHelloStarknet",
      "items": [
        {
          "type": "function",
          "name": "increase_balance",
          "inputs": [{ "name": "amount", "type": "core::felt252" }],
          "outputs": [],
          "state_mutability": "external"
        },
        {
          "type": "function",
          "name": "get_balance",
          "inputs": [],
          "outputs": [{ "type": "core::felt252" }],
          "state_mutability": "view"
        }
      ]
    }
 ```
 
 The `state_mutability` field uses both of these terms.

However, it's true that in starknet nothing prevents you from using `starknet_invoke` on a "view" function and thus mutating the state. This must be explicitly stated as you did.

So TLDR: keep most of your suggested changes, but keep using the `view` and `external` terms.

Uniblake:
Thank you for your comments! I will add them back and rephrase my changes.

Uniblake:
I added the view and external terms back, you are right this is "cairo" book not "starknet" book ;). It took me quite some time to think about how to avoid confusing the readers. Let me know if you feel anything still needs to be adjusted.

enitrat (dismissed):
perfect!

Issue #596, opened 2024-02-13
feat : 2.6.0


TAdev0:
- [x] if - let expression
- [x] while let support.
- [x] assert_lt!, assert_le!, assert_ge!, assert_gt! in tests
- [x] underscore pattern in match on tuples (ie you can do `_ =>` instead of (_ , _) =>`
- [x] Support for complex consts (Non-Zero, struct, enum, fixed size arrays).
- [x] Change all ` loop { arr.pop_front ...} ` loops to `while let Option::Some` loops - except the ones used for the `loop` keyword explicitly
- [x] Const array type https://doc.rust-lang.org/book/ch03-02-data-types.html#the-array-type

TAdev0:
https://github.com/starkware-libs/cairo/releases

okhaimie-dev:
@TAdev0 @enitrat 

Seems v2.6.0 is no more a release candidate and out, can I make a PR to include 

- `if let`
- `while let`

to the chapter on [control flow](https://book.cairo-lang.org/ch02-05-control-flow.html#control-flow)?

TAdev0:
Go ahead @okhaimie-dev  !

enitrat:
relevant pages:

https://doc.rust-lang.org/stable/book/ch06-03-if-let.html

I could see both `if let` and `while let` fit in here

okhaimie-dev:
@TAdev0 can you create a `ch06-03` file for this as eni suggested? I'd do it myself, but you asked me not to touch `SUMMARY.md` file to avoid conflicts.

I am going to get started writing on a throwaway branch while I wait for you to make the file addition.

TAdev0:
@okhaimie-dev you can modify the summary file now no worries :) you can create the file

TAdev0:
@enitrat  should we do a section about gas coupon?

enitrat:
No

PR #692, opened 2024-03-20
feat 2.6.0 complex consts

TAdev0:
#596

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/692)


TAdev0:
@enitrat  not sure that mentioning complex consts (struct enum, fixed size array) in data types chapter is ideal, what do you think? It might be good though, as it is just a small quotation that is it possible to do that

src/ch02-01-variables-and-mutability.md
```diff
@@ -108,23 +108,34 @@ Constants can only be declared in the global scope, which makes
...
 Nonetheless, it is possible to use the `consteval_int!` macro to create a `const` variable that is the result of some computation:
 
 ```rust, noplayground
-const ONE_HOUR_IN_SECONDS: u32 = consteval_int!(60 * 60);
+{{#include ../listings/ch02-common-programming-concepts/no_listing_00_consts/src/lib.cairo:consteval_const}}
 ```
 
 We will dive into more detail about macros in the [dedicated section](./ch11-06-macros.md).
 
+We can also declare more complexe `const` variables, including structs and enums that we will discuss in later chapters:
```
enitrat, line 127: Let's think about it the other way: which types cannot be instanciated as constants? 

My expectation is that I can create a const value of any type. Previously this was not the case and might have come surprising to the user. Now that we can have struct / enum variants as constants I think we should not enumerate what type you can create constants out of: the idea is just to introduce the _concept_ of const variables, which should transitively be applicable to all relevant types

TAdev0: what do you think now?

enitrat (approved):
great!

PR #708, opened 2024-03-22
fix: add SafeDispatcher

suiwater:
I don't understand what SafeDispatcher means when I see it elsewhere. So we need to add it here

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/708)


enitrat:
good, but you need to explain what's the difference between the two

TAdev0:
This will be part of the current rework of the cross contract section. I close this issue.

PR #709, opened 2024-03-22
fixed size array type

### LINKED ISSUE #596: feat : 2.6.0

TAdev0:
Closes #596 

@enitrat  i didn't create a dedicated file for the snippets because they are one line and it wont compile for now. Do you prefer to put these variable declarations in a main function with a TAG does_not_compile?

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/709)

### DIFF
--- a/src/ch02-02-data-types.md
+++ b/src/ch02-02-data-types.md
@@ -116,7 +116,7 @@ expression. We’ll cover how `if` expressions work in Cairo in the [Control Flo
 
 Cairo doesn't have a native type for strings but provides two ways to handle them: short strings using simple quotes and ByteArray using double quotes.
 
-#### Short strings
+#### Short Strings
 
 A short string is an ASCII string where each character is encoded on one byte (see the [ASCII table](https://www.asciitable.com/)). For example:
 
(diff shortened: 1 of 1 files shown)


enitrat:
@TAdev0 why do they not compile?

TAdev0:
@enitrat because the implementation of Drop is not yet a in a release

enitrat:
ok, then let's keep this open until it's properly supported

misicnenad:
@TAdev0 can this PR be reopened?
Just found out about const arrays, and I wish I knew about them sooner

PR #711, opened 2024-04-01
feat: inlining in Cairo

### LINKED ISSUE #268: Advanced Cairo: Inlining in Cairo
Ressources talking about inlining in Rust:

- https://matklad.github.io/2021/07/09/inline-in-rust.html
- https://doc.rust-lang.org/nightly/reference/attributes/codegen.html?highlight=inline#the-inline-attribute

In Cairo, it's a bit different - you basically avoid using a `function_call` libfunc by directly injecting the sierra code in the caller function's context. 

Inlining can, in some cases, improve performances, and should be considered when it makes sense. 
In this part, we will outline the mechanisms of inlining in Cairo, provide an in-depth example with a very basic comparison between inlined / not-inlined functions at the Sierra level / CASM level, and provide insights on when to / not to use inlining.

est. time: 3 to 4 hours

TAdev0:
Closes #268


enitrat (changes requested):
Great chapter, you understood it perfectly. Will probably need some improvements on the presentation of the Sierra file, will do a second round later

> Reviewed 6 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711#-:-NuQigWJ1AiS_bblJU8X:b97bz6o)* status: 6 of 10 files reviewed, 13 unresolved discussions (waiting on @TAdev0)

___
*[`listings/ch11-advanced-features/listing_02_inlining/src/lib.cairo` line 1 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQdAWaC4t9Nz-iMWXQ:-NuQdAWaC4t9Nz-iMWXR:b-ikpolx) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/listings/ch11-advanced-features/listing_02_inlining/src/lib.cairo#L1)):*
> ```Cairo
> fn main() {
> ```

I would use an example a lot simpler than ByteArray, which is likely to generate complex code. Can you use a simple program with only felts ?
___
*[`listings/ch11-advanced-features/listing_02_inlining/src/lib.cairo` line 14 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQdH5A3raRif-3EVU5:-NuQdH5B-yKr9-QUePnY:b-cpwpq9) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/listings/ch11-advanced-features/listing_02_inlining/src/lib.cairo#L14)):*
> ```Cairo
> }
> 
> fn not_inlined() -> ByteArray {
> ```

I would add an attribute to force non-inlining
___
*[`src/ch11-07-inlining-in-cairo.md` line 3 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQeQCEAdovgIT4cLvp:-NuQeQCEAdovgIT4cLvq:b-wnimed) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L3)):*
> ```Markdown
> # Inlining in Cairo
> 
> The process of _inlining_ is a well-known method for optimizing code execution. It is supported by many compilers, whether automatically or not. Basically, inlining allows to replace a call to a function with its actual body. In certain situations, such a transformation might improve performance by eliminating the overhead of the function call itself, leading to a faster code generation.
> ```

"faster code generation" is ambiguous. It's not clear if it refers to the speed of the compilation process or the speed of the generated code (or both?).

In reality it leads to improvements in efficiency: if you avoid the overhead of a function call for small functions, your program will execute less steps.



_Code quote:_
```Markdown
leading to a faster code generation.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 3 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQeeAI5AB8BqGc2XlE:-NuQeeAI5AB8BqGc2XlF:bw3p3n0) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L3)):*
> ```Markdown
> # Inlining in Cairo
> 
> The process of _inlining_ is a well-known method for optimizing code execution. It is supported by many compilers, whether automatically or not. Basically, inlining allows to replace a call to a function with its actual body. In certain situations, such a transformation might improve performance by eliminating the overhead of the function call itself, leading to a faster code generation.
> ```

Compilers typically do this automatically (this should not be left to the user to decide in most cases). In the case of Cairo, it is performed automatically using a heuristic about the size of the function.

I think the overall paragraph can be improved in clarity and didactism

"Inlining is a common code optimization technique supported by most compilers. It involves replacing a function call at the call site with the actual code of the called function, eliminating the overhead associated with the function call itself. This can improve performance by reducing the number of instructions executed, but may increase the total size of the program. When you're thinking about whether to inline a function, take into account things like how big it is, what parameters it has, how often it gets called, and how it might affect the size of your compiled code."



_Code quote:_
```Markdown
It is supported by many compilers, whether automatically or not.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 7 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQgRNI4I2IIOQei-aB:-NuQgRNI4I2IIOQei-aC:bvzrpdw) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L7)):*
> ```Markdown
> ## The `inline` attribute
> 
> In Cairo, the `inline` attribute suggests that the Sierra code corresponding to the attributed function should be directly injected in the caller function's context, rather than using a `function_call` libfunc to execute that code.
> ```

not really accurate: it has three variants one of which tells the compiler to not inline at all, which is a contradiction with this description



_Code quote:_
```Markdown
attribute suggests that the Sierra code corresponding to the attributed function should be directly injected in the caller function's context, rather than using a `function_call` libfunc to execute that code.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 14 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQh-KA6BAMI9MtJNA2:-NuQh-KA6BAMI9MtJNA3:bhhd6bm) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L14)):*
> ```Markdown
> - `#[inline(never)]` suggests that an inline expansion should never be performed.
> 
> > Note: the `inline` attribute in every form is a hint, with no requirements on the language to place a copy of the attributed function in the caller. This means that the attribute may be ignored by the compiler.
> ```

Why would the compiler ignore this? Is there a rule we can study?
___
*[`src/ch11-07-inlining-in-cairo.md` line 16 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQhDLn8oHvqPonjvoi:-NuQhDLn8oHvqPonjvoj:bbzd4ks) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L16)):*
> ```Markdown
> > Note: the `inline` attribute in every form is a hint, with no requirements on the language to place a copy of the attributed function in the caller. This means that the attribute may be ignored by the compiler.
> 
> Many of the Cairo corelib functions are inlined. User-defined functions may also be annotated with the `inline` attribute. Doing this reduces the total number of steps required when calling these attributed functions. Indeed, injecting the Sierra code at the caller site avoids the step-cost involved in calling functions and obtaining their arguments.
> ```

Do you know what this cost is?



_Code quote:_
```Markdown
the step-cost involved in calling functions and obtaining their arguments.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 16 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQhXEiAbzP7ANmrSY8:-NuQhXEiAbzP7ANmrSY9:bhjaltf) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L16)):*
> ```Markdown
> > Note: the `inline` attribute in every form is a hint, with no requirements on the language to place a copy of the attributed function in the caller. This means that the attribute may be ignored by the compiler.
> 
> Many of the Cairo corelib functions are inlined. User-defined functions may also be annotated with the `inline` attribute. Doing this reduces the total number of steps required when calling these attributed functions. Indeed, injecting the Sierra code at the caller site avoids the step-cost involved in calling functions and obtaining their arguments.
> ```

Annoting a function with an inlining attribute reduces the



_Code quote:_
```Markdown
Many of the Cairo corelib functions are inlined. User-defined functions may also be annotated with the `inline` attribute. Doing this
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 20 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQhjJC7pefOb37Y2b_:-NuQhjJC7pefOb37Y2ba:bhgf2hr) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L20)):*
> ```Markdown
> However, inlining can also lead to increased code size. Whenever a function is inlined, the call site contains a copy of the function's Sierra code, potentially leading to duplication of code across the codebase.
> 
> Therefore, inlining should be applied with cautious. Using the `#[inline]` or `#[inline(always)]` indiscriminately will lead to increased compile time. It is particularly useful to inline small functions, ideally with many arguments. The more frequently a function is called, the more beneficial inlining becomes in terms of performance. By doing so, the number of steps for the execution will be lower, while the code length will not grow that much or might even decrease in terms of total number of instructions.
> ```

echoes my question above: why ? (it is better to understand the _why_ than just being told why). 



_Code quote:_
```Markdown
It is particularly useful to inline small functions, ideally with many arguments.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 44 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQi8juDlfc4xH210uO:-NuQi8juDlfc4xH210uP:bombqjg) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L44)):*
> ```Markdown
> ```
> 
> Both `inlined` and `not_inlined` functions output their respective return values in the same manner, regardless of whether they are inlined or not.
> ```

I don't think this is valuable



_Code quote:_
```Markdown
First of all, let's run our program and see what the result is:

```shell
$ scarb cairo-run
First function call is inlined
Second function call is not inlined
Run completed successfully, returning []
```

Both `inlined` and `not_inlined` functions output their respective return values in the same manner, regardless of whether they are inlined or not.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 46 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQiBUJ9V_THcC8wrNF:-NuQiBUJ9V_THcC8wrNG:brhgl2a) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L46)):*
> ```Markdown
> Both `inlined` and `not_inlined` functions output their respective return values in the same manner, regardless of whether they are inlined or not.
> 
> Let's take a look at the corresponding Sierra code to see how inlining works under the hood. The `println!` macro usage here will greatly increase the Sierra code size while adding a lot of complexity in it, this is why we will focus on the following program that only calls  `inlined` and `not_inlined` functions without printing their return values: 
> ```

echoes my remark on the code itself: we should get rid of it and keep the example as short as possible.



_Code quote:_
```Markdown
The `println!` macro usage here will greatly increase the Sierra code size while adding a lot of complexity in it, this is why we will focus on the following program that only calls  `inlined` and `not_inlined` functions without printing their return values:
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 64 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQiRJp18qcimkGG9t3:-NuQiRJp18qcimkGG9t4:b-cqayrs) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L64)):*
> ```Markdown
> - Type and libfunc declarations.
> - Statements that constitute the program.
> - Declaration of the functions of the program.
> ```

The Sierra file is structured in three parts:
...



_Code quote:_
```Markdown
It includes:
- Type and libfunc declarations.
- Statements that constitute the program.
- Declaration of the functions of the program.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 66 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQie-k1_-5URoycYVO:-NuQie-k1_-5URoycYVP:b-6wx633) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L66)):*
> ```Markdown
> - Declaration of the functions of the program.
> 
> The Sierra code statements always match the order of function declarations in the Cairo program. This means that the `return([2])` statement on line 26, which is the first `return` instruction of the Sierra program, corresponds to the `main` function return instruction. The `return([3])` statement on line 32 corresponds to the `inlined` function return instruction, and the `return([3])` on line 38 is the return instruction of the `not_inlined` function.
> ```

Will review this once the base code used will be changed

TAdev0:
@enitrat  ready for review after updates

TAdev0 (commented):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 3 of 12 files reviewed, 5 unresolved discussions (waiting on @enitrat)

___
*[`src/ch11-07-inlining-in-cairo.md` line 18 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTvtGpDc-AG6TxEbu0:-NuUdR73BwfBiGEitSn1:b-896fix) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L18)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

compiled code.
</blockquote></details>

Done.
___
*[`src/ch11-07-inlining-in-cairo.md` line 65 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTx2cy6r8Q1ValYSME:-NuUoKx4DPGzAULG2NFI:b-2edn4i) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L65)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

It kills a bit the purpose of our demonstration if the compiler throws out this part of unused code...

Proposal: make `inlined` return 1, `not_inlined` return 2, add 1+2 in main and return the result

This should force the use of both return values from both functions
</blockquote></details>

yeah i was thinking the same, lets do this
___
*[`src/ch11-07-inlining-in-cairo.md` line 89 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTwi1ND4TGG4NunbGZ:-NuUoQRnEO5gzR4hWyR4:bobt98i) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L89)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I think we need one extra section about CASM to understand the benefits from the actually executed instructions. It should be a very simple program.

You will also be able notice and perhaps extrapolate the general rule regarding steps associated to function calls
</blockquote></details>

i tried looking at casm code, to be honest i cannot understand a lot for now, i need to learn more about casm opcodes

enitrat (changes requested):
> Reviewed 1 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 3 unresolved discussions (waiting on @TAdev0)

___
*[`src/ch11-07-inlining-in-cairo.md` line 89 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTwi1ND4TGG4NunbGZ:-NuUsUbZ-7jumnARtpHP:b-jkagyo) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L89)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

i tried looking at casm code, to be honest i cannot understand a lot for now, i need to learn more about casm opcodes
</blockquote></details>

it should be fairly simple in this case, as you should only have `call`, `ret` and addition of cells

TAdev0 (commented):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 3 unresolved discussions (waiting on @enitrat)

___
*[`src/ch11-07-inlining-in-cairo.md` line 89 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTwi1ND4TGG4NunbGZ:-NuV2E7G2iIAXrE3y0Mo:b2firs0) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L89)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

it should be fairly simple in this case, as you should only have `call`, `ret` and addition of cells
</blockquote></details>

here is the casm code for the first example adding inlined and not_inlined and returning the value:

```
1	call rel 3
2	ret
3	call rel 9
4	[ap + 0] = 1, ap++
5	[ap + 0] = [ap + -1] + [ap + -2], ap++
6	ret
7	[ap + 0] = 1, ap++
8	ret
9	[ap + 0] = 2, ap++
10	ret
11	ret
```

spent an hour on it with the cairo paper, it seems kind of cryptic so far

I understand that in the first line, `call rel` updates the pc and fp to `3` ? The cairo paper also talks about `call abs <address>` , no idea what is the difference its not explained.

And `ret` is supposed to reset the fp;
All of this to create a temporary stack, with fp initially equal to ap at the beginning of a function and then incremented and finally reset with `ret`

so yeah, we can deduce that the second line with `ret` means that this is the end of the `not_inlined` function, whose execution began when `call` updated the PC to the memory cell that contains its first instruction

and then another `call rell` to update to `9`, it shouldnt be a cell that contains the `inline` function first instruction is it is inlined, and the lines after with [ap...]. are , i guess, the inlined body of the `inline` function

I dont really understand from lines 3 to 11

enitrat (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 3 unresolved discussions (waiting on @TAdev0)

___
*[`src/ch11-07-inlining-in-cairo.md` line 89 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTwi1ND4TGG4NunbGZ:-NuVQKXFEwzPUkyOyIMb:b5fo9g3) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L89)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

here is the casm code for the first example adding inlined and not_inlined and returning the value:

```
1	call rel 3
2	ret
3	call rel 9
4	[ap + 0] = 1, ap++
5	[ap + 0] = [ap + -1] + [ap + -2], ap++
6	ret
7	[ap + 0] = 1, ap++
8	ret
9	[ap + 0] = 2, ap++
10	ret
11	ret
```

spent an hour on it with the cairo paper, it seems kind of cryptic so far

I understand that in the first line, `call rel` updates the pc and fp to `3` ? The cairo paper also talks about `call abs <address>` , no idea what is the difference its not explained.

And `ret` is supposed to reset the fp;
All of this to create a temporary stack, with fp initially equal to ap at the beginning of a function and then incremented and finally reset with `ret`

so yeah, we can deduce that the second line with `ret` means that this is the end of the `not_inlined` function, whose execution began when `call` updated the PC to the memory cell that contains its first instruction

and then another `call rell` to update to `9`, it shouldnt be a cell that contains the `inline` function first instruction is it is inlined, and the lines after with [ap...]. are , i guess, the inlined body of the `inline` function

I dont really understand from lines 3 to 11
</blockquote></details>

is the CASM file only 11 lines long?

Here's what I understand. Considering that each instruction offsets the PC by 1 and that each argument to these instructions as well. `call` is some kind of abstraction around a jump which, as you correctly pointed, updates pc and fp to simulate function frames

call rel 3 ->  increment pc by 3 and execute the instrution at this location. here, if we increment by 3, we get to `call rel 9` (argument `3` takes 1, ret takes `1`, next one is `call`)

`call rel 9` -> we get to the line `[ap + 0] = 2, ap++`

we execute `[ap + 0] = 2, ap++` so now in [ap-1] is stored 2

we `ret` back to the line `[ap + 0] = 1, ap++`

now: `[ap-1] = 1, [ap-2] = 2`

we sum these in `[ap + 0] = ...` s.t. `[ap-1] = 3, [ap-2] = 1, [ap-3]=3`

and then `ret` which is the final instruction executed

enitrat (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 3 unresolved discussions (waiting on @TAdev0)

___
*[`src/ch11-07-inlining-in-cairo.md` line 89 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTwi1ND4TGG4NunbGZ:-NuVVGlY6bAVNkMjGNnI:bwda9r3) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L89)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

is the CASM file only 11 lines long?

Here's what I understand. Considering that each instruction offsets the PC by 1 and that each argument to these instructions as well. `call` is some kind of abstraction around a jump which, as you correctly pointed, updates pc and fp to simulate function frames

call rel 3 ->  increment pc by 3 and execute the instrution at this location. here, if we increment by 3, we get to `call rel 9` (argument `3` takes 1, ret takes `1`, next one is `call`)

`call rel 9` -> we get to the line `[ap + 0] = 2, ap++`

we execute `[ap + 0] = 2, ap++` so now in [ap-1] is stored 2

we `ret` back to the line `[ap + 0] = 1, ap++`

now: `[ap-1] = 1, [ap-2] = 2`

we sum these in `[ap + 0] = ...` s.t. `[ap-1] = 3, [ap-2] = 1, [ap-3]=3`

and then `ret` which is the final instruction executed
</blockquote></details>

wait, actually! why don't you use https://cairovm.codes/ you will be able to see step-by-step execution

the interesting thing here is that we can see the function call for the non-inlined function but the other one is correctly inlined

TAdev0 (commented):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 3 unresolved discussions (waiting on @enitrat)

___
*[`src/ch11-07-inlining-in-cairo.md` line 89 at r3](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuTwi1ND4TGG4NunbGZ:-NuYKefN5bsTeH9JJ19O:b9ovj1o) ([raw file](https://github.com/cairo-book/cairo-book/blob/d4e3fa259041b45b5c59b958c0f364ceed901584/src/ch11-07-inlining-in-cairo.md#L89)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

wait, actually! why don't you use https://cairovm.codes/ you will be able to see step-by-step execution

the interesting thing here is that we can see the function call for the non-inlined function but the other one is correctly inlined
</blockquote></details>

yes it is 11 lines long only

I tried something, i think i understood the casm code, please correct anything that might be wrong

enitrat (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 7 unresolved discussions (waiting on @TAdev0)

___
*[`listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra` line 23 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0N2v9qrLFhYjeOLX:-NuZ0N2v9qrLFhYjeOLY:b-34q6el) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra#L23)):*
> ```
> 19	felt252_const<2>() -> ([0]) // 9
> 20	store_temp<felt252>([0]) -> ([0]) // 10
> 21	return([0]) // 11
> ```

I think it would be better to only number these lines, as they correspond to the actual sierra program. In the function declaration section we can see that main is at line0, inlined at line 6, not inlined at line 9



_Code quote:_
```
10	function_call<user@main::main::not_inlined>() -> ([0]) // 0
11	felt252_const<1>() -> ([1]) // 1
12	store_temp<felt252>([1]) -> ([1]) // 2
13	felt252_add([1], [0]) -> ([2]) // 3
14	store_temp<felt252>([2]) -> ([2]) // 4
15	return([2]) // 5
16	felt252_const<1>() -> ([0]) // 6
17	store_temp<felt252>([0]) -> ([0]) // 7
18	return([0]) // 8
19	felt252_const<2>() -> ([0]) // 9
20	store_temp<felt252>([0]) -> ([0]) // 10
21	return([0]) // 11
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 48 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0ssT-IyeYGMTesIz:-NuZ0ssT-IyeYGMTesJ-:b-4fft5d) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L48)):*
> ```Markdown
> - Declaration of the functions of the program.
> 
> The Sierra code statements always match the order of function declarations in the Cairo program. This means that the `return([2])` statement on line 15, which is the first `return` instruction of the Sierra program, corresponds to the `main` function return instruction. The `return([0])` statement on line 18 corresponds to the `inlined` function return instruction, and the `return([0])` on line 21 is the return instruction of the `not_inlined` function.
> ```

Perhaps an explanation on how the main sierra program is divided by functions before. we know that main is 0-5, inlined is 6-8, notinlined is 9-11 thanks to the function declaration section. It makes things clearer
___
*[`src/ch11-07-inlining-in-cairo.md` line 61 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0pSQ-n6VJWfIDWE-:-NuZ0pSQ-n6VJWfIDWE0:b-7cbclq) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L61)):*
> ```Markdown
> ```
> 
> On line 10, we can see the call to the `function_call` libfunc instantiated with the `not_inlined` function and previously declared on line 4, which will execute all the code from lines 19 to 20: 
> ```

wording problem



_Code quote:_
```Markdown
On line 10, we can see the call to the `function_call` libfunc instantiated with the `not_inlined` function and previously declared on line 4,
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 72 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0mROCZ1WcNMqGQS2:-NuZ0mROCZ1WcNMqGQS3:bcktdhz) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L72)):*
> ```Markdown
> ```rust,noplayground
> 16	felt252_const<1>() -> ([0])
> 17	store_temp<felt252>([0]) -> ([0])
> ```

I'm not sure what this is supposed to mean
___
*[`src/ch11-07-inlining-in-cairo.md` line 92 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ1Aj5B9P2kp-6FbQH:-NuZ1Aj69hi0yJU5g-Rq:bqn2s0q) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L92)):*
> ```Markdown
> ```
> 
> Now, let's take a look at the Casm code corresponding to this program to really understand the benefits of inlining.
> ```

I think this section is too technical for the purpose of the demonstration. We can probably illustrate it better by simply identifying the flow of the program, separating each function in blocks, and illustrate the execution process with arrows.

call rel 3 -> call rel 9 -> [ap + 0] = 2, ap++ -> [ap + 0] = 1, ap++ -> [ap + 0] = [ap + -1] + [ap + -2], ap++ -> ret

In the context of this chapter knowing what call and ret do exactly is not beneficial and is quite confusing, we can introduce them as simple function calls and return I believe.

It's also hard to textually describe this, perhaps an image would be easier to understand
___
*[`src/ch11-07-inlining-in-cairo.md` line 136 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ3HBn-BVBGbIjWN6-:-NuZ3HBn-BVBGbIjWN60:bp7a7ya) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L136)):*
> ```Markdown
> > The Cams code of our program clearly shows that there is a function call for the `non_inlined` function, while the `inlined` function is correctly inlined.
> 
> ## Additional Optimizations
> ```

I think we can remove this

TAdev0 (commented):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 7 unresolved discussions (waiting on @enitrat)

___
*[`listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra` line 23 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0N2v9qrLFhYjeOLX:-NuZ7CY3EkmplGgAGb9b:bz87tjx) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra#L23)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I think it would be better to only number these lines, as they correspond to the actual sierra program. In the function declaration section we can see that main is at line0, inlined at line 6, not inlined at line 9
</blockquote></details>

yeah this is why only the statements are annotated with ` // 0 ... to // 11`.  do you prefer this?

```
// libfunc declarations
libfunc function_call<user@main::main::not_inlined> = function_call<user@main::main::not_inlined>
libfunc felt252_const<1> = felt252_const<1>
libfunc store_temp<felt252> = store_temp<felt252>
libfunc felt252_add = felt252_add
libfunc felt252_const<2> = felt252_const<2>

// statements
00	function_call<user@main::main::not_inlined>() -> ([0]) 
01	felt252_const<1>() -> ([1]) 
02	store_temp<felt252>([1]) -> ([1]) 
03	felt252_add([1], [0]) -> ([2]) 
04	store_temp<felt252>([2]) -> ([2]) 
05	return([2]) // 5
06	felt252_const<1>() -> ([0]) 
07	store_temp<felt252>([0]) -> ([0])
08	return([0]) // 8
09	felt252_const<2>() -> ([0])
10	store_temp<felt252>([0]) -> ([0])
11	return([0])

....
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 61 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0pSQ-n6VJWfIDWE-:-NuZ8_x60SjKEVjNBii3:b-ffreo5) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L61)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

wording problem
</blockquote></details>

what do you mean?
___
*[`src/ch11-07-inlining-in-cairo.md` line 72 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0mROCZ1WcNMqGQS2:-NuZ8uJLB64wxIQZqwgs:b-hvalv6) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L72)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I'm not sure what this is supposed to mean
</blockquote></details>

store_temp seems to be always required to use later the value, either for returning or computation
___
*[`src/ch11-07-inlining-in-cairo.md` line 92 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ1Aj5B9P2kp-6FbQH:-NuZ9P6k8svGqGr7nfhl:b-4ikl0f) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L92)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I think this section is too technical for the purpose of the demonstration. We can probably illustrate it better by simply identifying the flow of the program, separating each function in blocks, and illustrate the execution process with arrows.

call rel 3 -> call rel 9 -> [ap + 0] = 2, ap++ -> [ap + 0] = 1, ap++ -> [ap + 0] = [ap + -1] + [ap + -2], ap++ -> ret

In the context of this chapter knowing what call and ret do exactly is not beneficial and is quite confusing, we can introduce them as simple function calls and return I believe.

It's also hard to textually describe this, perhaps an image would be easier to understand
</blockquote></details>

For me, but this is just my opinion, it doesnt make sense to create a Casm section in that case. The Sierra statements already show the inlining feature and what is different from non inlining. 
I dont see the point of a Casm section where we dont even explain what call / ret are, and where we use them exhaustively with some flow as you described. This would be much more confusing and frustrating, at least for me
___
*[`src/ch11-07-inlining-in-cairo.md` line 136 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ3HBn-BVBGbIjWN6-:-NuZ9H3f6eI-wBLYyn9i:b8unx03) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L136)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I think we can remove this
</blockquote></details>

you want to remove this whole section where we describe additional benefits of the inlining, especially in terms of code length? why?

enitrat (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 4 of 16 files reviewed, 7 unresolved discussions (waiting on @TAdev0)

___
*[`listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra` line 23 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0N2v9qrLFhYjeOLX:-NuZE1FQBIfLqc-tpBzN:ba118kj) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra#L23)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

yeah this is why only the statements are annotated with ` // 0 ... to // 11`.  do you prefer this?

```
// libfunc declarations
libfunc function_call<user@main::main::not_inlined> = function_call<user@main::main::not_inlined>
libfunc felt252_const<1> = felt252_const<1>
libfunc store_temp<felt252> = store_temp<felt252>
libfunc felt252_add = felt252_add
libfunc felt252_const<2> = felt252_const<2>

// statements
00	function_call<user@main::main::not_inlined>() -> ([0]) 
01	felt252_const<1>() -> ([1]) 
02	store_temp<felt252>([1]) -> ([1]) 
03	felt252_add([1], [0]) -> ([2]) 
04	store_temp<felt252>([2]) -> ([2]) 
05	return([2]) // 5
06	felt252_const<1>() -> ([0]) 
07	store_temp<felt252>([0]) -> ([0])
08	return([0]) // 8
09	felt252_const<2>() -> ([0])
10	store_temp<felt252>([0]) -> ([0])
11	return([0])

....
```
</blockquote></details>

yes especially considering that you never actually reference other lines by their number
___
*[`src/ch11-07-inlining-in-cairo.md` line 61 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0pSQ-n6VJWfIDWE-:-NuZEKUi87vH7IhB1-TB:bhggx33) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L61)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

what do you mean?
</blockquote></details>

The `function_call` libfunc is called on line 10 to execute the `not_inlined` function. This will execute the code from lines 19 to 20 and store the return value in the variable with id `0`
___
*[`src/ch11-07-inlining-in-cairo.md` line 72 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0mROCZ1WcNMqGQS2:-NuZEy031aDrtUJfWEj0:b-7jjnow) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L72)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

store_temp seems to be always required to use later the value, either for returning or computation
</blockquote></details>

“The program uses a single data type, `felt252`. It uses two library functions - `felt_const<1>`, which returns the constant felt252 `1`, and `store_temp<felt252>`, which pushes a constant value to memory. The program has a main functio`n` that starts at statement 0 and returns a variable of type `felt252`. During the execution of my program, we call the `felt_const<1>` libfunc to create a variable with id `[0]`. We push this variable to memory and retrieve another variable of id `[1]`, which I return at the end of the function.”

https://arc.net/l/quote/milsawzs

store_temp compiles in CASM to roughly

[ap] = x; ap++
___
*[`src/ch11-07-inlining-in-cairo.md` line 92 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ1Aj5B9P2kp-6FbQH:-NuZFIod-bCRDQl1sYeP:btcrw9p) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L92)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

For me, but this is just my opinion, it doesnt make sense to create a Casm section in that case. The Sierra statements already show the inlining feature and what is different from non inlining. 
I dont see the point of a Casm section where we dont even explain what call / ret are, and where we use them exhaustively with some flow as you described. This would be much more confusing and frustrating, at least for me
</blockquote></details>

In that case we must make sure that the CASM description is entirely accurate. As it is now it's a bit imprecise and just dumps info at the reader by assuming that he will be familiar with what ap, pc, fp are.

If you just do a high-order description, you can easily follow what's going on with the actual execution and you don't really need to know more about the inner workings of cairo.

If I come from another language and see `call` and `ret` I can assume that these statements represent function calls and return instructions. 

If you start saying that call updates PC and FP while ret resets FP to a value prior to call, you jump to the internals directly and lose the focus we had on inlining specifically.

But I understand your point. we just need to think about how to make it easily digestible
___
*[`src/ch11-07-inlining-in-cairo.md` line 136 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ3HBn-BVBGbIjWN6-:-NuZGIBFA4hGcfZCxPDf:begr3kh) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L136)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

you want to remove this whole section where we describe additional benefits of the inlining, especially in terms of code length? why?
</blockquote></details>

ok, sorry my bad I read too fast to understand correctly. it's still a "niche" use case in which the compiler just removes unused code from the compiled code. but let's keep it

TAdev0 (commented):
> Reviewed 3 of 7 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 7 of 16 files reviewed, 7 unresolved discussions (waiting on @enitrat)

___
*[`listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra` line 23 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0N2v9qrLFhYjeOLX:-NuZRqatC4kU4i4yJT6Y:btn720r) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/listings/ch11-advanced-features/listing_03_inlining_example/src/inlining.sierra#L23)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

yes especially considering that you never actually reference other lines by their number
</blockquote></details>

actually i reference once with "previously declared on line 4" but i can modify this. will do this.
___
*[`src/ch11-07-inlining-in-cairo.md` line 48 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0ssT-IyeYGMTesIz:-NuZVTXH-4ZVUmH7vAyv:b-896fja) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L48)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Perhaps an explanation on how the main sierra program is divided by functions before. we know that main is 0-5, inlined is 6-8, notinlined is 9-11 thanks to the function declaration section. It makes things clearer
</blockquote></details>

Done!
___
*[`src/ch11-07-inlining-in-cairo.md` line 61 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0pSQ-n6VJWfIDWE-:-NuZVuS--nNOGYUbN52k:b-vl7y2i) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L61)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

The `function_call` libfunc is called on line 10 to execute the `not_inlined` function. This will execute the code from lines 19 to 20 and store the return value in the variable with id `0`
</blockquote></details>

better! my sentence was not clear. I wanted to say that the function_call libfunc is always instantiated, i.e., you would declare multiple function_call libfunc if you wanted to call different functions. but anyway this is not important
___
*[`src/ch11-07-inlining-in-cairo.md` line 72 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ0mROCZ1WcNMqGQS2:-NuZXLz-5BBz9IgzhX-K:b-25g9uw) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L72)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

“The program uses a single data type, `felt252`. It uses two library functions - `felt_const<1>`, which returns the constant felt252 `1`, and `store_temp<felt252>`, which pushes a constant value to memory. The program has a main functio`n` that starts at statement 0 and returns a variable of type `felt252`. During the execution of my program, we call the `felt_const<1>` libfunc to create a variable with id `[0]`. We push this variable to memory and retrieve another variable of id `[1]`, which I return at the end of the function.”

https://arc.net/l/quote/milsawzs

store_temp compiles in CASM to roughly

[ap] = x; ap++
</blockquote></details>

ok i took inspiration and integrated your quote , it seems better now
___
*[`src/ch11-07-inlining-in-cairo.md` line 92 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ1Aj5B9P2kp-6FbQH:-NuZYoKJBGdnfSMFa2Z8:bh3twcb) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L92)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

In that case we must make sure that the CASM description is entirely accurate. As it is now it's a bit imprecise and just dumps info at the reader by assuming that he will be familiar with what ap, pc, fp are.

If you just do a high-order description, you can easily follow what's going on with the actual execution and you don't really need to know more about the inner workings of cairo.

If I come from another language and see `call` and `ret` I can assume that these statements represent function calls and return instructions. 

If you start saying that call updates PC and FP while ret resets FP to a value prior to call, you jump to the internals directly and lose the focus we had on inlining specifically.

But I understand your point. we just need to think about how to make it easily digestible
</blockquote></details>

will try to improve it tomorrow
___
*[`src/ch11-07-inlining-in-cairo.md` line 136 at r6](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuZ3HBn-BVBGbIjWN6-:-NuZYmtH3E3ZWqwN2IzL:b2ni9b4) ([raw file](https://github.com/cairo-book/cairo-book/blob/4eed3adf77e8a2ff350f811e3a1a306026f5b0b5/src/ch11-07-inlining-in-cairo.md#L136)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

ok, sorry my bad I read too fast to understand correctly. it's still a "niche" use case in which the compiler just removes unused code from the compiled code. but let's keep it
</blockquote></details>

ok!

TAdev0:
@enitrat  can you tell me what you think of the current version?

enitrat (changes requested):
> Reviewed 4 of 7 files at r2, 2 of 4 files at r3, 1 of 2 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: all files reviewed, 12 unresolved discussions (waiting on @TAdev0)

___
*[`src/ch11-07-inlining-in-cairo.md` line 16 at r1](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NuQhDLn8oHvqPonjvoi:-NuduXnQ80o3hT3jzHsp:b52izmv) ([raw file](https://github.com/cairo-book/cairo-book/blob/c4a94a6afd01f0468a05739a3c61d2de29d1b6c4/src/ch11-07-inlining-in-cairo.md#L16)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

perfect :)
___
*[`src/ch11-07-inlining-in-cairo.md` line 56 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudv24f3Fsnz--sFC_x:-Nudv24f3Fsnz--sFC_y:b3wgfw5) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L56)):*
> ```Markdown
> 
> ```rust,noplayground
> 00 function_call<user@main::main::not_inlined>() -> ([0])
> ```

can use a #include instead no ? in case we ever change this
___
*[`src/ch11-07-inlining-in-cairo.md` line 80 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudvJUu2lJ27sP66S9N:-NudvJUu2lJ27sP66S9O:b1xazoo) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L80)):*
> ```Markdown
> ```
> 
> The only difference is that the inlined code will store the `felt252_const` value in a variable with id `1`, because `[0]` is not available:
> ```

refers to a previous variable ?



_Code quote:_
```Markdown
`[0]` is not available
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 116 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudxcLaBqYsknn1JIGm:-NudxcLaBqYsknn1JIGn:btqdyn4) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L116)):*
> ```Markdown
> 11	ret
> ```
> ```

Add a link to cairovm.codes to follow along the step-by-step execution
___
*[`src/ch11-07-inlining-in-cairo.md` line 122 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudwkxeBxTsUURvdzXu:-NudwkxeBxTsUURvdzXv:undefined) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L122)):*
> ```Markdown
> 
> We can now decompose how these instructions are executed to understand what this code does:
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> ```





_Suggestion:_
```Markdown
increments the PC by 3
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 123 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudwoH72gKwNr6MVc91:-NudwoH72gKwNr6MVc92:b45h) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L123)):*
> ```Markdown
> We can now decompose how these instructions are executed to understand what this code does:
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> - `call rel 9` updates the PC to 9 and executes the instruction at this location.
> ```





_Suggestion:_
```Markdown
increments the PC by 9
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 123 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudwtOW6yac1zpMZ8XA:-NudwtOW6yac1zpMZ8XB:undefined) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L123)):*
> ```Markdown
> We can now decompose how these instructions are executed to understand what this code does:
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> - `call rel 9` updates the PC to 9 and executes the instruction at this location.
> ```





_Suggestion:_
```Markdown
and executes the instruction at this location, at line 9.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 124 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudxMVg77aUcz8ulFJu:-NudxMVg77aUcz8ulFJv:undefined) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L124)):*
> ```Markdown
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> - `call rel 9` updates the PC to 9 and executes the instruction at this location.
> - `[ap + 0] = 2, ap++`: `ap` stands for Allocation Pointer, which points to the first memory cell that has not been used by the program so far. This means we store the value `2` in `[ap - 1]`, as we applied `ap++` at the end of the line. Then, we go to the next line which is `ret`.
> ```





_Suggestion:_
```Markdown
This means we store the value `2` in the next free memory cell indicated by the current value of `ap`, after which we increment `ap` by 1. Then, we go to the next line which is `ret`.
```
___
*[`src/ch11-07-inlining-in-cairo.md` line 131 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudxq6h1WJMQW7SjNSJ:-Nudxq6h1WJMQW7SjNSK:bwnxdmx) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L131)):*
> ```Markdown
> - `ret`: last instruction executed as there is no more `call` instruction where to jump right after. This is the actual return instruction of the Cairo `main` function.
> 
> Note the pattern of a `call` instruction followed immediately by a `ret` instruction. This is a tail recursion, where the return values of the called function are forwarded.
> ```

I don't think this is useful here - do you know what it corresponds to?
___
*[`src/ch11-07-inlining-in-cairo.md` line 135 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudy0Xb5AyGt-wb0z4O:-Nudy0Xb5AyGt-wb0z4P:b97pzmm) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L135)):*
> ```Markdown
> To summary: 
> - `call rel 3` corresponds to the `main` function, which is obviously not inlined.
> - `call rel 9` triggers the call the `not_inlined` function, which returns `2` and stores it at the final location `[ap-3]`.
> ```

why is the execution not starting with `call rel 9`? (genuine question)
___
*[`src/ch11-07-inlining-in-cairo.md` line 188 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudyNRzBbn0OaLQJKVI:-NudyNRzBbn0OaLQJKVJ:b1k35nx) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L188)):*
> ```Markdown
> Inlining is a compiler optimization technique that can be very useful in various situations. Inlining a function allows to get rid of the overhead of calling a function with the `function_call` libfunc by injecting the Sierra code directly in the caller function's context, while potentially optimizing the Sierra code executed to reduce the number of steps. If used effectively, inlining can even reduce code length as shown in the previous example.
> 
> Nevertheless, applying the `inline` attribute to a function with a lot of code and few parameters might result in an increased code size, especially if the inlined function is used many times in the codebase. Use inlining only where it makes sense to use it.
> ```

Add a statement about why inlining should in most cases not be applied by the user itself: the compiler is supposed to handle this properly.

TAdev0 (commented):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: all files reviewed, 12 unresolved discussions (waiting on @enitrat)

___
*[`src/ch11-07-inlining-in-cairo.md` line 56 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudv24f3Fsnz--sFC_x:-NuecDCRClcl15KFE6qH:b-7a7dbi) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L56)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

can use a #include instead no ? in case we ever change this
</blockquote></details>

anchors don't work in .sierra files  ...
but i can put all this in cairo file with a TAG not to compile?
___
*[`src/ch11-07-inlining-in-cairo.md` line 80 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudvJUu2lJ27sP66S9N:-NuecgiwAQWLHhZ68r3J:bd82rd) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L80)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

refers to a previous variable ?
</blockquote></details>

yes thanks, previously assigned is even better i think in the context
___
*[`src/ch11-07-inlining-in-cairo.md` line 116 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudxcLaBqYsknn1JIGm:-NueePCK5oQdoAB0Daj8:b-896fix) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L116)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Add a link to cairovm.codes to follow along the step-by-step execution
</blockquote></details>

Done.
___
*[`src/ch11-07-inlining-in-cairo.md` line 122 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudwkxeBxTsUURvdzXu:-NuefCEB9EX4TgbrKZZR:b-896fix) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L122)):*
> ```Markdown
> 
> We can now decompose how these instructions are executed to understand what this code does:
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> ```

Done.
___
*[`src/ch11-07-inlining-in-cairo.md` line 123 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudwoH72gKwNr6MVc91:-NuefD4NAeCS8PYeGR0J:b-896fix) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L123)):*
> ```Markdown
> We can now decompose how these instructions are executed to understand what this code does:
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> - `call rel 9` updates the PC to 9 and executes the instruction at this location.
> ```

Done.
___
*[`src/ch11-07-inlining-in-cairo.md` line 123 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudwtOW6yac1zpMZ8XA:-NuefzfIBhq8u9oJmahh:b-896fix) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L123)):*
> ```Markdown
> We can now decompose how these instructions are executed to understand what this code does:
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> - `call rel 9` updates the PC to 9 and executes the instruction at this location.
> ```

Done.
___
*[`src/ch11-07-inlining-in-cairo.md` line 124 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudxMVg77aUcz8ulFJu:-NuegK6I6XLR9rY9j4ei:biuimdb) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L124)):*
> ```Markdown
> - `call rel 3`: this instruction updates the PC to 3 and executes the instruction at this location, which is `call rel 9`.
> - `call rel 9` updates the PC to 9 and executes the instruction at this location.
> - `[ap + 0] = 2, ap++`: `ap` stands for Allocation Pointer, which points to the first memory cell that has not been used by the program so far. This means we store the value `2` in `[ap - 1]`, as we applied `ap++` at the end of the line. Then, we go to the next line which is `ret`.
> ```

very nice!
___
*[`src/ch11-07-inlining-in-cairo.md` line 131 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudxq6h1WJMQW7SjNSJ:-Nueg0gqBPfhLbUpFZuX:bu8npxb) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L131)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I don't think this is useful here - do you know what it corresponds to?
</blockquote></details>

I think i understand it 60% :p
___
*[`src/ch11-07-inlining-in-cairo.md` line 135 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudy0Xb5AyGt-wb0z4O:-NuehBqrCaV9m9Ih-qfg:bxmbvab) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L135)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

why is the execution not starting with `call rel 9`? (genuine question)
</blockquote></details>

simply because call rel 3 is the instruction at PC = 1 ? 
___
*[`src/ch11-07-inlining-in-cairo.md` line 188 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-NudyNRzBbn0OaLQJKVI:-NuejdcgAWIRHrVMd4nz:b-896fix) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L188)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Add a statement about why inlining should in most cases not be applied by the user itself: the compiler is supposed to handle this properly.
</blockquote></details>

Done.

enitrat (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/711)* status: 15 of 16 files reviewed, 3 unresolved discussions (waiting on @TAdev0)

___
*[`src/ch11-07-inlining-in-cairo.md` line 56 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudv24f3Fsnz--sFC_x:-NuiS-y06PF6YFZ5gXac:b-8h2z9l) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L56)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

anchors don't work in .sierra files  ...
but i can put all this in cairo file with a TAG not to compile?
</blockquote></details>

ok let's keep this outside then
___
*[`src/ch11-07-inlining-in-cairo.md` line 135 at r8](https://reviewable.io/reviews/cairo-book/cairo-book/711#-Nudy0Xb5AyGt-wb0z4O:-NuiSEy_BeQg_aLLC_5K:b-nos1mi) ([raw file](https://github.com/cairo-book/cairo-book/blob/111162e19cd802bc61848888de615839559cc853/src/ch11-07-inlining-in-cairo.md#L135)):*
<details><summary><i>Previously, TAdev0 (Tristan) wrote…</i></summary><blockquote>

simply because call rel 3 is the instruction at PC = 1 ?
</blockquote></details>

yes but why isn't it starting with call rel 9 ? like the call rel 3 only does

call rel 3
ret

so nothing really... idk

PR #724, opened 2024-04-07
feat: appendix with common error messages

### LINKED ISSUE #716: feat: document common error messages
e.g.

 `Got an exception while executing a hint: Hint Error: Failed to deserialize param #1`
 
 -> means that the execution failed because an entrypoint was called without the expected arguments
 
 Can be done in Appendix

TAdev0:
Closes #716

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/724)


TAdev0:
@enitrat  ready for review.
Tell me if you think of any other common error. Variable previously moved and variable not dropped were the 2 errors i encountered when i got started

enitrat (changes requested):
> Reviewed 2 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/724)* status: 2 of 10 files reviewed, 3 unresolved discussions (waiting on @TAdev0)

___
*[`src/appendix-05-common-error-messages.md` line 7 at r2](https://reviewable.io/reviews/cairo-book/cairo-book/724#-Nv2ve3T18NubWw5QicV:-Nv2ve3T18NubWw5QicW:b-tg31zs) ([raw file](https://github.com/cairo-book/cairo-book/blob/a2a3c34791bcea43d39e57e19fca0c5e4aa1319a/src/appendix-05-common-error-messages.md#L7)):*
> ```Markdown
> - `Variable not dropped.`: this error message means that you are trying to make a variable with a type that do not implement the `Drop` trait go out of scope, withtout destroying it. Make sure that variables that need to be dropped at the end of the execution of a function implement the `Drop` trait or the `Destruct` trait. See [Ownership](ch04-01-what-is-ownership.md#destroying-values---example-with-feltdict) section.
> 
> - `Variable was previously moved.`: this error message means that you are trying to use a variable whose ownership has already been transferred to another function. When a variable doesn't implement the `Copy` trait, it is passed by value to functions, and ownership of the variable is transferred to the function. Such a variable cannot be used anymore in the current context after its ownership has been transferred. It is often useful to use the `clone` method in this situation.
> ```

not necessarily! if it needs to be mutated -> ref, otherwise, @ 



_Code quote:_
```Markdown
It is often useful to use the `clone` method in this situation.
```
___
*[`src/appendix-05-common-error-messages.md` line 9 at r2](https://reviewable.io/reviews/cairo-book/cairo-book/724#-Nv2vnuH5jGVk4XzWq0w:-Nv2vnuH5jGVk4XzWq0x:undefined) ([raw file](https://github.com/cairo-book/cairo-book/blob/a2a3c34791bcea43d39e57e19fca0c5e4aa1319a/src/appendix-05-common-error-messages.md#L9)):*
> ```Markdown
> - `Variable was previously moved.`: this error message means that you are trying to use a variable whose ownership has already been transferred to another function. When a variable doesn't implement the `Copy` trait, it is passed by value to functions, and ownership of the variable is transferred to the function. Such a variable cannot be used anymore in the current context after its ownership has been transferred. It is often useful to use the `clone` method in this situation.
> 
> - `error: Trait has no implementation in context: core::fmt::Display::<package_name::struct_name>`: this error message is encountered if you try to print an instance of a custom data type with `{}` placeholders in a `print!` or `println!` macro. To mitigate this issue, you need to either manually implement the `fmt` method of the `Display` trait for your type, or use the `Debug` trait by applying `derive(Debug)` to your type, allowing to print your instance by adding `:?` in `{}` placeholders.
> ```





_Suggestion:_
```Markdown
you need to either manually implement the `Display` trait for your type
```
___
*[`src/appendix-08-common-starknet-error-messages.md` line 1 at r2](https://reviewable.io/reviews/cairo-book/cairo-book/724#-Nv2vz31ERTOAmNWeb9H:-Nv2vz31ERTOAmNWeb9I:bowfs7s) ([raw file](https://github.com/cairo-book/cairo-book/blob/a2a3c34791bcea43d39e57e19fca0c5e4aa1319a/src/appendix-08-common-starknet-error-messages.md#L1)):*
> ```Markdown
> # Appendix A - Common Starknet Error Messages
> ```

I think fragmentation makes content retrieval harder. Perhaps it makes sense to keep everything grouped together, even if starknet / cairo - related ?

enitrat (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/724)* status: 2 of 10 files reviewed, 4 unresolved discussions (waiting on @TAdev0)

___
*[`src/appendix-08-common-starknet-error-messages.md` line 7 at r2](https://reviewable.io/reviews/cairo-book/cairo-book/724#-Nv2wEfl9g651V0jFMHP:-Nv2wEfl9g651V0jFMHQ:bj2hd9e) ([raw file](https://github.com/cairo-book/cairo-book/blob/a2a3c34791bcea43d39e57e19fca0c5e4aa1319a/src/appendix-08-common-starknet-error-messages.md#L7)):*
> ```Markdown
> ## General Error Messages
> 
> - `Got an exception while executing a hint: Hint Error: Failed to deserialize param #x.`: this error means that the execution failed because an entrypoint was called without the expected arguments. Make sure that the arguments you provide when calling an entrypoint are correct. There is a classic issue with `u256` variables, which are actually structs of 2 `u128`. Therefore, when calling a function that takes a `u256` as argument, you need to pass 2 values.
> ```

+ for example this is an error that comes from the Serde trait - we encounter it very often in the context of Starknet entrypoints, but in reality it's just a Serde issue

TAdev0:
@enitrat  ok Will do only one error section in Cairo appendix !

enitrat:
@TAdev0 Can you add visibility-related errors?

TAdev0:
@enitrat sure

enitrat:
@TAdev0 conflicts

PR #726, opened 2024-04-10
Update all titles + various improvements

TAdev0:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/726)

### DIFF
--- a/cairo-documentation-style-guide.md
+++ b/cairo-documentation-style-guide.md
@@ -39,7 +39,6 @@ The style guide is based on the best practices collected from the following book
 - [Terminology](#terminology)
   - [Recommended Terms](#recommended-terms)
 - [Admonitions](#admonitions)
-- [Appendix A Existing Style Guides](#appendix-a-existing-style-guides)
   - [Documentation](#documentation)
   - [Code](#code)
 
@@ -56,17 +55,16 @@ In heading titles, capitalize the first letter of every word **except for**:
 
 - Articles (a, an, the); unless an article is the first word.
 
-  > **Defining an Enum**<br>
-  > **Generic Types and Traits**
+  > **Defining an Enum**
 
 - Coordinating conjunctions (and, but, for, or, nor).
 
   > **Packages and Crates**
 
-- Prepositions of four letters or less; unless these prepositions are the first or last words.
-  - Prepositions of _five_ letters and above should be capitalized (Before, Through, Versus, Among, Under, Between, Without, etc.).
-
-  > **Peripherals as State Machines**
+- Prepositions of four letters or less should not be capitalized, unless these prepositions are the first or last words. Prepositions of five letters and above should be capitalized (Before, Through, Versus, Among, Under, Between, Without, etc.).
+  
+  > **Using Structs to Structure Related Data**
+  > **Components: Under the Hood**
 
 Do not capitalize names of functions, commands, packages, websites, etc.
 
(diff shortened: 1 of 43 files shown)


cairo-documentation-style-guide.md
```diff
@@ -56,17 +56,16 @@ In heading titles, capitalize the first letter of every word **except for**:
 
 - Articles (a, an, the); unless an article is the first word.
 
-  > **Defining an Enum**<br>
-  > **Generic Types and Traits**
```
enitrat, line 60: multiple examples is good

cairo-documentation-style-guide.md
```diff
@@ -56,17 +56,16 @@ In heading titles, capitalize the first letter of every word **except for**:
...
+  > **Defining an Enum**
 
 - Coordinating conjunctions (and, but, for, or, nor).
 
   > **Packages and Crates**
 
-- Prepositions of four letters or less; unless these prepositions are the first or last words.
-  - Prepositions of _five_ letters and above should be capitalized (Before, Through, Versus, Among, Under, Between, Without, etc.).
```
enitrat, line 67: you can keep emphasization on _five_

cairo-documentation-style-guide.md
```diff
@@ -126,13 +125,13 @@ See also, the [Using `monospace`](#using-monospace) section.
 
 The following types of lists are usually used in documentation:
 
-- **Bullet list** -- use it if the order or items is not important
-- **Numbered list** -- use it if the order of items is important, such as when describing a process
+- **Bullet list** -- use it if the order or items is not important.
```
enitrat, line 128: as per the rule

> If a bullet point is a full sentence, you can end it with a full stop.

> If a list has at least one full stop, end all other list items with a full stop.

Without full stop was the correct way

cairo-documentation-style-guide.md
```diff
@@ -227,17 +226,17 @@ This chapter lists the terms that have inconsistencies in spelling, usage, etc.
 If you spot other issues with terminology, please add the terms here in alphabetical order using the formatting as follows:
 
 - _Recommended term_
-  - Avoid: Add typical phrases in which this term is found
-  - Use: Add recommended phrases
-  - Note: Add more information if needed
+  - Avoid: Add typical phrases in which this term is found.
```
enitrat, line 229: Actually in most your corrections of bulleted lists, the full stop after each item is in most cases not justified, as you're making list of incomplete sentences (also known as fragment lists).

Full stops should only be kept for list of __complete sentences__.

e.g.:


The committee members decided on several actions:

- The secretary will respond to each recommendation.
- The secretary will allocate responses that need more work to members.
- Members will discuss the recommendations at the next meeting on 9 March.

Counterexample, a sentence that was turned into a fragmented list:

For example:

'a' is equivalent to 0x61
'b' is equivalent to 0x62
'c' is equivalent to 0x63
0x616263 is equivalent to 'abc'.

cairo-documentation-style-guide.md
TAdev0: @enitrat  the 2nd example highlights the case of coordinating conjunctions, not articles, thats why i removed it. Can add it back after `> **Packages and Crates**` on line 63 if you want

cairo-documentation-style-guide.md
TAdev0: ok

cairo-documentation-style-guide.md
TAdev0: ok, but ' use it if the order or items is not important. ' is not a full sentence ?

cairo-documentation-style-guide.md
enitrat: ah ok, yeah why not

TAdev0:
@enitrat  comments addressed, should be good

Issue #729, opened 2024-04-14
feat: Add Quizzes to all Chapters of the Book


enitrat:
As there have been a lot of requests to contribute lately and to avoid people squatting issues without working on them, please ensure that:
- ⚠️ You have read the entire Cairo Book ⚠️ 
- ⚠️ You have read and understood what makes a good quiz ⚠️ 
- ⚠️ You will be available to bring changes post-review ⚠️ 
- The chapter you want to work on is not being worked on (check the list)
- Given the simplicity of the task, if you want to add a quizz, you must open a pull request in the next 2 days - if you're not available, you can always come back a bit later!

Tasklist:

The Cairo Programming Language:
- [x] Chapter 1 #660 
- [x] Chapter 2 #680 
- [x] Chapter 3.1 Array  #744 
- [x] Chapter 3.2 Dict #755 
- [x] Chapter 4.1 Ownership #760 
- [x] Chapter 4.2 References and Snapshot
- [x] Chapter 5.1 Defining/instantiating structs #759 
- [x] Chapter 5.2 Example #782 
- [x] Chapter 5.3 Method syntax #783 
- [x] Chapter 6.1  enum #769 
- [x] Chapter 6.2 match #771 
- [x] Chapter 6.3 if let / while let #792 
- [x] Chapter 7 Managing packages / modules  #791 
- [x]  Chapter 8 - Generic Types and Traits #785 
- [x] Chapter 8.1  Generic Data Types
- [x] Chapter 8.2  Traits in Cairo
- [x] Chapter 9.1 Unrecoverable errors with panic #787 
- [x] Chapter 9.2 Recoverable errors with Result #775 
- [x] Chapter 10.1 How To Write Tests #779 
- [x] Chapter 10.2 Tests Organisation #788 

- [x] Chapter 11.1 Custom Data Structures #797 
- [x] Chapter 11.3 Smart Pointers
- [x] Chapter 11.4 Operators overloading #799 

Add more relevant quiz. Can take inspiration from the Rust book.

## How to write a good quiz

A good quiz realizes the following points:
- Asks about the concepts in the current chapter
- Encourages in-depth reflection on these concepts
- introduces subtleties to ensure that the concepts are correctly understood

### Example 1 - a bad quiz

>  What happens when you run this function?
```rust
fn main() -> u8 {
    let mut arr = ArrayTrait::<u8>::new();
    arr.append(100);
    arr.append(2);
    let x = *arr.at(2);
[...]

PR #732, opened 2024-04-17
Add more quiz

Zeegaths:
#729 Added three quizzes to chapter 3 - Arrays

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/732)


TAdev0:
Hi @Zeegaths ! 
Your PR is empty, can you check again your last commit and push it correctly?

Zeegaths:
@TAdev0  done. sorry for the oversight

TAdev0 (changes requested):

quizzes/ch03-01-arrays.toml
```diff
@@ -0,0 +1,78 @@
...
+
+
+
+
+
+[[questions]]
+id = "0xcccccc"
+type = "MultipleChoice"
+prompt.prompt = """
+What happens when you run this function?
```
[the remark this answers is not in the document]
Zeegaths: I've made the changes

enitrat:
@Zeegaths The goal of quizzes is to challenge the reader on its understanding of the material. We do not want to ask simple questions about the reading capacities of the reader, we want to challenge its comprehension. As such, we need to orient our questions in this purpose

What is useful:
- Asking questions about "deep" mechanisms: e.g., 
  - what would make a an array access panic - wrapped in a layer of obsfucation and with distractors, e.g. a function that would search for a value in up to `n` elements in an array, 
  - verifying that the immutability mechanism is understood properly 
  - verifying that the user understood what the array represents in memory: what happens if it's "owned" by two variables?

What is not useful:
- Asking questions about surface stuff (how to access an array's element - this is _trivial_). or, maximum __one__
- Asking questions not related to arrays directly (e.g. your two last question are not related to arrays)

Zeegaths:
Ok got it

quizzes/ch03-01-arrays.toml
```diff
@@ -2,43 +2,33 @@
...
+    a.append(3);
+
+    let first_value = a.pop_front().unwrap();
+
+    println!("The first value is {}", first_value);  
 }
 ```
 """
-answer.answer = "`arr.pop_front()`"
+answer.answer = "It panics"
```
enitrat: This is not true, your program doesn't even compile.

Zeegaths:
I've added quizzes targeting mutability and access. lmk if i should make more changes

enitrat:
@Zeegaths I advise you spend some more time studying and understanding how Cairo works, this will make it easier to propose quizzes that are challenging. With a little more experience you will be able to also identify what you have been struggling with and take inspiration from some questions.

quizzes/ch03-01-arrays.toml
Zeegaths: so it panics? or should i explicitly state it doesn't compile

Zeegaths:
ok

quizzes/ch03-01-arrays.toml
enitrat: No, the example is not correct for what you are trying to demonstrate

Zeegaths:
I think i had the wrong commit.., I have updated it, please check it out

Zeegaths:
Thank you for the clarification. I have made the fixes

enitrat:
https://github.com/cairo-book/cairo-book/pull/732#issuecomment-2061219239 Please take my comment into consideration, I even added specific examples of what you could do. What you're proposing for now is not of good quality

Zeegaths:
Thanks for the feedback.

Issue #733, opened 2024-04-17
Test running forever


gaetbout:
In https://book.cairo-lang.org/ch10-01-how-to-write-tests.html
If hitting Run this code on a test it runs forever:
![image](https://github.com/cairo-book/cairo-book/assets/16206518/d3ac6685-3cfa-426b-9090-6c7d0d58ab16)

It shouldn't run forever and either tell it isn't possible or run the tests

enitrat:
@cryptonerdcn is WASM-Cairo able to run cairo-tests?

cryptonerdcn:
@gaetbout @enitrat WASM-Cairo still can’t run test. Can you open an issue to request this feature at here https://github.com/cryptonerdcn/wasm-cairo/issues ?

PR #748, opened 2024-04-22
Feature/wasm cairo0.8.8

cryptonerdcn:
@enitrat @gaetbout 

スクリーンショット 2024-04-22 21 05 45

Add run tests feature to WASM-Cairo to solve these issues:

https://github.com/cairo-book/cairo-book/issues/733

and 

https://github.com/cryptonerdcn/wasm-cairo/issues/18

You can see more about WASM-Cairo 0.8.8&0.8.7 here:

https://github.com/cryptonerdcn/wasm-cairo/releases/tag/v0.8.8
and
https://github.com/cryptonerdcn/wasm-cairo/releases/tag/v0.8.7

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/748)

### DIFF
--- a/theme/book.js
+++ b/theme/book.js
@@ -115,22 +115,50 @@ function playground_text(playground, hidden = true) {
     let text = playground_text(code_block);
 
     result_block.innerText = "Running...";
-    window
-      .runFunc(text)
-      .then((data) => {
-        if (data.trim() === "") {
-          result_block.innerText = "No output";
-          result_block.classList.add("result-no-output");
-        } else {
-          result_block.innerText = data;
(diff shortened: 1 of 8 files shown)


enitrat (approved):

cryptonerdcn:
Because  `#![allow(unused)]` and `fn main()` are auto-added to the Cairo file which is for testing, I modified `run_cairo_code` and added a function to clean the Cairo code before passing the code to the test function.
This may not be a good solution(because it used regex and I did not test on all the `.cairo` files), so maybe we can update it later if we need.

PR #755, opened 2024-05-02
Add quiz for dictionaries chapter

Symmaque:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/755)


quizzes/ch03-02-dictionaries.toml
```diff
@@ -0,0 +1,68 @@
+[[questions]]
+type = "Tracing"
+prompt.program = """
+use core::dict::Felt252DictTrait;
```
TAdev0, line 4: > `use core::dict::Felt252DictTrait;`

This import can be removed

quizzes/ch03-02-dictionaries.toml
```diff
@@ -0,0 +1,68 @@
...
+    balances.insert('Alex', 100);
+    balances.insert('Maria', 200);
+    let john_balance = balances.get('John') + 100;
+    println!("John : {}", john_balance);
+}
+"""
+answer.doesCompile = true
+answer.stdout = "John : 100"
+context = """
+The code compiles and runs without panicking because the 'get' method returns 0 for a missing key.
```
TAdev0, line 16: for keys whose associated value is not initialized

quizzes/ch03-02-dictionaries.toml
```diff
@@ -0,0 +1,68 @@
...
+"""
+id = "8b538a1f-0cf9-4c57-8304-c4b07e0134dd"
+
+[[questions]]
+type = "MultipleChoice"
+prompt.prompt = """
+What will be the output of this code snippet?
+
+```
+use core::dict::Felt252DictTrait;
```
TAdev0, line 27: remove this import

quizzes/ch03-02-dictionaries.toml
```diff
@@ -0,0 +1,68 @@
...
+```
+"""
+prompt.distractors = [
+  "Alex : -50",
+  "Alex : 0",
+  "Alex : 100",
+]
+answer.answer = "There will be a runtime panic."
+context = """
+The value type of this dictionary is `u64`, which is an unsigned integer. So the variable `alex_balance` is a unsigned integer that can't be negative.
```
TAdev0, line 45: The type of elements contained in this dictionary

quizzes/ch03-02-dictionaries.toml
```diff
@@ -0,0 +1,68 @@
...
+The subtraction operation will cause a runtime panic.
+"""
+id = "af5bc548-fb84-487f-958e-0622d242dc6f"
+
+[[questions]]
+type = "Tracing"
+prompt.program = """
+use core::dict::Felt252DictTrait;
```
TAdev0, line 53: remove import

quizzes/ch03-02-dictionaries.toml
```diff
@@ -0,0 +1,68 @@
...
+fn main() {
+    let mut balances: Felt252Dict<i64> = Default::default();
```
enitrat, line 55: I'd rather use a Span as an example rather than an integer (as there's no real reason why i64 wouldn't be supported, it technically could as it fits in 1 single felt)

quizzes/ch03-02-dictionaries.toml
```diff
@@ -0,0 +1,68 @@
...
+    balances.insert('Alex', 100);
+    balances.insert('Maria', 200);
+    let alex_balance = balances.get('Alex') - 150;
+    println!("Alex : {}", alex_balance);
+}
+"""
+answer.doesCompile = false
+answer.lineNumber = 4
+context = """
+The value type of this dictionary is `i64`, which is not natively supported by the `Felt252Dict`. The code will not compile.
```
TAdev0, line 65: can you add precision about the need for the type to implement `zero_default` method of `Felt252DictValue<T>` trait, and the usage of `Nullable<T>` in such cases?

TAdev0 (changes requested):
Overall, looks good! nevertheless, all quizzes are related to the `get` method to retrieve values from a dict at some point. It also introduced types not supported natively which is nice.

Could you add 1 or 2 more ,  for example one about squashing  and one about entry/finalize?

quizzes/ch03-02-dictionaries.toml
enitrat: no required in the context here, the full answer is already in the chapter's content

enitrat (approved):

PR #760, opened 2024-05-10
feat: added quiz to part 1 of chapter 4

Ayoazeez26:


quizzes/ch04-01-what-is-ownership.toml
```diff
@@ -0,0 +1,35 @@
+[[questions]]
+type = "Tracing"
+prompt.program = """
```
enitrat, line 3: use descriptive struct, functions and variables names. 
- What is `Dec` ? What is `n9` ? Could `bar` do something, like print the value of x, and the quiz asks what's the stdout of the program?

you could typically use `Point {x, y}` and print the x coordinate.

quizzes/ch04-01-what-is-ownership.toml
```diff
@@ -0,0 +1,35 @@
...
+type = "MultipleChoice"
+prompt.prompt = "The reason why dictionary data types must be `squashed` when destructed is:"
+answer.answer = "so that sequence of access can be proven"
+prompt.distractors = [
+  "dictionaries can easily simulate the existence of mutable memory when there is none",
+  "cairo implements dictionarites as a list of entries in order to simulate mutability",
+]
+prompt.answerIndex = 0
+context = """
+Squashing dictionaries in Cairo improves efficiency by reducing memory usage and optimizing data access
```
enitrat, line 34: dictionaries in Cairo are types that must be "squashed" when destructed, so that the sequence of access can be proven. This has nothing to do with efficiency !

quizzes/ch04-01-what-is-ownership.toml
```diff
@@ -0,0 +1,35 @@
...
+"""
+answer.doesCompile = false
+answer.lineNumber = 2
+context = """
+The `Copy` Trait was not implemented on the Dec custom type
+"""
+
+[[questions]]
+type = "MultipleChoice"
+prompt.prompt = "The reason why dictionary data types must be `squashed` when destructed is:"
```
enitrat, line 26: A better prompt related to this topic would be something that challenges the comprehension of this mechanism, and the purpose of `Drop` globally.

For example (to improve):
What happens if a dictionary is not squashed inside a Cairo program ?
1. A malicious prover could inject values and the program would not be sound
2. The compiler will prevent this from happening with the `Destruct` mechanism


(both answers are valid - is it possible to have multiple valid answers? )

enitrat (changes requested):
Some improvements to be done, and new questions to write. We can play with the `@` `mut` `ref` modifiers, asking whether programs compile, and if so, what the output is.

enitrat:
ping @Ayoazeez26

Ayoazeez26:
Hi @enitrat I'll push a commit soon, just finishing up

quizzes/ch04-01-what-is-ownership.toml
```diff
@@ -1,35 +1,46 @@
 [[questions]]
 type = "Tracing"
 prompt.program = """
-struct Dec {
-    a: u32,
-    b: u32,
+const PI: u32 = 3.142;
```
enitrat, line 4: seeing this makes me believe that you haven't tried your program and don't really understand how Cairo works - you cannot have decimal numbers in Cairo

quizzes/ch04-01-what-is-ownership.toml
```diff
@@ -1,35 +1,46 @@
...
+fn main() {
+    let person_1 = Person { name: "Vicra", age: 12 }
+    println!("The name of the first user is {}")
+}
+"""
+answer.doesCompile = false
+answer.lineNumber = 2
 context = """
-Squashing dictionaries in Cairo improves efficiency by reducing memory usage and optimizing data access
+`Debug` and traits were not implemented on the Person struct, which causes the code block not to compile
```
enitrat, line 44: This is not related to Chapter 4! + you missed some words in the context...

enitrat (changes requested):

enitrat:
I am going to close this as I don't think this challenges well the concepts seen in this section, it only scratches the surface of the inner mechanism of ownership, there are no mentions of the concept of moving and dropping values and thus I see very little value in having these questions

PR #769, opened 2024-05-14
Add enums quiz

Symmaque:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/769)

### DIFF
--- a/quizzes/ch06-01-enums.toml
+++ b/quizzes/ch06-01-enums.toml
@@ -20,26 +20,27 @@ id = "8b786183-ef9f-43f8-89a6-0e9c0e87c576"
 [[questions]]
 type = "MultipleChoice"
 prompt.prompt = """
-What will be the output of this code snippet?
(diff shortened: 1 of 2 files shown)


quizzes/ch06-01-enums.toml
```diff
@@ -0,0 +1,45 @@
...
+}
+"""
+answer.doesCompile = false
+context = """
+Cairo does not have null pointers, so the null keyword does not exist. 
+An Option type should be used to represent the possibility of an object being null.
+"""
+id = "8b786183-ef9f-43f8-89a6-0e9c0e87c576"
+
+[[questions]]
```
enitrat, line 20: I wouldn't keep this one, `match` is introduced in the next section, the question itself is not that interesting imo. Perhaps replace it with a question about _when_ to use enums or structs

enitrat (approved):

PR #776, opened 2024-05-14
chore: improve links ch08 to ch11

TAdev0:


src/ch08-01-generic-data-types.md
```diff
@@ -20,7 +20,7 @@ The `largest_list` function compares two lists of the same type and returns the
 {{#rustdoc_include ../listings/ch08-generic-types-and-traits/no_listing_02_with_tdrop/src/lib.cairo}}
 ```
 
-The new `largest_list` function includes in its definition the requirement that whatever generic type is placed there, it must be droppable. The `main` function remains unchanged, the compiler is smart enough to deduce which concrete type is being used and if it implements the `Drop` trait.
+The new `largest_list` function includes in its definition the requirement that whatever generic type is placed there, it must be droppable. This is what we call _trait bounds_. The `main` function remains unchanged, the compiler is smart enough to deduce which concrete type is being used and if it implements the `Drop` trait.
```
enitrat, line 23: perhaps introducing trait bounds before the traits chapter was a mistake and we should've stopped at the non-compiling version, only to fix it in the next section?

src/ch10-01-how-to-write-tests.md
```diff
@@ -121,12 +119,10 @@ Remember in [Chapter 5](ch05-03-method-syntax.md), we used a `Rectangle` struct
 ```
 
 {{#label rectangle}}
-<span class="caption">Listing {{#ref rectangle}}: Using the `Rectangle` struct and its `can_hold` method from Chapter 5.</span>
+<span class="caption">Listing {{#ref rectangle}}: Using the `Rectangle` struct and its `can_hold` method from chapter 5</span>
```
enitrat, line 122:
```suggestion
<span class="caption">Listing {{#ref rectangle}}: Using the `Rectangle` struct and its `can_hold` method from Chapter 5</span>
```

src/ch10-01-how-to-write-tests.md
```diff
@@ -149,6 +145,9 @@ It does pass! Let’s add another test, this time asserting that a smaller recta
 {{#rustdoc_include ../listings/ch10-testing-cairo-programs/listing_10_03/src/lib.cairo:test2}}
 ```
 
+{{#label another-test}}
+<span class="caption">Listing {{#ref another-test}}: Adding another test in _lib.cairo_ that will pass</span>
```
enitrat, line 149: Not all code listings have to be captioned and numerotated, especially if they're not mentioned in the corpus.

src/ch10-01-how-to-write-tests.md
```diff
@@ -60,8 +60,6 @@ It’s possible to mark a test as ignored so it doesn’t run in a particular in
 
 Let’s start to customize the test to our own needs. First change the name of the `it_works` function to a different name, such as `exploration`, like so:
 
-<span class="filename">Filename: src/lib.cairo</span>
```
enitrat, line 63: Why deleting all these? Don't you think it's useful to follow along and not mix things up? or do you think we should just assume the code is located in `lib.cairo`?

src/ch10-01-how-to-write-tests.md
TAdev0: yeah sure i know, i added this one because in this chapter, each time the test file is modified with a new test added, there is a listing quote. otherwise, there is just the code. I added this one here for this reason

src/ch10-01-how-to-write-tests.md
TAdev0: same here, i left it where a new situation is introduced, removed it elswhere. If you look closely, i didnt remove all of them. Only where it made sense to do so

src/ch08-01-generic-data-types.md
TAdev0: we followed the flow of the rust book. I read it again right now, and i feel it is good like this. You kind of need to introduce bounds when talking about generic functions.

Because you need to introduce generic function, then generic traits i guess.
before for generic function, you quickly need to add trait bounds to your generic functions.

up to you, i dont know what is the best

enitrat (approved):

PR #777, opened 2024-05-14
feat: Make a distinction between `CairoRun` and `CairoRunGas` 

TAdev0:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/777)

### DIFF
--- a/cairo-listings/src/cmd.rs
+++ b/cairo-listings/src/cmd.rs
@@ -5,6 +5,7 @@ pub enum ScarbCmd {
     Format(),
     Build(),
     CairoRun(),
+    CairorunGas(),
     Test(),
 }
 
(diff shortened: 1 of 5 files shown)


TAdev0:
@enitrat  this is an idea.
It allows to run either `scarb cairo-run` or `scarb cairo-run --available-gas ..` , depending on the presence of a new tag called `GasConsumption`

I would still need to create the tag, and modify the appropriate listings

enitrat:
This is not the right approach to the problem. If there is a requirement for the runner to support a `--available-gas` argument, the correct approach would not be to have a command matching cairo-run specifically with gas, but to add support for command arguments in general - and this is already implemented

PR #782, opened 2024-05-15
feat: add quiz questions for example program using structs

No-bodyq:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/782)

### DIFF
--- a/quizzes/ch05-02-an-example-program-using-structs.toml
+++ b/quizzes/ch05-02-an-example-program-using-structs.toml
@@ -8,8 +8,7 @@ struct Rectangle {
 }
 
 fn perimeter(r: Rectangle) -> u64 {
-    let perimeter = 2 * (r.width + r.height);
-    perimeter
+    2 * (r.width + r.height)
 }
 
 fn main(){
@@ -22,6 +21,7 @@ fn main(){
 }
 """
 answer.doesCompile = false
+context = "The area function takes ownership of its argument rectangle, which doesn't implement `Copy`. Calling perimeter(rectangle1) therefore moves rectangle1, meaning it cannot be used on the next line."
 
 [[questions]]
 type = "Tracing"
--- a/quizzes/ch05-02-an-example-program-using-structs.toml
+++ b/quizzes/ch05-02-an-example-program-using-structs.toml
@@ -1,0 +1,22 @@
+[[questions]]
+type="Tracing"
+prompt.program = """
+#[derive(Drop)]
+struct Rectangle {
+    width: u64,
+    height: u64,
+}
+
+fn perimeter(r: Rectangle) -> u64 {
+    let perimeter = 2 * (r.width + r.height);
+    perimeter
+}
+
+fn main(){
+    let rectangle1 = Rectangle{
+        width: 20,
+        height: 10,
+    };
+    println!("{}", perimeter(rectangle1));
+}
+"""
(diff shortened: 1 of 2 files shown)


quizzes/ch05-02-an-example-program-using-structs.toml
```diff
@@ -0,0 +1,38 @@
...
+fn main(){
+    let rectangle1 = Rectangle{
+        width: 20,
+        height: 10,
+    };
+    println!("{}", perimeter(rectangle1));
+}
+"""
+answer.doesCompile = true
+answer.stdout = "60"
```
enitrat, line 24: it's interesting that you decided to not derive copy on Rectangle, so that `perimeter` takes ownership of it. I think it's more challenging and interesting to highlight this fact in the quiz

```suggestion
fn main(){
    let rectangle1 = Rectangle{
        width: 20,
        height: 10,
    };
    let p = perimeter(rectangle1);
    println!("2 * ({} + {}) = {}", rectangle1.width, rectangle1.height, p);
}
"""
answer.doesCompile = false
```


and that would indeed not compile!

enitrat (commented):
great ideas :)

quizzes/ch05-02-an-example-program-using-structs.toml
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Drop)]
+struct Rectangle {
+    width: u64,
+    height: u64,
+}
+
+fn perimeter(r: Rectangle) -> u64 {
+    let perimeter = 2 * (r.width + r.height);
+    perimeter
+}
```
enitrat, line 13:
```suggestion
fn perimeter(r: Rectangle) -> u64 {
    2 * (r.width + r.height)
}
```

quizzes/ch05-02-an-example-program-using-structs.toml
```diff
@@ -0,0 +1,38 @@
...
+        width: 20,
+        height: 10,
+    };
+    let p = perimeter(rectangle1);
+    println!("2 * ({} + {}) = {}", rectangle1.width, rectangle1.height, p);
+}
+"""
+answer.doesCompile = false
+
+[[questions]]
```
enitrat, line 26:
```suggestion
answer.doesCompile = false
context = "The area function takes ownership of its argument rectangle, which doesn't implement `Copy`. Calling perimeter(rectangle1) therefore moves rectangle1, meaning it cannot be used on the next line."

[[questions]]
```

quizzes/ch05-02-an-example-program-using-structs.toml
```diff
@@ -0,0 +1,38 @@
...
+struct Point { x: u128, y: u128, }
+
+fn main() {
+    let p = Point { x: 5, y: 10 };
+    let Point { z, w } = p;
+    println!("The values of z and w are: ({}, {})", z, w);
+}
+"""
+answer.doesCompile = false
+context = "Destructuring refers to the process of extracting individual fields from a struct and binding them to separate variables. When destructuring a struct, you can either use variables with the same names as the struct fields or explicitly bind the fields to variables with different names using the syntax `field_name: variable_name`."
```
enitrat, line 38:
```suggestion
context = "Destructuring refers to the process of extracting individual fields from a struct and binding them to separate variables. When destructuring a struct, you can either use variables with the same names as the struct fields or explicitly bind the fields to variables with different names using the syntax `field_name: variable_name`."

```

enitrat (approved):

PR #784, opened 2024-05-15
Add package management quizzes

Symmaque:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/784)


quizzes/ch07-03-paths-in-module-tree-1.toml
```diff
@@ -0,0 +1,53 @@
+[[questions]]
+type = "MultipleChoice"
+prompt.prompt = """
+What is the first word of an absolute path ?
+"""
+prompt.distractors = ["Folder name", "Module name", "Package name"]
+answer.answer = "Crate name"
```
enitrat, line 7: Actually, in Cairo that would be the Package name but as noted previously with a set theory POV, package = crate . But the difference between a package and a crate is very blurry in cairo... If we consider the following scarb.toml

```
[package]
name = "my_package"
```

then you would start absolute paths with `my_package`

quizzes/ch07-01-packages-crates.toml
```diff
@@ -0,0 +1,16 @@
+
```
enitrat, line 1: This is a slight deviation from rust, a Cairo package only has one crate (lib), unlike rust where you can have a binary crate and a lib crate

enitrat: It's not well defined yet, perhaps we should just say package == crate because a package is always one crate 🤔

enitrat: and thus update the context to say smth like "packages and crates are equivalent in cairo"

quizzes/ch07-04-bringing-paths-into-scope.toml
```diff
@@ -0,0 +1,47 @@
+[[questions]]
+type = "MultipleChoice"
+prompt.prompt = """
+Which of the following statements best describes the function of the `use` keyword?
+"""
+prompt.distractors = [
+  "`use` reduces the verbosity of referring to items in the used path",
```
enitrat, line 7: duplicate with answer, must have been something else initially?

enitrat (commented):
great :)

PR #803, opened 2024-05-21
add quiz for references and snapshots

Symmaque:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/803)

### DIFF
--- a/quizzes/ch04-02-references-and-snapshots.toml
+++ b/quizzes/ch04-02-references-and-snapshots.toml
@@ -45,32 +45,30 @@ You should write `let mut rect = Rectangle { height: 10, width: 20 };` instead.
 type = "Tracing"
 prompt.program = """
 fn sum(arr: @Array<u128>) -> u128 {
-  let mut span = arr.span();
-  let mut sum = 0;
-  while let Option::Some(x) = span.pop_front() {
-    sum += *x;
-  };
-  sum
+    let mut span = arr.span();
+    let mut sum = 0;
+    while let Option::Some(x) = span.pop_front() {
+        sum += *x;
+    };
+    sum
 }
 
-    arr.len()
-}
 fn main() {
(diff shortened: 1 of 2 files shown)


quizzes/ch04-02-references-and-snapshots.toml
```diff
@@ -0,0 +1,140 @@
...
+"""
+answer.doesCompile = false
+context = """
+The code does not compile because a `ref` can only be used on mutable variables.
+You should write `let mut rect = Rectangle { height: 10, width: 20 };` instead.
+"""
+[[questions]]
+type = "Tracing"
+prompt.program = """
+fn calculate_length(arr: @Array<u128>) -> usize {
```
enitrat, line 47: 1. calculate_length is just doing `arr.len()` perhaps something like `sum` that sums all array elements using a loop would be nice?

```suggestion
fn sum(arr: @Array<u128>) -> u128 {
  let mut span = arr.span();
  let mut sum = 0;
  while let Option::Some(x) = span.pop_front() {
    sum += *x;
  };
  sum
}

```

quizzes/ch04-02-references-and-snapshots.toml
```diff
@@ -0,0 +1,140 @@
...
+    let mut snap2 = snap;
+    arr1.append(4);
+    snap2 = @arr1;
+    println!("{}", calculate_length(snap2));
+}
+"""
+answer.doesCompile = true
+answer.stdout = "4"
+context = """
+First `snap2` is a snapshot of a memory location that contains the array `arr1`.
```
enitrat, line 62: Wouldn't it be more interesting to pass snap instead of snap2 (which corresponds to the latest state of `arr`)?

enitrat (commented):
small nit

enitrat (approved):
lgtm

PR #807, opened 2024-05-23
add traits quiz

Symmaque:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/807)

### DIFF
--- a/quizzes/ch08-02-traits.toml
+++ b/quizzes/ch08-02-traits.toml
@@ -1,0 +1,130 @@
+[[questions]]
+type = "Tracing"
+prompt.program = """
+pub trait MakeNoise<T> {
+    fn make_noise(self: @T) {
+        println!("noise");
+    }
+}
+
+pub mod dog {
+    use super::MakeNoise;
+
+    #[derive(Drop)]
+    pub struct Dog {}
+
+    impl Barking of MakeNoise<Dog> {
+        fn make_noise(self: @Dog) {
+            println!("bark");
+        }
+    }
+}
+
+pub mod cat {
+    use super::MakeNoise;
+
+    #[derive(Drop)]
+    pub struct Cat {}
+}
+
+fn main() {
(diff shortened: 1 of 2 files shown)


quizzes/ch08-02-traits.toml
```diff
@@ -0,0 +1,130 @@
...
+    let dog = dog::Dog {};
+    dog.make_noise();
+
+    let cat = cat::Cat {};
+    cat.make_noise();
+}
+"""
+answer.doesCompile = false
+context = """
+This is not possible for trait function to have a body in cairo.
```
enitrat, line 40: Actually I think this is coming in the next update, but for now it's indeed valid :)

quizzes/ch08-02-traits.toml
```diff
@@ -0,0 +1,130 @@
...
+    let my_shape = Shape{};
+    // the line goes here
+}
+```
+"""
+prompt.distractors = ["`draw(my_shape);`", "`draw(@my_shape);`"]
+answer.answer = "`my_shape.draw();`"
+context = """
+The `draw` method is a method implemented for the `Shape` type, so it should be called on the `my_shape` object.
+It is also possible to call the method on the snapshot of the object by using the `@` operator : `(@my_shape).draw();` for example.
```
enitrat, line 73: yes, it's actually strictly equivalent, basically the compiler transforms my_shape.draw() to (@my_shape).draw()

it's better to say that you can call it with `Drawable::draw(@my_shape)`

enitrat, line 73:
```suggestion
It is also possible to call the method on the snapshot of the object by using the `@` operator : `Drawable::draw(@my_shape)` for example.
```

enitrat (commented):
great :)

enitrat (approved):

Jemiiah:
I was still doing the work!!

PR #811, opened 2024-05-24
Updates for chapters 5 and 6: fixed grammar and wording, fix bug in `find_value_iterative`, simplify code

misicnenad:
- add missing `break` in `fn find_value_iterative` -> fixes a bug where the loop keeps iterating through the span, which would cause it to assign `result` to a later index.
Repro code:
```rust
fn main() {
    let mut sp: Span<felt252> = array![1, 2, 2].span();
    assert!(find_value_iterative(sp, 2) == Option::Some(1), "Error!");  // currently panics
}
```
- remove suffix `1` from names where it makes little sense (`width1 -> width`, `rectangle1->rectangle` etc.).
- format code.
- add missing backticks around code-related words.
- general fixes in text: wrong variable names, unclear description, improved wording, grammar.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/811)


src/ch05-03-method-syntax.md
```diff
@@ -84,17 +84,17 @@ Here, we expect that `rect1` can hold `rect2` but not `rect3`.
...
+Let's create the method `new` which creates a `Rectangle` from a `width` and a `height`, and a method `compare` which compares two `Rectangle` instances, and returns `true` if both rectangles have the same area and `false` otherwise.
 
 ```rust
 {{#include ../listings/ch05-using-structs-to-structure-related-data/no_listing_05_class_methods/src/lib.cairo:trait_impl}}
 
 {{#include ../listings/ch05-using-structs-to-structure-related-data/no_listing_05_class_methods/src/lib.cairo:main}}
 ```
 
-Note that the `compare` function could also be written as an _instance method_ with `self` as the first rectangle. In this case, instead of using the method with `Rectangle::compare(rect1, rect2)`, it is called with `rect1.compare(rect2)`.
+Note that the `compare` function could also be written as an _instance method_ with `self` as the first rectangle. In this case, instead of using the method with `RectangleTrait::compare(@rect1, @rect2)`, it is called with `rect1.compare(@rect2)`.
```
enitrat, line 97: > it is called with `rect1.compare(@rect2)`.

`rect1.compare(rect2)` should be fine

src/ch06-02-the-match-control-flow-construct.md
```diff
@@ -13,11 +13,11 @@ Speaking of coins, let’s use them as an example using `match`! We can write a
 {{#label match-enum}}
 <span class="caption">Listing {{#ref match-enum}}: An enum and a `match` expression that has the variants of the enum as its patterns</span>
 
-Let’s break down the `match` expression in the `value_in_cents` function. First, we list the `match` keyword followed by an expression, which in this case is the value `coin`. This seems very similar to a conditional expression used with if, but there’s a big difference: with if, the condition needs to evaluate to a Boolean value, but here it can be any type. The type of `coin` in this example is the `Coin` enum that we defined on the first line.
+Let’s break down the `match` expression in the `value_in_cents` function. First, we list the `match` keyword followed by an expression, which in this case is the value `coin`. This seems very similar to a conditional expression used with the `if` statement, but there’s a big difference: with `if`, the condition needs to evaluate to a `Boolean` value, but here it can be any type. The type of `coin` in this example is the `Coin` enum that we defined on the first line.
```
enitrat, line 16:
```suggestion
Let’s break down the `match` expression in the `value_in_cents` function. First, we list the `match` keyword followed by an expression, which in this case is the value `coin`. This seems very similar to a conditional expression used with the `if` statement, but there’s a big difference: with `if`, the condition needs to evaluate to a boolean value, but here it can be any type. The type of `coin` in this example is the `Coin` enum that we defined on the first line.
```

think this is better

src/ch06-02-the-match-control-flow-construct.md
```diff
@@ -13,11 +13,11 @@ Speaking of coins, let’s use them as an example using `match`! We can write a
...
 
 Next are the `match` arms. An arm has two parts: a pattern and some code. The first arm here has a pattern that is the value `Coin::Penny` and then the `=>` operator that separates the pattern and the code to run. The code in this case is just the value `1`. Each arm is separated from the next with a comma.
 
-When the `match` expression executes, it compares the resultant value against the pattern of each arm, in the order they are given. If a pattern matches the value, the code associated with that pattern is executed. If that pattern doesn’t match the value, execution continues to the next arm, much as in a coin-sorting machine. We can have as many arms as we need: in the above example, our `match` has four arms.
+When the `match` expression executes, it compares the resultant value against the pattern of each arm in the order they are given. If a pattern matches the value, the code associated with that pattern is executed. If the pattern doesn’t match, execution continues to the next arm, much like in a coin-sorting machine. We can have as many arms as we need; in the above example, our `match` has four arms.
```
enitrat, line 20:
```suggestion
When the `match` expression executes, it compares the resultant value against the pattern of each arm, in the order they are given. If a pattern matches the value, the code associated with that pattern is executed. If that pattern doesn’t match, execution continues to the next arm, much like in a coin-sorting machine. We can have as many arms as we need; in the above example, our `match` has four arms.
```

> first comma
I think it's nice to have the sentence pause here to emphasize the second part of the sentence

> the -> that

that clarifies we're still talking about the previous pattern, while `the` is more generic

not strongly opiniated about the other changes but I think the original paragraph was fine tbh

src/ch06-02-the-match-control-flow-construct.md
```diff
@@ -52,15 +52,15 @@ Because `state` is an `UsState` enum which implements the `Debug` trait, we can
...
 
 [debug trait]: ./appendix-03-derivable-traits.html#debug-for-printing-and-debugging
 
 ## Matching with `Option<T>`
 
 In the previous section, we wanted to get the inner `T` value out of the `Some` case when using `Option<T>`; we can also handle `Option<T>` using `match`, as we did with the `Coin` enum! Instead of comparing coins, we’ll compare the variants of `Option<T>`, but the way the `match` expression works remains the same.
 
-Let’s say we want to write a function that takes an `Option<u8>` and, if there’s a value inside, adds `1` to that value. If there isn’t a value inside, the function should return the `None` value and not attempt to perform any operations.
+Let’s say we want to write a function that takes an `Option<u8>` and, if there’s a value inside, adds `1` to that value. If there is no value inside, the function should return `None` and not attempt to perform any more operations.
```
enitrat, line 63:
```suggestion
Let’s say we want to write a function that takes an `Option<u8>` and, if there’s a value inside, adds `1` to that value. If there is no value inside, the function should return the `None` value and not attempt to perform any operations.
```

I would keep the last sentence

src/ch06-03-concise-control-flow-with-if-let-and-while-let.md
```diff
@@ -71,4 +70,4 @@ The `while let` syntax is similar to the `if let` syntax, but it allows you to l
 {{#rustdoc_include ../listings/ch06-enums-and-pattern-matching/no_listing_18_while_let/src/lib.cairo}}
 ```
 
-Using `while let` provides a more concise and idiomatic way of writing this loop compared to a traditional `while` loop with explicit pattern matching or handling of the `Option` type. However, as with `if let`, you lose the exhaustive checking that a `match` expression provides, so you need to be careful to handle any remaining cases outside the while let loop if necessary.
+Using `while let` provides a more concise and idiomatic way of writing this loop compared to a traditional `while` loop with explicit pattern matching or handling of the `Option` type. However, as with `if let`, you lose the exhaustive checking that the `match` expression provides, so you need to be careful to handle any remaining cases outside the `while let` loop if necessary.
```
enitrat, line 73:
```suggestion
Using `while let` provides a more concise and idiomatic way of writing this loop compared to a traditional `while` loop with explicit pattern matching or handling of the `Option` type. However, as with `if let`, you lose the exhaustive checking that a `match` expression provides, so you need to be careful to handle any remaining cases outside the `while let` loop if necessary.
```

> the -> a

we're talking about match in general, not a specific instance

enitrat (commented):
thank you very much for the careful review!

src/ch05-03-method-syntax.md
misicnenad: You're right. Checked it out locally, and the method could be implemented in a way that doesn't require snapshoting.

Just a heads up that I updated the text before this: [it is called with -> it would be called with](https://github.com/cairo-book/cairo-book/pull/811/commits/84b5179fc407ab43576af8577581b71d314ee71d)

src/ch06-02-the-match-control-flow-construct.md
misicnenad: I found that the changed version had a more natural "flow", but will revert if you think it was fine.

src/ch06-03-concise-control-flow-with-if-let-and-while-let.md
misicnenad: makes sense

misicnenad:
> thank you very much for the careful review!

thank you for maintaining this wonderful resource for Cairo learning!

src/ch06-03-concise-control-flow-with-if-let-and-while-let.md
```diff
@@ -39,16 +39,15 @@ between `match` and `if let` depends on what you’re doing in your particular
 situation and whether gaining conciseness is an appropriate trade-off for
 losing exhaustive checking.
 
-In other words, you can think of `if let` as syntax sugar for a `match` that
+In other words, you can think of `if let` as syntactic sugar for a `match` that
 runs code when the value matches one pattern and then ignores all other values.
 
-We can include an `else` with an `if let`. The block of code that goes with the
-`else` is the same as the block of code that would go with the `_` case in the
-`match` expression that is equivalent to the `if let` and `else`. Recall the
```
misicnenad, line 47: The ``that is equivalent to the `if let` and `else` `` part is just a bit redundant and unclear imo.

enitrat (approved):
💯

PR #819, opened 2024-05-27
Associated functions

Symmaque:
add clarification and don'ts for methods
add example of implementation of PartialEq trait in appendix

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/819)


listings/ch05-using-structs-to-structure-related-data/no_listing_05_class_methods/src/lib.cairo
```diff
@@ -29,7 +35,7 @@ fn main() {
     let rect1 = RectangleTrait::new(30, 50);
     let rect2 = RectangleTrait::new(10, 40);
 
-    println!("Are rect1 and rect2 equals ? {}", RectangleTrait::compare(@rect1, @rect2));
+    println!("The average Rectangle of {} and {} is {}", @rect1, @rect2, RectangleTrait::avg(@rect1, @rect2));
```
enitrat, line 38: do you need the Display trait specificallly? Otherwise, you can use the debug trait, like this, and derive it on the Rectangle type

```suggestion
    println!("The average Rectangle of {:?} and {:?} is {}", @rect1, @rect2, RectangleTrait::avg(@rect1, @rect2));
```

src/appendix-03-derivable-traits.md
```diff
@@ -57,12 +57,18 @@ The `PartialEq` trait allows for comparison between instances of a type for equa
 
 When `PartialEq` is derived on structs, two instances are equal only if all fields are equal, and the instances are not equal if any fields are not equal. When derived on enums, each variant is equal to itself and not equal to the other variants.
 
+Moreover you can implement your own `eq` (equal) and `ne` (not equal) functions suiting your `struct`. Here is an exemple where we consider that two rectangles are equal if they have the same area:
```
enitrat, line 60:
```suggestion
Moreover, you can write your own implemention of the `PartialEq` trait for your type, if you don't want to or can't derive it. Here is an example in which we consider that two rectangles are equal if they have the same area:
```

src/appendix-03-derivable-traits.md
```diff
@@ -57,12 +57,18 @@ The `PartialEq` trait allows for comparison between instances of a type for equa
...
+{{#include ../listings/appendix/listing_04_implpartialeq/src/lib.cairo}}
+```
+
 The `PartialEq` trait is required, for example, with the use of the `assert_eq!` macro in tests, which needs to be able to compare two instances of a type for equality.
 
 Here is an example:
 
 ```rust
-{{#include ../listings/appendix/listing_04_partialeq/src/lib.cairo}}
+{{#include ../listings/appendix/listing_05_partialeq/src/lib.cairo}}
```
enitrat, line 71: for next time: no need to rename the listings in the right order, just add a new one

src/ch05-03-method-syntax.md
```diff
@@ -1,10 +1,50 @@
-# Method Syntax
+# Associated functions
+
+Functions defined within an `impl` block are called _associated functions_ because they have to be associated with one and only one type.
```
enitrat, line 3: I don't think this is the right approach, not all function inside an `impl` are associated functions, and technically you could write an impl for multiple types.

It's tricky because in Cairo we only have **impls of traits** and not **impl of types** (we have both in rust). So while in Rust, all functions under `impl MyStruct {}` would be associated functions, it's not really the case in Cairo. To bypass that we often use the pattern

```rust
trait MyTypeTrait {
  fn new() -> MyType;
  fn foo(self: MyType)
  fn bar(a: MyType, b: MyType);
}

```

In that case, we're just introducing a useless trait to make something similar to `impl MyType`, hence we often abstract it with the `#[generate_trait]` attribute.

I would keep this part as it was before, perhaps move it to the end of the page, where we can add guidelines and do's / don't regarding trait usage

src/ch05-03-method-syntax.md
```diff
@@ -80,21 +120,21 @@ Let’s practice using methods by implementing another method on the `Rectangle`
 
 Here, we expect that `rect1` can hold `rect2` but not `rect3`.
 
-## Methods Without `self` Parameter
+## Associated functions without `self` Parameter / Class methods
```
enitrat, line 123:
```suggestion
## Associated functions
```

I think shorter is good enough

src/ch05-03-method-syntax.md
```diff
@@ -80,21 +120,21 @@ Let’s practice using methods by implementing another method on the `Rectangle`
...
 
-In Cairo, we can also define a method which doesn't act on a specific instance (so, without any `self` parameter) but which still manipulates the related type. This is what we call _class methods_ in Object-Oriented programming. As these methods are not called from an instance, we don't use them with the `<instance_name>.<method_name>` syntax but with the `<Trait_or_Impl_name>::<method_name>` syntax as you will see in the next example.
+In Cairo, we can also define a associated function which doesn't act on a specific instance (so, without any `self` parameter) but which still manipulates the related type. This is what we call _class methods_ in Object-Oriented programming. As these methods are not called from an instance, we don't use them with the `<instance_name>.<method_name>` syntax but with the `<Trait_or_Impl_name>::<method_name>` syntax as you will see in the next example.
```
enitrat, line 125: Actually, the part that you wrote above can be used here as a primer / introductory note to this section, where we define what we call an associated function and how it would be materialized in Cairo, and then keep  the existing text

src/ch05-03-method-syntax.md
```diff
@@ -1,12 +1,13 @@
-# Method Syntax
+# Associated functions
 
-_Methods_ are similar to functions: we declare them with the `fn` keyword and a
-name, they can have parameters and a return value, and they contain some code
-that’s run when the method is called from somewhere else.
+We call _associated functions_ functions that are associated with a `Type`. There are 2 ways for a function to be associated with a `Type`:
```
enitrat, line 3:
```suggestion
We call _associated functions_ functions that are defined inside an `impl` block and are associated with a specific type. There are 2 ways for a function to be associated with a type `T`:
```

src/ch05-03-method-syntax.md
```diff
@@ -1,12 +1,13 @@
...
 
-Unlike functions, methods are defined within the context of a type, either with their first parameter `self`
-which represents the instance of the type the method is being called on (also called _instance methods_),
-or by using this type for their parameters and/or return value (also called _class methods_ in Object-Oriented programming).
+- By having its first parameter called `self` and of this `Type`.
```
enitrat, line 5:
```suggestion
- By having its first parameter called `self` and of this type `T`.
```

src/ch05-03-method-syntax.md
```diff
@@ -1,12 +1,13 @@
...
+  For instance : `area(self :@Rectangle) -> u64` is an associated function to the type `Rectangle`.
```
enitrat, line 6:
```suggestion
  For instance : `fn area(self :@Rectangle) -> u64` is a function associated to the type `Rectangle`.
```

src/ch05-03-method-syntax.md
```diff
@@ -1,12 +1,13 @@
...
+  This kind of associated function is called a **method**.
```
enitrat, line 7:
```suggestion
We call these funtions **methods**, and their first parameter is always `self`, which represents the instance of the type the method is being called on..
```

src/ch05-03-method-syntax.md
```diff
@@ -1,12 +1,13 @@
...
+- By returning a instance of type `Type` or having at least one parameter of type `Type`.
```
enitrat, line 8:
```suggestion
- By returning an instance of this type `T`, or having at least one parameter of type `T`.
```

src/ch05-03-method-syntax.md
```diff
@@ -1,12 +1,13 @@
...
+  For instance : `avg(lhs :@Rectangle, rhs: @Rectangle) -> Rectangle` is an associated function to the type `Rectangle`.
```
enitrat, line 9:
```suggestion
  For instance, `fn avg(lhs :@Rectangle, rhs: @Rectangle) -> Rectangle` is function associated to the type `Rectangle`.
```

src/ch05-03-method-syntax.md
```diff
@@ -1,12 +1,13 @@
...
+  This kind of associated function is not a **method** because it hasn't `self` as first parameter.
```
enitrat, line 10:
```suggestion
  This kind of associated functions don’t have `self` as their first parameter (and thus are not methods) because they don’t need an instance of the type to work with. 
```

src/ch05-03-method-syntax.md
```diff
@@ -25,7 +26,7 @@ Then, we implement this trait in `RectangleImpl` with the `impl` keyword. In the
 {{#include ../listings/ch05-using-structs-to-structure-related-data/no_listing_01_define_methods/src/lib.cairo:trait_implementation}}
 ```
 
-Finally, we call this `area` method on the `Rectangle` instance `rect1` using the `<instance_name>.<method_name>` syntax. The instance `rect1` will be passed to the `area` method as the `self` parameter.
+Finally, we call this `area` method on the `Rectangle` instance `rect1` using the `<instance_name>.<method_name>` syntax. The instance `rect1` will be passed to the `area` method as the `self` parameter. Methods in Cairo are equivalent to _instance methods_ in object-oriented programming.
```
enitrat, line 29:
```suggestion
Finally, we call this `area` method on the `Rectangle` instance `rect1` using the `<instance_name>.<method_name>` syntax. The instance `rect1` will be passed to the `area` method as the `self` parameter.
```

src/ch05-03-method-syntax.md
```diff
@@ -80,21 +81,21 @@ Let’s practice using methods by implementing another method on the `Rectangle`
 
 Here, we expect that `rect1` can hold `rect2` but not `rect3`.
 
-## Methods Without `self` Parameter
+## Associated functions without `self`
 
-In Cairo, we can also define a method which doesn't act on a specific instance (so, without any `self` parameter) but which still manipulates the related type. This is what we call _class methods_ in Object-Oriented programming. As these methods are not called from an instance, we don't use them with the `<instance_name>.<method_name>` syntax but with the `<Trait_or_Impl_name>::<method_name>` syntax as you will see in the next example.
+In Cairo, we can also define a function which doesn't act on a specific instance (so, without any `self` parameter) but which still manipulates the related type and thus is an _associated function_. This is what we call _class methods_ in Object-Oriented programming. As these functions are not called from an instance, we don't use them with the `<instance_name>.<method_name>` syntax but with the `<Trait_or_Impl_name>::<method_name>` syntax as you will see in the next example.
```
enitrat, line 86:
```suggestion
In Cairo, we can also define a function which doesn't act on a specific instance (so, without any `self` parameter) but which still manipulates the related type. This is what we call _class methods_ in Object-Oriented programming. As these functions are not called from an instance, we don't use them with the `<instance_name>.<method_name>` syntax but with the `<Trait_or_Impl_name>::<method_name>` syntax as you will see in the next example.
```

src/ch05-03-method-syntax.md
```diff
@@ -108,13 +109,52 @@ the following code is equivalent to the code shown in the _Methods with several
 There’s no strong reason to separate these methods into multiple `trait` and `impl`
 blocks here, but this is valid syntax.
 
+## Guidelines about associated functions
+
+_associated functions_ defined within an `impl` block should all be associated with the same type.
```
enitrat, line 114:
```suggestion
_Associated functions_ defined within an `impl` block should all be associated with the same type.
```

enitrat (changes requested):

enitrat (approved):

PR #821, opened 2024-05-28
feat: missing derivable traits

### LINKED ISSUE #818: fix: add missing derivable traits
I've noticed there are 2 traits missing in the [Derivable traits](https://book.cairo-lang.org/appendix-03-derivable-traits.html) section:

* the `Default` trait
* the `starknet::Store` trait - I know the section is pure Cairo and the trait itself is Starknet specific, but I think it should be included there as the section is a handy reference of all the derivable traits (and let's be honest, Cairo is 95% Starknet so far)

milancermak:
Resolves #818.

Also contains a one character typo fix 😇

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/821)


enitrat (approved):
great :)

Issue #763, opened 2024-05-13
bug: Changing language is not working


davmart2:
When trying to change the language, an error appears.

[Document not found (404)](https://book.cairo-lang.org/#document-not-found-404)
This URL is invalid, sorry. Please use the navigation bar or search to continue.

TAdev0:
Indeed, it doest work. @enitrat  might i be related to the recent upgrade of version of all packages in the cargo.toml file?

Dprof-in-tech:
Hello @enitrat  i would love to work on this issue

enitrat:
sure go ahead

Dprof-in-tech:
Hello guys. So the error I can see is that a particular file
Can't be found but the file is available in the specified directory 

Could be a build issue. Still trying to figure it out

Dprof-in-tech:
oh yeah i'm still working on it chief. Have an idea I want to try out and if it doesn't work, I would make sure to raise a question here

Dprof-in-tech:
hi @TAdev0 i've tried changing the file path even duplicating it into the directory where it is needed but its absolutely to being read. This could be reassigned to someone else if there's anyone willing to pick it up, meanwhile I would keep trying to fix it and would send in a pr if I can

Thanks

TAdev0:
This issue is free to grab :)

addegbenga:
Can I work on this

TAdev0:
sure @addegbenga  go ahead :)

As you can see on the website:

https://book.cairo-lang.org/

trying to change the langage from english to any other triggers a [Document not found (404) error

CristopherAguilar10:
hello @TAdev0, @davmart2  , I wanted to make my first issue and this seems quite interesting to do, I think I have a possible solution and as there are no updates I would like to work on it.

PR #840, opened 2024-06-22
Language change issue 

addegbenga:
This PR fixes an issue where attempting to change the language from English to any other triggers a "Document not found (404)" error [#763](https://github.com/cairo-book/cairo-book/issues/763)

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/840)

### DIFF
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -7,7 +7,7 @@ name: Deploy mdBook site to Pages
 on:
   push:
     branches:
-      - tl
+      - main
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -1,4 +1,4 @@
-# Sample workflow for building and deploying a mdBook site to GitHub Pages
+# Sample workflow for building and deploying an mdBook site to GitHub Pages
 #
 # To get started with mdBook see: https://rust-lang.github.io/mdBook/index.html
 #
@@ -7,7 +7,7 @@ name: Deploy mdBook site to Pages
 on:
   push:
     branches:
-      - main
+      - tl
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


TAdev0:
thanks for the quick PR @addegbenga  . Were you able to test these changes?

addegbenga:
> thanks for the quick PR @addegbenga . Were you able to test these changes?

Yeah sure I did a quick deployment to make sure it works.

TAdev0:
so the missing`/html` was the key! thanks!

TAdev0 (approved):

.github/workflows/mdbook.yml
```diff
@@ -44,23 +44,28 @@ jobs:
       - name: Install mdbook
         uses: ./.github/workflows/install-mdbook
 
-      - name: Build book (English)
-        run: mdbook build -d book
+      - name: Build all Books
```
enitrat, line 47: Thank you, that looks good (moving book/lang/html to book/html/lang, now that we have book/html and book/markdown directories, this was required)

Can you just keep the file as it was before (One step is Build book, the other step is Build all translations) ? There should only be one line changed which is `mv book/$po_lang/html book/html/$po_lang`

addegbenga: Alright will do that

TAdev0 (approved):

TAdev0:
thanks @addegbenga !

TAdev0:
@addegbenga  i just checked and it seems only some chapters or pages are indeed translated, do you have the same issue on your side?

addegbenga:
> @addegbenga i just checked and it seems only some chapters or pages are indeed translated, do you have the same issue on your side?

Yes, I'm seeing the same issue on my end too. I'm not sure why it's happening, but I can do a quick check. It doesn't seem related to the build, so I can try to resolve it in a separate PR.

TAdev0:
would be great, thanks ! it seems the same .md files are not translated for all langages

addegbenga:
> would be great, thanks ! it seems the same .md files are not translated for all languages

I conducted a quick check last night and found issues with some of the po.xx files:
1. Missing translations.
2. Syntax errors in some po.xx files, causing certain markdown text to not be translated properly.

For example, I tested the es.po file and observed these issues in the attached image. As a result, some pages are translated while others are not.

**Tips for Fixing the Issues**

1. Review and Correct: Carefully review the po.xx files to fix all errors and add any missing translations.
2. Automate Formatting: Implement a script to format po.xx files automatically in the CI pipeline.
3. Use Translation Tools: Advise contributors to use tools like Poedit to review translations to catch errors early(I think this is already in readme).
4. Write a Guide: Create a proper guide on testing translations locally before making a pull request.

I think these tips can be worded better and  be opened for contribution to the community.


Screenshot 2024-06-23 at 10 15 32

TAdev0:
thanks for investigating @addegbenga 

I noticed its always the same chapters that are not translated for all languages, do you know why? I mean, the creation of new files for new langages is automatic right?

addegbenga:
> thanks for investigating @addegbenga
> 
> I noticed its always the same chapters that are not translated for all languages, do you know why? I mean, the creation of new files for new langages is automatic right?

There are two main reasons why this happen:

1. The sections in po.xx file for that chapter are not properly translated.
2. The translation for some sections in that chapter is missing.

Yes, file creation for new languages is automatic when you build.

I can't confirm that this issue happens for the same chapters in all languages. from my observation, when I switch to different languages such as Spanish (es) and French (fr), I notice a mix of translated and untranslated sections.
You can check Array chapter and switch between es and fr to see):

addegbenga:
In the first image you see that the second paragraph was not translated. 

I used Poedit to fix this, resulting in the corrected version in the second screenshot on my local machine. 
Thats mostly the issue we are facing for all the chapters.

Screenshot 2024-06-23 at 11 53 36
Screenshot 2024-06-23 at 11 53 55

TAdev0:
@addegbenga  how can we solve this so that just running the tool will work ? any idea why the translation tool itself doesnt work properly?

PR #841, opened 2024-06-24
feat: Bitwise Operators (OR operator)

TeddyNotBear:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/841)


TAdev0 (dismissed):
thanks @TeddyNotBear  size is a bit different, and i think currently all pgns in this chapter are aligned and have the same width. Could you just make it the same width?

TAdev0 (approved):
lgtm! thanks @TeddyNotBear

PR #842, opened 2024-06-24
bugfix: bit mistake in 'optimizing storage costs' chapter - combine.png image.

remybar:
The 8th bit in the result of OR operator in the `combine.png` file is not correct (should be 1 instead of 0).


TAdev0:
hi @remybar  , #841  just solved this error

remybar:
> hi @remybar , #841 just solved this error

Hi Tristan ! Ok, no problem, I close this PR ;-)

Issue #463, opened 2023-12-05
Advanced Cairo: gas optimisation for break conditions of loops


feltroidprime:
The example shown in https://book.cairo-lang.org/ch02-05-control-flow.html?highlight=loop#repetition-with-loops

currently demonstrates 

```
fn main() {
    let mut i: usize = 0;
    loop {
        if i > 10 {
            break;
        }
        'again'.print();
        i += 1;
    }
}
```

as a main example

This doesn't show best practice which is to avoid at all cost comparison operators to break out of loops. 

It is much cheaper to compare numbers `i == 11` rather than doing `i>10`, as internally ">" or "<" does range checks and calls the relevant libfunc, whereas numbers equality costs 2 steps, and zero checks `==0` or `!=0` cost 1 step. 
The best case, if possible, is to start from the latest index (10), decrement i, and check i==0. 
```

fn main() {
    let mut i: usize = 10;
    loop {
        if i == 0 {
            'again'.print();
            break;
        }
        'again'.print();
        i -= 1;
    }
}
```

enitrat:
lol, i just noticed this is an infinite loop.

Indeed it's better to use strict equality checks - provided that i is increased by incrementation of ones.

feltroidprime:
> lol, i just noticed this is an infinite loop.

copy pasted the wrong example. Edited.

enitrat:
@feltroidprime do you want to write a section about gas golfing in Cairo (1)?

enitrat:
>  whereas numbers equality costs 2 steps, and zero checks ==0 or !=0 cost 1 step.

Honestly, given all the overhead that the sierra abstractions imposes, I don't even think it's worth mentioning apart from in deep step optimisation chapters

enitrat:
Will be for chapter advanced cairo, probably gas optimizations

TAdev0:
closing as we will not add a dedicated chapter for gas optimisation. I'll push a PR to add in control flow section informations about avoiding range check builtin usage with equality check

PR #846, opened 2024-06-25
feat: break and continue conditions

TAdev0:
@enitrat  i think my example is not ideal because i need to add a sub operation in the printing for each iteration to print the same result...

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/846)


src/ch02-05-control-flow.md
```diff
@@ -149,6 +149,14 @@ Let's add a `continue` statement to our loop to skip the `println!` statement wh
 
 Executing this program will not print the value of `i` when it is equal to `5`.
 
+> When defining `break` or `continue` conditions in loops, it is better design to avoid comparison operators. Indeed, it is much cheaper to check if `i == 11` instead of `i > 10`, because comparison operators involve additional Cairo steps related to `range_check` Cairo-VM builtin usage.
```
TAdev0, line 152: i know its brutal in chapter 2 to talk about range check builtin. but later we can add a reference here for the cairo vm book section about builtins. Maybe i should add right now a ''future'' reference?

enitrat (commented):
I'm not sure what was the intention but this is already covered in the control flow chapter and I don't see a value added

TAdev0:
@enitrat  where is it covered?

it aims to resolve this issue : 

https://github.com/cairo-book/cairo-book/issues/463

just added a short section explaining why its better design to compare equality and not comparison for loops

TAdev0:
the value added is that by reading this, people might start using == for looping break and continue, thus saving useless computation

enitrat:
Will close this as I don't want to bloat the book with small nits about gas consumption. Eventually it will be meaningless

TAdev0:
makes sense, now i finally agree with you

PR #850, opened 2024-06-27
feat: add section about defining custom type casts with Into and TryInto traits

### LINKED ISSUE #844: feat: add section about defining custom type casts with `Into` and `TryInto` traits
Maybe add this to [Advanced Features](https://book.cairo-lang.org/ch11-00-advanced-features.html#advanced-features) section @enitrat @TAdev0 ?

misicnenad:
Closes #844

### DIFF
--- a/src/ch11-09-conversion.md
+++ b/src/ch11-09-conversion.md
@@ -25,3 +25,13 @@ We can do similarly for defining a conversion for our own type.
 ```rust
 {{#include ../listings/ch11-advanced-features/no_listing_15_into/src/lib.cairo}}
 ```
+
+## TryInto
+
+Similar to `Into`, `TryInto` is a generic trait for converting between types. Unlike `Into`, the `TryInto` trait is used for fallible conversions, and as such, returns [Option][option].
+
+[option]: ./ch06-01-enums.md#the-option-enum-and-its-advantages
+
+```rust
+{{#include ../listings/ch11-advanced-features/no_listing_16_tryinto/src/lib.cairo}}
+```
--- a/src/ch11-09-conversion.md
+++ b/src/ch11-09-conversion.md
@@ -1,0 +1,27 @@
+# Conversion
+
+Primitive types can be converted to each other through [casting][casting].
+
+Cairo addresses conversion between custom types (i.e., struct and enum) by the use of `Into` and `TryInto` traits. For converting custom types to string (`ByteArray`), refer to the ["Printing" chapter][printing].
+
+[casting]: ./ch02-02-data-types.md#type-casting
+[printing]: ./ch11-08-printing.md#printing-custom-data-types
+
+## Into
+
+The `Into` trait allows for a type to define how to convert itself into another type, hence providing a very simple mechanism for converting between several types. There are numerous implementations of this trait within the standard library for conversion of primitive and common types.
+
+For example we can easily convert a `felt252` into a `u256`
+
+```rust,noplayground
+    let my_felt: felt252 = 100;
+    let my_u256: u256 = my_felt.into();
+```
+
+Using the `Into` trait will typically require specification of the type to convert into as the compiler is unable to determine this most of the time. However this is a small trade-off considering we get the functionality for free.
+
+We can do similarly for defining a conversion for our own type.
+
+```rust
+{{#include ../listings/ch11-advanced-features/no_listing_15_into/src/lib.cairo}}
+```
--- a/listings/ch11-advanced-features/no_listing_15_into/Scarb.toml
+++ b/listings/ch11-advanced-features/no_listing_15_into/Scarb.toml
@@ -1,0 +1,9 @@
+[package]
+name = "no_listing_15_into"
+version = "0.1.0"
+edition = "2023_11"
+
+# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest
+
+[dependencies]
+# foo = { path = "vendor/foo" }
--- a/listings/ch11-advanced-features/no_listing_15_into/src/lib.cairo
+++ b/listings/ch11-advanced-features/no_listing_15_into/src/lib.cairo
@@ -1,0 +1,19 @@
+use core::traits::Into;
+
+#[derive(Drop, Debug)]
+struct Number {
+    value: u32,
+}
+
+impl U32IntoNumber of Into<u32, Number> {
+    fn into(self: u32) -> Number {
+        Number { value: self }
+    }
+}
+
+fn main() {
+    let int: u32 = 5;
+    // Compiler will complain if you remove the type annotation
+    let num: Number = int.into();
+    println!("My number is {:?}", num);
+}
--- a/listings/ch11-advanced-features/no_listing_16_tryinto/Scarb.toml
+++ b/listings/ch11-advanced-features/no_listing_16_tryinto/Scarb.toml
@@ -1,0 +1,9 @@
+[package]
+name = "no_listing_16_tryinto"
+version = "0.1.0"
+edition = "2023_11"
+
+# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest
+
+[dependencies]
+# foo = { path = "vendor/foo" }
--- a/listings/ch11-advanced-features/no_listing_16_tryinto/src/lib.cairo
+++ b/listings/ch11-advanced-features/no_listing_16_tryinto/src/lib.cairo
@@ -1,0 +1,26 @@
+use core::traits::TryInto;
+
+#[derive(Drop, Debug, PartialEq)]
+struct EvenNumber {
+    value: u32
+}
+
+impl TryU32IntoEvenNumber of TryInto<u32, EvenNumber> {
+    fn try_into(self: u32) -> Option<EvenNumber> {
+        if self % 2 == 0 {
+            Option::Some(EvenNumber { value: self })
+        } else {
+            Option::None
+        }
+    }
+}
+
+fn main() {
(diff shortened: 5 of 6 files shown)


src/ch11-09-conversion.md
```diff
@@ -0,0 +1,37 @@
+# Conversion
```
enitrat, line 1:
```suggestion
# Type Conversions
```

also rename the md file and edit SUMMARY.md

src/ch11-09-conversion.md
```diff
@@ -0,0 +1,37 @@
...
+
+Primitive types can be converted to each other through [casting][casting].
```
enitrat, line 3: !! I just realise that this is wrong, it's not a _cast_, it's an explicit _conversion_ ! The wrong terminology was used.

primitive types don't have a special treatment

src/ch11-09-conversion.md
```diff
@@ -0,0 +1,37 @@
...
+
+Cairo addresses conversion between custom types (i.e., struct and enum) by the use of `Into` and `TryInto` traits. For converting custom types to string (`ByteArray`), refer to the ["Printing" chapter][printing].
```
enitrat, line 5:
```suggestion
Cairo addresses conversion between types by the use of `Into` and `TryInto` traits. 
```

thus > all types

and it's not areally a conversion to ByteArray: it's just a _formatting_ to a bytearray

enitrat (changes requested):
Thank you! just a few remarks (on stuff also unrelated to your PR - but great if you can address them here)

src/ch11-09-conversion.md
misicnenad: What do you think about moving the whole `Data Types > Type Conversion` section into this new one?

Since primitive types don't have a special treatment, it makes more sense to group everything related to type conversion.

misicnenad: # Type Conversion

Cairo addresses conversion between types by using the `try_into` and `into` methods provided by the `TryInto` and `Into` traits from the core library. There are numerous implementations of these traits within the standard library for conversion of scalar types.

## Into

The `Into` trait allows for a type to define how to convert itself into another type. It can be used for type conversion when success is guaranteed, such as when the source type is smaller than the destination type. 

To perform the conversion, call `var.into()` on the source value to convert it to another type. The new variable's type must be explicitly defined, as demonstrated in the example below.

```rust,noplayground
    let my_u8: u8 = 10;
    let my_u16: u16 = my_u8.into();
    let my_u32: u32 = my_u16.into();
    let my_u64: u64 = my_u32.into();
    let my_u128: u128 = my_u64.into();

    let my_felt252 = 10;
    // As a felt252 is smaller than a u256, we can use the into() method
    let my_u256: u256 = my_felt252.into();
    let my_other_felt252: felt252 = my_u8.into();
    let my_third_felt252: felt252 = my_u16.into();
```

Defining a conversion for a custom type using the `Into` trait will typically require specification of the type to convert into, as the compiler is unable to determine this most of the time. However this is a small trade-off considering we get the functionality for free.

```rust
{{#include ../listings/ch11-advanced-features/no_listing_15_into/src/lib.cairo}}
```

## TryInto

Similar to `Into`, `TryInto` is a generic trait for converting between types. Unlike `Into`, the `TryInto` trait is used for fallible conversions, and as such, returns [Option<T>][option]. An example of a fallible conversion is when the target type might not fit the source value.

[option]: ./ch06-01-enums.md#the-option-enum-and-its-advantages

Also similar to `Into` is the process to perform the conversion; just call `var.try_into()` on the source value to convert it to another type. The new variable's type also must be explicitly defined, as demonstrated in the example below.

```rust,noplayground
    let my_u256: u256 = 10;
    // Since a u256 might not fit in a felt252, we need to unwrap the Option<T> type
    let my_felt252: felt252 = my_u256.try_into().unwrap();
    let my_u128: u128 = my_felt252.try_into().unwrap();
    let my_u64: u64 = my_u128.try_into().unwrap();
    let my_u32: u32 = my_u64.try_into().unwrap();
    let my_u16: u16 = my_u32.try_into().unwrap();
    let my_u8: u8 = my_u16.try_into().unwrap();
```

Below is an example of implementing the `TryInto` trait for a custom type.

```rust
{{#include ../listings/ch11-advanced-features/no_listing_16_tryinto/src/lib.cairo}}
```

enitrat: looks good

enitrat: actually the opposite: move everything related to type conversions to Data Types > Type Conversions

misicnenad: Makes sense

misicnenad: Moving this to Data Types, I had a couple of notes:
- The beginning of this chapter states `"This section covers two subsets of data types: scalars and compounds."`, and although it has a subheader for "Scalar Types", it doesn't have the same one for "Compound Types"; it just mentions in the "The Tuple Type" that `"A _tuple_ is a general way of grouping together a number of values with a
variety of types into one compound type"`. On the other hand, "The Rust Book" includes both [tuples and arrays under a separate section "Compound Types"](https://doc.rust-lang.org/book/ch03-02-data-types.html#compound-types). Would it make sense for us to do the same (move arrays from "Common Collections" to "Data Types")?
If nothing else, we could at least have the "Compound Types" subtitle above tuples. Or clearly indicate that tuples are _the only_ compound type (if that's the case). Or we could _not_ mention "compound" types at all.
- Describing type conversion before describing tuples may make it seem like tuples are not "convertable". We should maybe move the "Type Conversion" section to the end of the "Data Types" chapter.

enitrat: @misicnenad This is something that can be done post 2.7.0 release, where we will introduce the notion of const arrays

src/ch02-02-data-types.md
```diff
@@ -207,4 +193,63 @@ Its size is always zero, and it is guaranteed to not exist in the compiled code.
 
 You might be wondering why you would even need a unit type? In Cairo, everything is an expression, and an expression that returns nothing actually returns `()` implicitly.
 
-{{#quiz ../quizzes/ch02-02-data-types.toml}}
\ No newline at end of file
+{{#quiz ../quizzes/ch02-02-data-types.toml}}
+
+## Type Conversion
+
+Cairo addresses conversion between types by using the `try_into` and `into` methods provided by the `TryInto` and `Into` traits from the core library. There are numerous implementations of these traits within the standard library for conversion of scalar types, and they can be implemented for custom types as well.
```
enitrat, line 200:
```suggestion
Cairo addresses conversion between types by using the `try_into` and `into` methods provided by the `TryInto` and `Into` traits from the core library. There are numerous implementations of these traits within the standard library for conversion between types, and they can be implemented for custom types as well.
```

src/ch02-02-data-types.md
```diff
@@ -207,4 +193,63 @@ Its size is always zero, and it is guaranteed to not exist in the compiled code.
...
+
+### Into
+
+The `Into` trait allows for a type to define how to convert itself into another type. It can be used for type conversion when success is guaranteed, such as when the source type is smaller than the destination type. 
+
+To perform the conversion, call `var.into()` on the source value to convert it to another type. The new variable's type must be explicitly defined, as demonstrated in the example below.
+
+```rust,noplayground
```
enitrat, line 208: Can you make it a `.cairo` file, as with other code snippets?

src/ch02-02-data-types.md
```diff
@@ -207,4 +193,63 @@ Its size is always zero, and it is guaranteed to not exist in the compiled code.
...
+### TryInto
+
+Similar to `Into`, `TryInto` is a generic trait for converting between types. Unlike `Into`, the `TryInto` trait is used for fallible conversions, and as such, returns [Option\<T\>][option]. An example of a fallible conversion is when the target type might not fit the source value.
+
+[option]: ./ch06-01-enums.md#the-option-enum-and-its-advantages
+
+Also similar to `Into` is the process to perform the conversion; just call `var.try_into()` on the source value to convert it to another type. The new variable's type also must be explicitly defined, as demonstrated in the example below.
+
+```rust,noplayground
+    let my_u256: u256 = 10;
```
enitrat, line 237: same

src/ch02-02-data-types.md
```diff
@@ -207,4 +193,63 @@ Its size is always zero, and it is guaranteed to not exist in the compiled code.
...
+    let my_u64: u64 = my_u128.try_into().unwrap();
+    let my_u32: u32 = my_u64.try_into().unwrap();
+    let my_u16: u16 = my_u32.try_into().unwrap();
+    let my_u8: u8 = my_u16.try_into().unwrap();
+
+    let my_large_u16: u16 = 2048;
+    let my_large_u8: u8 = my_large_u16.try_into().unwrap(); // panics with 'Option::unwrap failed.'
+```
+
+Below is an example of implementing the `TryInto` trait for a custom type.
```
enitrat, line 251: Both example 39 and  40 would probably be better placed at the end of https://book.cairo-lang.org/ch05-02-an-example-program-using-structs.html in a new subsection "Conversions of Custom Types" where you could TryInto Rectangle -> Square (fails if width != height)  and Into Square -> Rectangle (never fails). wdyt?

enitrat (commented):
Thank you for the reflections on how to make things clearer :)

src/ch02-02-data-types.md
misicnenad: Sounds good. Will add a link here in Data Types stating something like `For custom type conversion, see the "Conversions of Custom Types" section`.

enitrat (approved):

PR #856, opened 2024-07-02
docs: ch02-05-control-flow.md

kkawula:
This is not real information, currently 'break' isn't needed for compilation

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/856)


enitrat (commented):
Indeed, it looks like it's no longer required! 

Can you modify `no_listing_32_infinite_loop` and remove 
```
        if (false) {
            break;
        }
```

ty!

TAdev0 (approved):

PR #861, opened 2024-07-10
Remove the text about the need to specify array type to use subscr. operator for arrays

misicnenad:
Not sure if this was a limitation at one point in time, but today this is not a requirement.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/861)


misicnenad:
> I think we can completely `no_listing_05_array_at_2` and its paragraph no?

It would be nice to have an example of using `arr[index]` to access an array item.

Maybe merge `no_listing_05_array_at_2` with `no_listing_05_array_at`?
```rust
fn main() {
    let mut a = ArrayTrait::new();
    a.append(0);
    a.append(1);

    let item_using_at = *a.at(0);
    let item_using_subscripting = a[1];
    
    // maybe even add `println` here?
}
```

enitrat:
yep let's do this

misicnenad:
> yep let's do this

@enitrat done

TAdev0 (approved):

enitrat (approved):

PR #870, opened 2024-07-18
add default implementation part

### LINKED ISSUE #865: feat: Default implementations
Default trait implementation is now available, similar to Rust's https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations

Symmaque:
Closes #865 

This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/870)


src/ch08-02-traits-in-cairo.md
```diff
@@ -87,6 +87,56 @@ This code prints the following:
 
 Other crates that depend on the _aggregator_ crate can also bring the `Summary` trait into scope to implement `Summary` on their own types.
 
+## Default Implementations
+
+Sometimes it’s useful to have default behavior for some or all of the methods in a trait instead of requiring implementations for all methods on every type. Then, as we implement the trait on a particular type, we can keep or override each method’s default behavior.
+
+Below we specify a default string for the `summarize` method of the `Summary` trait instead of only defining the method signature, as we did previously.
```
enitrat, line 94:
```suggestion
In Listing {{#ref default_impl}} we specify a default string for the `summarize` method of the `Summary` trait instead of only defining the method signature, as we did in Listing {{#ref trait_signature}}.
```

Can you also add referenced captions for other listings that are being referenced in other paragraphs ?

src/ch08-02-traits-in-cairo.md
```diff
@@ -87,6 +87,56 @@ This code prints the following:
...
+
+<span class="caption">Filename: src/lib.cairo</span>
+
+```rust
+{{#rustdoc_include ../listings/ch08-generic-types-and-traits/no_listing_19_def_impl/src/lib.cairo:trait}}
+```
+
+<span class="caption">Defining a `Summary` trait with a default implementation of the `summarize` method</span>
```
enitrat, line 102:
```suggestion

{{#label default_impl}}
<span class="caption"> Listing {{#ref default_impl}}: Defining a `Summary` trait with a default implementation of the `summarize` method</span>
```

src/ch08-02-traits-in-cairo.md
```diff
@@ -87,6 +87,56 @@ This code prints the following:
...
+{{#rustdoc_include ../listings/ch08-generic-types-and-traits/no_listing_19_def_impl/src/lib.cairo:main}}
+```
+
+This code prints `New article available! (Read more...)`.
+
+Creating a default implementation doesn’t require us to change anything about the previous implementation of `Summary` on `Tweet`. The reason is that the syntax for overriding a default implementation is the same as the syntax for implementing a trait method that doesn’t have a default implementation.
+
+<!-- NOT AVAILABLE IN CAIRO FOR NOW -->
+
+<!-- Default implementations can call other methods in the same trait, even if those other methods don’t have a default implementation. In this way, a trait can provide a lot of useful functionality and only require implementors to specify a small part of it. For example, we could define the `Summary` trait to have a `summarize_author` method whose implementation is required, and then define a `summarize` method that has a default implementation that calls the `summarize_author` method:
```
enitrat, line 118: Possible using `Self::foo(self)` e.g.

```rust
pub trait Summary<T> {
    fn summarize_author(self: @T) -> ByteArray;

    fn summarize(
        self: @T
    ) -> ByteArray {
        format!("(Read more from {}...)", Self::summarize_author(self))
    }
}
```

enitrat (approved):

Issue #553, opened 2024-02-03
Advanced Cairo: loops are compiled to recursion


enitrat:

Enenche23:
i am interested in this, would like to handle it

ebukizy1:
will love to work on this. can i be assign to do this.

TAdev0:
@ebukizy1 thanks for your interest. Do you know how to tackle this issue ? It requires advanced knowledge about sierra / casm inner working.

Dprof-in-tech:
hi @enitrat I would love to be assigned to this task. I'm a fullstack Js dev learning cairo and I know I can create magic with this.

TAdev0:
hi @Dprof-in-tech  thanks for your interest! A few people already commented, do you have knowledge about Sierra and Casm ? can you shortly describ how you would tackle this issue?

enitrat:
hey @FriendlyLifeguard it's an old issue i'm not sure what this should look like now, you can give a try but I would probably need to add more details

TAdev0:
hey @FriendlyLifeguard  any update or draft PR ?

okhaimie-dev:
@TAdev0 can you assign this to me?

I'd love to explore it

enitrat:
@FriendlyLifeguard is on it, i've been discussing via dm

FriendlyLifeguard:
Hello, I will be pushing my drafts this week!

TAdev0:
@enitrat  issue already addressed, closing it

PR #871, opened 2024-07-18
add Advanced Cairo: Loops and idiomatic looping

FriendlyLifeguard:
Completed issue: #553 

Please let me know of any feedback in regards to explanations, code examples and the flow of each section!

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/871)


src/ch11-09-loops-and-idiomatic-looping.md
```diff
@@ -0,0 +1,137 @@
...
+{{#include ../listings/ch11-advanced-features/listing_06_loop/src/examples/simple_ownership_example.cairo}}
+```
+
+This code will output `Failed to infer` error due to the variable `x` being out of scope. 
+
+The issue arises from Cairo's unique ownership which states that once a variable goes out of scope, it is destroyed. (see [Ownership][ownership]) Therefore, the variable that you are trying to manipulate inside `loop` must be declared outside of the `loop` block. 
+
+[ownership]: ./ch04-01-what-is-ownership.md
+
+A way to mitigate this issue is by declaring a variable that will equal to the output of the `loop` block.
```
enitrat, line 95: this is a bit out of topic - it has nothing to do with loops specifically, and has more to do with variable visibility (as you noted)

src/ch11-09-loops-and-idiomatic-looping.md
```diff
@@ -0,0 +1,137 @@
...
+
+On here, we are trying to match a specific index in the `pizza_ingredients` array and in particular, we are iterating through the array until we find the `mushroom` variable then append the variable to the new `result` array. 
+
+```rust,noplayground
+{{#include ../listings/ch11-advanced-features/listing_06_loop/src/examples/ownership_example_complex.cairo}}
+```
+
+This code will output compile-time error with `variable was previously used here` message . Due to the way ownership works in Cairo, once a variable is moved, its original reference is destroyed and a new variable is created under the new scope. In this case, the `result` array is moved under `Option::Some` scope  and the compiler will not allow the variable to be used again under a new scope under `Option::None` condition.
+
+To mitigate the issue, we can snapshot the array (see [Snapshots][Snapshots]) to create an immutable reference of the array instead of taking ownership of the array.
```
enitrat, line 111: I didn't understand correctly what you're trying to explain here:

I think what you meant is that doing `println!` fails because the array `result` is _moved_. inside the loop, and thus once we're outside of the loop scope, we have not brought back ownerhsip of the `result` variable to the main context.

Because loops compile to recursion (see 1.), this is equivalent to doing

```rust
let result = array![]
rec(result);

fn rec(result: Array<felt252>{}
```

thus, if you want to keep accessing the `result` variable outside of the loop, you need to give back ownership. In a recursive function, that would be with `ref` (or an explicit return type), in a loop, you just give it back:

```
let result = array![];

// 1
fn rec(ref result){}
//
fn rec(result) -> Array<felt252> {return result;}
// 3
let result = loop{ ... };
```

src/ch11-09-loops-and-idiomatic-looping.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+This code will output compile-time error with `variable was previously used here` message . Due to the way ownership works in Cairo, once a variable is moved, its original reference is destroyed and a new variable is created under the new scope. In this case, the `result` array is moved under `Option::Some` scope  and the compiler will not allow the variable to be used again under a new scope under `Option::None` condition.
+
+To mitigate the issue, we can snapshot the array (see [Snapshots][Snapshots]) to create an immutable reference of the array instead of taking ownership of the array. 
+
+[Snapshots]: ./ch04-02-references-and-snapshots.md#snapshots
+
+
+## Comparing Gas Costs of Accessing Arrays with Loop
```
enitrat, line 116: Well, this entire section doesn't really make sense, you're comparing a O(n) operation (looping through all the elements of the array) vs a O(1) operation (accessing a single element at a known index)

You're comparing apples to peaches :/

enitrat (commented):
Thanks! I see you've explored Sierra and tried to get a good understanding of how things work, that's super cool :) Things I would do to improve:

- completely remove last section
- rework section 1, remove the sierra code snippets (not informative), and rework the explanation so that the main conclusion is:
- Loops internally compile to recursive functions
- Syntax <insert program using a loop> will compile exactly like <insert program with recursive function>

Section 2: keep the second ownership issue you mentioned about the "variable moved", justify why this error appears related to your findings in section 1, using the elements I gave you in comments

src/ch11-09-loops-and-idiomatic-looping.md
FriendlyLifeguard: For this one, you mentioned: "Also explains the differences in gas when looping over arrays with pop_front and match vs. just accessing arr[i]" when you first proposed the issue.

This section does seem unrelated to the overall section and yes I agree with your point. My original intent was to purely compare gas costs between the two methods. 

I will go ahead and remove the section.

enitrat: I see, the original idea was to compare a

loop { let x = match arr. pop_front... 

vs

let mut i = 0
loop { let x = arr[i]...}

but ultimately it's perhaps not very important

FriendlyLifeguard: Ah i see, thank you for clarifying.

listings/ch11-advanced-features/listing_06_loop/src/examples/ownership_example_complex.cairo
```diff
@@ -4,11 +4,34 @@ fn main() {
 
     let mushroom: felt252 = 7887337214842990445;
 
+<<<<<<< HEAD
```
enitrat, line 7: merge conflict

FriendlyLifeguard: oh haha, I just pushed a fix!

enitrat:
hey, after re-reading in depth, I really don't think that the concepts explained on the ownership are really related to loops themselves; you're just describing basic ownership rules.

I haven't been able to repro an example where loops were wrongly capturing ownership, so the most likely explanation is that it was a bug fixed since this issue was opened 6 months ago. 

For the recursion / loop explanation we can keep it!

enitrat (approved):

FriendlyLifeguard:
Understood. Thank you for the clarification!

PR #880, opened 2024-07-21
docs: Update ch14-01-contract-storage.md

### LINKED ISSUE #877: docs: update explanation on legacy map entries access
see change from release 2.7: https://github.com/cairo-book/cairo-book/pull/876/files#diff-f047a06659d1cf05314b1480b35ce835a22563e431b8a8609a57459e92992196R67

Adapt the explanation in the contract storage page to match these.
Also, calling `storage_var.address()` is no longer possible - we can however call `storage_var.address` which returns a `felt252` - adapt accordingly

blocksorcerer:
Closes #877

Updates explanation for chapter 14-01 in accordance to version 2.7.0

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/880)

### DIFF
--- a/src/ch14-01-contract-storage.md
+++ b/src/ch14-01-contract-storage.md
@@ -24,7 +24,7 @@ The address of a storage variable is computed as follows:
   
 - If the variable is a [mapping][storage mappings] with a key `k`, the address of the value at key `k` is `h(sn_keccak(variable_name),k)`, where ℎ is the Pedersen hash and the final value is taken modulo \\( {2^{251}} - 256\\). If the key is composed of more than one `felt252`, the address of the value will be `h(...h(h(sn_keccak(variable_name),k_1),k_2),...,k_n)`, with `k_1,...,k_n` being all `felt252` that constitute the key. In the case of mappings to complex values (e.g., tuples or structs), then this complex value lies in a continuous segment starting from the address calculated with the previous formula. Note that 256 field elements is the current limitation on the maximal size of a complex storage value.
 
-You can access the address of a storage variable by calling the `address` function on the variable, which returns a `StorageBaseAddress` value.
+You can access the base address of a storage variable by calling the `address` attribute on the variable, which returns a `felt252` value.
 
 ```rust, noplayground
 {{#rustdoc_include ../listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo:owner_address}}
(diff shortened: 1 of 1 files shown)


src/ch14-01-contract-storage.md
```diff
@@ -55,7 +55,7 @@ To write a new value to the storage slot of a storage variable, we call the `wri
 {{#rustdoc_include ../listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo:write_owner}}
 ```
 
-In this second example, we need to pass `user` and `name` as arguments, because `names` is a mapping, with `user` as key and `name` as value.
+In this second example, we need to pass `user` and `name` as arguments because `names` is a mapping, with `user` as the key and `name` as the value. To write the value, we call the `write` function on the entry obtained from `self.names.entry(user)`.
```
enitrat, line 58:
```suggestion
In this second example, we're working with the `names` mapping, where `user` serves as the key and `name` as the value. First, we need to retrieve the storage entry path corresponding to the `user` key, just as we did when reading the value. Once we have this entry path, we can then call the `write` function on it, passing `name` as the value to be stored.
```

other suggestion, I think it's a but more explicit, as the proposed changes didn't really take into consideration the two-step process entry -> action

src/ch14-01-contract-storage.md
```diff
@@ -43,7 +43,7 @@ To read the value of the `owner` storage variable, which is a single value, we c
 {{#rustdoc_include ../listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo:read_owner}}
 ```
 
-To read the value of the storage variable `names`, which is a mapping from `ContractAddress` to `felt252`, we call the `read` function on the `names` variable, passing in the key `address` as a parameter. If the mapping had more than one key, we would pass in the other keys as parameters as well.
+To read the value of the storage variable names, which is a mapping from `ContractAddress` to `felt252`, we call the `read` function on the entry obtained from `self.names.entry(address)`. The `entry` function is defined in the `StoragePathEntry` trait and returns a `StoragePath<Self::Value>`, which allows us to access the value associated with the specified key `address`. If the mapping had more than one key, we would similarly obtain entries for each key and call `read` on them as needed.
```
enitrat, line 46:
```suggestion
To read the value of the storage variable `names`, which is a mapping from `ContractAddress` to `felt252`, we first need to retrieve the entry path for the specific key in the mapping. We do this by calling the `entry` method on the `names` variable, passing in the `address` as a parameter. This gives us access to the specific entry in the mapping.
Once we have the entry path, we can then call the `read` function on it to retrieve the stored value. 
If the mapping had more than one key, we would chain multiple `entry` calls before the final `read`, passing in each key as a parameter.
```

suggestion for clarity: feel free to also improve. Overall, we want to keep things simple, and not refer to corelib internals too much

blocksorcerer: Fixed in 154c8a9

src/ch14-01-contract-storage.md
blocksorcerer: Implemented changes in 154c8a9

enitrat (approved):
LGTM :)

PR #905, opened 2024-08-02
ci: update mdbook-i18n-helpers

enitrat:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/905)


po/tr.po
```diff
@@ -879,7 +879,7 @@ msgid ""
 "[package]\n"
 "name = \"hello_world\"\n"
 "version = \"0.1.0\"\n"
-"edition = \"2024_07\"\n"
+"edition = \"2023_11\"\n"
```
TAdev0, line 882: why reversing edition for this specific package here?

Or why was it initially 2024_07 before your change, I see it's still 2023_11 in this chapter on the book

TAdev0 (approved):

po/tr.po
TAdev0: ok i see when you create a new package with 2.7, you still have 2023_11 edition

PR #906, opened 2024-08-05
feat: add Cairo VM introduction

zmalatrax:
First proposal for the Cairo VM introduction chapter.

It details what will be covered in this sub-book and gives a short introduction of the virtual machine concept in computer science to contextualize the Cairo VM.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/906)


src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
...
+
+## Virtual Machine
+
+Virtual Machines (VMs) are software emulations of physical computers.
+They provide a complete programming environment through an API which
+includes everything required for the correct executions of programs above it.
+
+Every virtual machine API includes an instruction set architecture (ISA)
+in which to express programs. It could be the same instruction set as some
+physical machine, or an artificial one to implement in software and generate
```
enitrat, line 22: >  or an artificial one to implement in software and generate
with a compiler.

This is a bit unclear

src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
...
+The Cairo VM is also a process VM, similar to the JVM:
+
+- A Cairo program `prgm.cairo` is compiled into compilation artifacts `prgm.json`,
+  containing _Cairo bytecode_ (_CASM_, the Cairo instruction set, and extra data).
+- As seen in the [introduction](ch00-00-introduction.md), Cairo Zero directly compiles to CASM
+  while Cairo first compiles to _Sierra_ and then to a safe subset of CASM.
+- The Cairo VM _interprets_ the provided CASM and generates a trace of the program execution.
+
+An ongoing project, [Cairo Native][cairo-native] works on providing
+Sierra to machine code compilation, including JIT, for executing Cairo programs.
```
enitrat, line 48: is cairo native also covering JIT ?

src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
 # Introduction
+
+Ever wondered how your Cairo programs were executed?
+
+Everything happens in the Cairo Virtual Machine, or _Cairo VM_ for short.
```
enitrat, line 5: I think "Everything" is a bit of an exageration here as the Cairo -> CASM step happens outside of it, and proving as well

src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
...
+physical machine, or an artificial one to implement in software and generate
+with a compiler.
+
+Those that emulate an OS are called _System Virtual Machines_, such as Xen and VMWare.
+We're not interested in them here.
+
+The other ones we're interested in are _Process Virtual Machines_.
+They provide the environment needed by a single user-level process.
+
+The most well-known process VM might be the Java Virtual Machine (JVM).
```
enitrat, line 31: I would like to see an emphasis on the fact that the JVM is designed for general-purpose computing, while the Cairo VM is specifically built for generating provable program executions

src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
...
+- Given a Java program `prgm.java`, it is compiled into a class `prgm.class`,
+  containing _Java bytecode_ (JVM instructions and metadata).
+- The JVM verifies that the bytecode is safe to run.
+- The bytecode is either interpreted (slow) or compiled to machine code just in time (JIT, fast).
+- If using JIT, the bytecode is translated to machine code while executing the program.
+
+The Cairo VM is also a process VM, similar to the JVM:
+
+- A Cairo program `prgm.cairo` is compiled into compilation artifacts `prgm.json`,
+  containing _Cairo bytecode_ (_CASM_, the Cairo instruction set, and extra data).
```
enitrat, line 42:
```suggestion
  containing _Cairo assembly_ (_CASM_, the Cairo instruction set, and extra data).
```

CASM -> Cairo ASseMbly ?

src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
...
+- As seen in the [introduction](ch00-00-introduction.md), Cairo Zero directly compiles to CASM
+  while Cairo first compiles to _Sierra_ and then to a safe subset of CASM.
+- The Cairo VM _interprets_ the provided CASM and generates a trace of the program execution.
```
enitrat, line 45: one liner on how  this trace is used for generating proofs, which is a the main purpose of executing on this VM

src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
...
+
+An ongoing project, [Cairo Native][cairo-native] works on providing
+Sierra to machine code compilation, including JIT, for executing Cairo programs.
+
+Even though the high-level flow of both VMs is similar, their actual architectures
+are extremely different. This is because Cairo is built to prove program execution
```
enitrat, line 51: > are extremely different. 

one-liner explanation of what's different: different instruction set, etc etc

src/ch200-introduction.md
```diff
@@ -1 +1,58 @@
...
+rather than merely performing it.
```
enitrat, line 52:
```suggestion
rather than merely executing it.
```

enitrat (commented):
great :)

src/ch200-introduction.md
zmalatrax: "it" refers to "program execution", no? It's a bit unclear, I'd rather reformulate as

```md
This is because Java programs are merely executed
while Cairo ones have their execution proven.
```

gerceboss:
@zkcarter, would like to collaborate in writing the book, can you share your discord or telegram to discuss more?

src/ch200-introduction.md
zmalatrax: Yep, I think there is a bit of semantics here

--> Assembly is a human readable form of some native machine language
--> Bytecode is a code to be interpreted by a virtual machine, which can have an assembly language

The compilation artifacts contains encoded instruction so it's more about bytecode than assembly imo

Might change to something like

```suggestion
containing _Cairo bytecode_ (encoded CASM, the Cairo instruction set, and extra data).
```

see [stackoverflow](https://stackoverflow.com/a/1783167)

NOTE: The compiled contract class (starknet compilation artifacts named the field containing the program instructions `bytecode`

src/ch200-introduction.md
zmalatrax: Indeed, kind of a stretch ; I was referring to the execution phase which happens in the VM but we should remind the context it integrates on

src/ch200-introduction.md
zmalatrax: From their docs, it can either run as AOT or JIT.
JIT being the default mode

To be confirmed

src/ch200-introduction.md
enitrat: good 👍

src/ch200-introduction.md
```diff
@@ -35,21 +39,30 @@ The most well-known process VM might be the Java Virtual Machine (JVM).
...
+  main purpose of Cairo.
 
 An ongoing project, [Cairo Native][cairo-native] works on providing
-Sierra to machine code compilation, including JIT, for executing Cairo programs.
+Sierra to machine code compilation, including JIT and AOT, for executing Cairo programs.
 
 Even though the high-level flow of both VMs is similar, their actual architectures
-are extremely different. This is because Cairo is built to prove program execution
-rather than merely performing it.
+are extremely different: the instruction set, the non-deterministic memory model and
```
enitrat, line 63: >Even though the high-level flow of both VMs is similar, their actual architectures
are extremely different. This is because Cairo is built to prove program execution	are extremely different: the instruction set, the non-deterministic memory model and
the output This is because Java programs are merely executed while
Cairo ones have their execution proven.

i still think this reads oddly

zmalatrax: Agree, might as well remove the last sentence ("This is because Java..."), the fact that Java only do computation while Cairo proves was stated in a previous paragraph of the intro

src/ch200-introduction.md
```diff
@@ -53,16 +52,23 @@ _provable_ general-purpose computing.
...
+</div>
+
 An ongoing project, [Cairo Native][cairo-native] works on providing
 Sierra to machine code compilation, including JIT and AOT, for executing Cairo programs.
 
 Even though the high-level flow of both VMs is similar, their actual architectures
-are extremely different: the instruction set, the non-deterministic memory model and
-the output. This is because Java programs are merely executed while
-Cairo ones have their execution proven.
+are extremely different: the instruction set, the memory model, Cairo non-determinism
```
enitrat, line 70:
```suggestion
are extremely different: the instruction set, the memory model, Cairo's non-determinism
```

enitrat (approved):

PR #907, opened 2024-08-05
feat: add phantom types

### LINKED ISSUE #893: feat: phantom types
a type can be marked with the `#[phantom]` attribute so that no instance of that type can be created. Add an example that does not compile about trying to instantiate a such type in the `Advanced Cairo Features` section.

Iwueseiter:
Resolves #893

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/907)


enitrat (changes requested):

enitrat:
It's hard to both learn about a concept and explain it correctly, so let me help you with that with two examples. I'll let you work your way from there

Example 1: Units of Measurement

We define two phantom types: Meters and Feet.
We create a Distance<T> struct that uses a generic type T.
The add_distances function only works when both distances are of the same type.
This prevents us from accidentally adding meters to feet, which would be a logical error.


Example 2: Connection States

We define Connected and Disconnected as phantom types.
The Connection<T> struct represents a connection in either state.
The send_data function only accepts Connection<Connected>.
This ensures at compile-time that we can't send data on a disconnected connection.


Benefits

Type Safety: Prevents logical errors by catching them at compile-time.
Zero Runtime Cost: Phantom types don't add any data to the struct, so there's no performance impact.
Clear Intentions: Makes the code's intentions clearer to other developers.

enitrat:
up

PR #924, opened 2024-08-22
feat: macros with Scarb

### LINKED ISSUE #884: feat: macros with scarb
scope: 0.75 day

https://x.com/swmansionxyz/status/1815343318529933421?t=e15zu78w0ZgaAgGMAlSzog&s=19

okhaimie-dev:
Closes #884 

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/924)


src/ch11-10-procedural-macros.md
```diff
@@ -0,0 +1,88 @@
...
+2. attributes ( `#[macro]` )
+3. derive ( `#[derive(Macro)]` )
+
+> To use procedural macros in your Cairo project, you need to have the Rust toolchain setup on your machine. This is because Cairo procedural macros are implemented in Rust. To set up Rust, visit rustup.rs and follow the installation instructions for your operating system.
+
+## How to Use an Already Existing Procedural Macro
+
+To use a procedural macros in your project, a developer needs to add the macro as a dependency in their `Scarb.toml` file and then import and use it in their Cairo program.
+
+> You do not need to know Rust to use procedural macros in your Cairo project.
```
enitrat, line 17:
```suggestion
> You do not need to know Rust to use already existing procedural macros in your Cairo project.
```

src/ch11-10-procedural-macros.md
```diff
@@ -0,0 +1,88 @@
...
+
+```toml
```
enitrat, line 19: code examples should come from a package under listings/

src/ch11-10-procedural-macros.md
```diff
@@ -0,0 +1,88 @@
...
+    field: u32,
+}
+
+// This attribute macro would modify the function in some way
+#[macro_two]
+fn main() -> u32 {
+    let: u32 a = ExampleStruct { field: 1 };
+    let: u32 b = ExampleStruct { field: 2 };
+    // This expression macro performs an operation on the two values
+    macro_one!(a.to_value(), b.to_value());
```
enitrat, line 50: prefer using real use cases, like the pow! macro from Alexandria

src/ch11-10-procedural-macros.md
okhaimie-dev: Alright, I'll work more on this

enitrat:
@okhaimie-dev any update?

src/ch11-10-procedural-macros.md
```diff
@@ -0,0 +1,142 @@
+# Procedural Macros
+
+Cairo procedural macros are actually Rust functions that takes Cairo code as input and returns a modified Cairo code as output. The main concept behind this is to grant developers the ability to create custom macros that supports Cairo packages.
```
enitrat, line 3: This is a bit hard as an introduction to macros. Assume that at this point the reader doesn't event know _what_ macros are nor what they could be used for. Upon reading the introduction, a reader should be able to decide whether or not to continue this chapter in ~ 10 seconds

see: https://doc.rust-lang.org/book/ch19-06-macros.html#macros

enitrat (commented):
This is good work, however I would generally make a few changes on the approach:

- The overall tone and page structure is a bit off compared to the rest of the book: mostly descriptive instead of active tone
- The example should be made as succint as possible, only showing off strict requirements to build a macro from A to Z
- There needs to be a clear separation between the three types of macros
- Given the topic is quite complex, we don't want to "info dump" on readers and just let them deal with it. There should be a lifeline that the reader can follow to guide him through the chapter

If that feels like too much (because the issue has been opened for 2 months), you can just ensure the content is there and I will take care of the form.

okhaimie-dev:
> If that feels like too much (because the issue has been opened for 2 months), you can just ensure the content is there and I will take care of the form.

alright, I will work on improving it today, but if I don't get it to a good state, it is fine if you take it from there. thanks

PR #938, opened 2024-08-28
docs: Rework copy trait explanation

### LINKED ISSUE #923: rework: `Copy` trait explanation
> The important thing to note here is that this is a completely free operation because variables are a Cairo abstraction only and because values in Cairo are always immutable.

This line is a bit confusing when it comes to explaining the Copy trait. The copy mechanism plays only when interacting with functions: a value is either _moved_ or _copied_ (or snapped!).

In Rust, copy is a bit different: 
```rust
let x: i32 = 1;
let y:i32 = x;
```

This will allocate 8 bits of memory for both x and y. _Conceptually_, in Cairo, this would only allocate one cell, `x` and `y` being "labels" to this cell. In reality, it's up to the compiler to determine how x and y might be used later and figure out what to do with memory, as Cairo is a high level language. Unless x and y are consumed by something else (a computation, a function...) no memory would even be alocated.

All in all, copy is more of a guardrail mechanism that ensures a value owns no mutable memory, and is thus safe to use without moving it. And the only types in Cairo that do own mutable memory are Dicts and Arrays.

FriendlyLifeguard:
Fixes #923 

Please let me know if any further revision is needed!

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/938)

### DIFF
--- a/src/ch04-01-what-is-ownership.md
+++ b/src/ch04-01-what-is-ownership.md
@@ -88,8 +88,7 @@ Thankfully, this code does not actually compile. Once we have passed the array t
 
 ## The `Copy` Trait
 
-If a type implements the `Copy` trait, passing a value of that type to a function does not move the value. Instead, a new variable is created, referring to the same value.
-The important thing to note here is that this is a completely free operation because variables are a Cairo abstraction only and because _values_ in Cairo are always immutable. This, in particular, conceptually differs from the Rust version of the `Copy` trait, where the value is potentially copied in memory.
+If a type implements the `Copy` trait, passing a value of that type to a function does not move the value. Instead, a new variable is created, referring to the same value. On the other hand, Rust version of the `Copy` trait creates a a new copy of its value in memory. We can view `Copy` trait in Cairo as a guardrail mechanism to ensure data types adhere to Cairo's linear type system and allow existing data types to be referenced without moving them.
 
 All basic types previously described in ["Data Types"][data types] implement by default the `Copy` trait.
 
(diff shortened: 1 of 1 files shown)


src/ch04-01-what-is-ownership.md
```diff
@@ -88,8 +88,7 @@ Thankfully, this code does not actually compile. Once we have passed the array t
 
 ## The `Copy` Trait
 
-If a type implements the `Copy` trait, passing a value of that type to a function does not move the value. Instead, a new variable is created, referring to the same value.
-The important thing to note here is that this is a completely free operation because variables are a Cairo abstraction only and because _values_ in Cairo are always immutable. This, in particular, conceptually differs from the Rust version of the `Copy` trait, where the value is potentially copied in memory.
+If a type implements the `Copy` trait, passing a value of that type to a function does not move the value. Instead, a new variable is created, referring to the same value. On the other hand, Rust version of the `Copy` trait creates a a new copy of its value in memory. We can view `Copy` trait in Cairo as a guardrail mechanism to ensure data types adhere to Cairo's linear type system and allow existing data types to be referenced without moving them.
```
enitrat, line 91:
```suggestion
The `Copy` trait allows simple types to be duplicated by copying felts, without allocating new memory segments. This contrasts with Cairo's default "move" semantics, which transfer ownership of values to ensure memory safety and prevent issues like multiple writes to the same memory cell. `Copy` is implemented for types where duplication is safe and efficient, bypassing the need for move semantics. Types like `Array` and `Felt252Dict` cannot implement `Copy`, as manipulating them in different scopes is forbidden by the type system.
```

Proposing this version that links the Copy trait to the move and ownership transfer semantics, with a specific relation with array and dicts.

FriendlyLifeguard: Agreed!

enitrat (approved):

PR #944, opened 2024-08-29
chore: remove inexistent import generating an error when building book

TAdev0:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/944)


TAdev0:
@enitrat  we can remove this as it doesn't bring anything given that the path to example doesn't exist. It won't help in any way to retrieve the listing it corresponds to (and I don't thin the repo still contains it anyway as when we did the big rework all listings have been cleared)

enitrat (approved):

PR #955, opened 2024-09-03
chore : remove dead listing in mappings chapter

TAdev0:
@enitrat  there is a listing that doesnt point to anything at the end of Mapping Storages chapter. i dont think any listing is required here

Screenshot 2024-09-03 at 10 34 50

also updated an anchor to snake case

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/955)


TAdev0:
@enitrat  ready to merge

enitrat (approved):

PR #956, opened 2024-09-03
chore: remove formal verification chapter

TAdev0:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/956)


enitrat:
I would remove this as non of these are usable in production

TAdev0:
remove the references or the whole chapter?

could be nice to leave the chapter with the short explanation? dont know actually

enitrat:
I don't see the point of talking about formal verification if it's not something you can leverage for your contracts security. If there's any tool easily usable, why not, but I don't think it's the case today

TAdev0:
@enitrat  ready to merge

PR #965, opened 2024-09-07
chore: remove rust related runner

TAdev0:
This can be removed now the rust runner is not used anywhere in the book

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/965)


enitrat:
keeping this as a reference for modifying this file

PR #969, opened 2024-09-08
Update lib.cairo

CountryCousin:
There is no need to import `use core::dict::Felt252Dict;`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/969)


TAdev0:
With the last edition it's required normally

PR #991, opened 2024-09-25
feat: add default variant to storage enum

### LINKED ISSUE #990: fix: add detail on `#[default]` for enums
When using enumerations in a contract's storage, it's important to define a default enumeration variant. We should add a section on that in the enumeration page or contract storage section.

Example:
```rust
#[derive(Drop, starknet::Store, Serde)]
pub enum Answer {
    None,
    Yes,
    No,
}
```
If we use this enum in the storage, reading an uninitialized entry will panic with:
```
Error: ContractError: Execution error: Execution failed. Failure reason: (0x556e6b6e6f776e20656e756d20696e64696361746f723a ('Unknown enum indicator:'), 0x0 ('')).
```
To avoid that, you must define a default variant, to indicate Cairo compiler what to expect in the storage.
```rust
#[derive(Drop, starknet::Store, Serde)]
pub enum Answer {
    #[default]
    None,
    Yes,
    No,
}
```

Some code to play around the concept:
```rust
#[derive(Drop, starknet::Store, Serde)]
pub enum Answer {
    None,
    Yes,
    #[default]
    No,
}

#[starknet::interface]
trait ISimple<T> {
    fn w(ref self: T, k: felt252, t: Answer);
    fn r(ref self: T, k: felt252) -> Answer;
}

#[starknet::contract]
mod c1 {
    use starknet::storage::{
        StoragePointerReadAccess, StoragePointerWriteAccess, StoragePathEntry, Map
    };
    use super::{Answer, ISimple};

    #[storage]
    struct Storage {
        data: Map<felt252, Answer>,
    }

    #[abi(embed_v0)]
    impl SimpleStorage of ISimple<ContractState> {
        fn w(ref self: ContractState, k: felt252, t: Answer) {
            self.data.entry(k).write(t)
        }

        fn r(ref self: ContractState, k: felt252) -> Answer {
            self.data.entry(k).read()
        }
    }

}
```

enitrat:
Closes #990 

@glihm sounds like the most concise way of conveying this info, wdyt?

- - -
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/991)

### DIFF
--- a/listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo
+++ b/listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo
@@ -187,3 +187,4 @@ mod NameRegistry {
     // ANCHOR_END: state_internal
 }
 //ANCHOR_END: all
+
--- a/listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo
+++ b/listings/ch14-building-starknet-smart-contracts/listing_01_reference_contract/src/lib.cairo
@@ -58,8 +58,9 @@ mod NameRegistry {
     //ANCHOR: enum_store
     #[derive(Copy, Drop, Serde, starknet::Store)]
     pub enum RegistrationType {
-        finite: u64,
-        infinite
+        Finite: u64,
+        #[default]
+        Infinite
     }
     //ANCHOR_END: enum_store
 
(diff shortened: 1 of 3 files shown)


TAdev0 (approved):

glihm (approved):
Yep, very nicely written!

An other thing we may mention is that if you don't want a default variant you can instead wrap the type into an option, which could be better in some cases. And as None is the default for option, it will always work.

But may not be necessary, what you have is already a very informative stuff. 👍

PR #995, opened 2024-10-03
chore: add precisions for `#[cairofmt::skip]` attribute usage

TAdev0:
This change is [Reviewable](https://reviewable.io/reviews/cairo-book/cairo-book/995)


enitrat:
probably simpler to replace "for things" by "for items"

TAdev0:
@enitrat  true. Modules, functions and structs are items, but can we consider a statement as an item?

enitrat:
closing to keep the same formulation as rustfmt

PR #1030, opened 2024-11-26
Update ch02-01-variables-and-mutability.md

mananuf:
by default cairo's variables  are immutable and remain immutable, what happens under the hood using 'mut' is that, the corresponding Sierra variable storing its value is first dropped as it's no longer used, and then a new variable is created with the updated value. so we can say the mut keyword is used to make variables "act" in a mutable way. rather than saying "make them mutable"


enitrat:
the _variable_ itself is mutable.

if you compile Cairo to CASM and run on the CairoVM then the memory is write-once
If you compile to native instructions through LLVM then you're interacting with mutable memory

PR #1041, opened 2024-12-02
Update(at method) ch03-01-arrays.md

dimka90:
From the at method section for getting items from an array, the variables 'first' and 'second' were assumed to have been declared as a reference was made to them but they were never declared in the code section. so I declared the first and second variables
let first = *(a.at(0);
let second = *(a.at(1));
I followed the method of dereferencing the array before accessing the values cause that what the rest of the code section used. The same thing can be achieved without dereferencing the array let first = a.at(0);
let second = a.at(1);


dimka90:
@enitrat Could you specify where the change should be made? I'd like to fix it.

PR #1042, opened 2024-12-02
fix: array `at` code example incoherent with explanation

dimka90:
From the at method section for getting items from an array, the variables 'first' and 'second' were assumed to have been declared as a reference was made to them but they were never declared in the code section. so I declared the first and second variables let first = *a.at(0);
let second = *a.at(1);
I followed the method of dereferencing the array before accessing the values cause that what the rest of the code section used. The same result  can be achieved without dereferencing the array  let first = a.at(0);
let second = a.at(1);

### DIFF
--- a/listings/ch03-common-collections/no_listing_04_array_at/src/lib.cairo
+++ b/listings/ch03-common-collections/no_listing_04_array_at/src/lib.cairo
@@ -3,6 +3,10 @@ fn main() {
     a.append(0);
     a.append(1);
 
+    // The first element of the array 'a', accessed using the .at() method.
+    let first = *a.at(0);
+    // The second element of the array 'a', accessed using the .at() method.
+    let second = *a.at(1);
     // using `at()` method
     assert(*a.at(0) == 0, 'item mismatch on index 0');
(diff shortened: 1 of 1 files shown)


listings/ch03-common-collections/no_listing_04_array_at/src/lib.cairo
```diff
@@ -3,6 +3,10 @@ fn main() {
     a.append(0);
     a.append(1);
 
+    // The first element of the array 'a', accessed using the .at() method.
+    let first = *a.at(0);
+    // The second element of the array 'a', accessed using the .at() method.
+    let second = *a.at(1);
```
enitrat, line 9:
```suggestion
    let first = *a.at(0);
    let second = *a.at(1);
```

+ use `first` and `second` in the assertions below

enitrat (commented):
thank you, that's indeed valid!

dimka90:
> thank you, that's indeed valid!

@enitrat Have added the assertion for both variables as suggested.

enitrat (approved):

PR #1044, opened 2024-12-05
Ci/dependency caching

### LINKED ISSUE #1038: ci: cache dependencies
Improve the current CI jobs to cache the dependencies installed:
- asdf (with scarb & starknet-foundry)
- rust
- pnpm and the node modules
- the mdbook dependencies built from source.

Instead of re-installing / building these dependencies on each workflow run, we can just restore a previous time, reducing CI workflow times.

0xdevcollins:
Closes #1038


0xdevcollins:
@enitrat Please review

.github/workflows/ci.yml
```diff
@@ -27,9 +48,31 @@ jobs:
...
+      - name: Cache mdbook dependencies
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: mdbook-${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            mdbook-${{ runner.os }}-
```
enitrat, line 74: This is just caching the rust dependencies, it's strictly equivalent to the step above, thus has no effect

The MDBook Cache Coverage is actually incomplete, as the current cache only covers Cargo registry and git, but misses several important artifacts:
- The downloaded mdbook binary
- The downloaded mdbook-last-changed binary
- The cloned mdbook-quiz-cairo repository
- The generated bindings from cargo make init-bindings


You should consider separating Binary and Build Caches:

- One for pre-built binaries (mdbook, mdbook-last-changed)
- Another for compiled dependencies (cargo installations), which will end up being covered by the `Cache Rust dependencies` step

something like:

```
- name: Cache mdbook binaries
  uses: actions/cache@v3
  with:
    path: |
      ./bin
      ~/.local/bin
      ~/.cargo/bin
    key: mdbook-bins-${{ runner.os }}-v1-${{ hashFiles('.tool-versions') }}

- name: Cache mdbook build artifacts
  uses: actions/cache@v3
  with:
    path: |
      ~/.cargo/registry
      ~/.cargo/git
      ./target
      ./mdbook-quiz-cairo
    key: mdbook-build-${{ runner.os }}-v1-${{ hashFiles('mdbook-cairo/Cargo.lock', 'mdbook-quiz-cairo/Cargo.lock') }}
```

.github/workflows/ci.yml
```diff
@@ -27,9 +48,31 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: rust-${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            rust-${{ runner.os }}-
+
```
enitrat, line 61: I would actually recommend removing this manual rust cache altogether, remove the `setup-rust-cache` action, and replace these two steps by a simple:

```
    - name: Setup Rust cache
      uses: Swatinem/rust-cache@v2
```

enitrat (changes requested):
A good start, but the real benefits of caching are not yet present because we re-run the "install mdbook" step each run. Because it was re-ran, the CI passed.

You can skip it if the cache-hit was true with:
`if: steps.<id>.outputs.cache-hit != 'true'`

enitrat:
I think you need to make sure the cargo binaries are in path, let me see

enitrat:
Looks like we're good here @0xdevcollins, let me know if you have any questions 💯

.github/workflows/ci.yml
```diff
@@ -1,6 +1,9 @@
 name: CI
 
 on:
+  push:
+    branches:
+      - ci/dependency-caching
```
enitrat, line 6:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -13,37 +16,79 @@ jobs:
...
         run: cargo install --path cairo-listings --locked
         shell: bash
 
       - name: Run cairo-listings and generate summary
         run: cairo-listings verify >> $GITHUB_STEP_SUMMARY
         shell: bash
 
+      - name: Clean up mdbook-quiz-cairo
+        run: rm -rf mdbook-quiz-cairo
+        shell: bash
```
enitrat, line 90:
```suggestion
```

enitrat (approved):

PR #1047, opened 2024-12-07
feat: `keccak_syscall`

### LINKED ISSUE #1013: feat: keccak hashes

TAdev0:
Doesn't close #1013  as i remember you wanted to add examples in the Hashes chapter


enitrat:
i think this page is going to be deleted once we have the corelib doc. it's redundant and should not really be there

enitrat (approved):

PR #1048, opened 2024-12-09
fix: code examples not matching reference contract

Thibrac:
- Simplification of the NameRegistry contract to match the topic
- Move NameRegistry to 14.2 so that it is displayed only for its page
- Add windows to display the entire contract in the examples


enitrat (commented):
Great! One last thing, can you remove the `owner`'s storage variable, remove the get_owner* functions, and for the "free function" you could return the total_names base address instead.

This is because the concept of `owner` is no longer useful in this contract, thus we want to cut noise and keep only relevant elements.

Thibrac:
> Great! One last thing, can you remove the `owner`'s storage variable, remove the get_owner* functions, and for the "free function" you could return the total_names base address instead.
> 
> This is because the concept of `owner` is no longer useful in this contract, thus we want to cut noise and keep only relevant elements.

Done :)

enitrat (approved):

Issue #1031, opened 2024-12-01
epic: Cairo 2.9


enitrat:
## Instructions update
- [x] #1032

## Updates of existing content
- [x] #1035
- [x] Hexadecimal formatting with format!("{:x}")

# New content
- [x] Closures
- [x] `#[executable]` attribute and off-chain proofs generation
- [x] #1122

## Extra checks
It seems that we can now create short strings for types other than felt252, e.g `let x:u64 = 'string'`. This might require re-working the types section about short strings

enitrat:
hey @remybar, that's a great idea please follow the architecture of the Rust book https://doc.rust-lang.org/book/ch13-01-closures.html

PR #1051, opened 2024-12-19
feat: closures

remybar:
Refer to https://github.com/cairo-book/cairo-book/issues/1031.

Note: I proposed the implementation of `map` and `filter` for arrays to show how to provide a closure as function argument. Currently, there is a bug in the inlining analysis (the process which determines if a function should be inlined or not), so we need to use `#[inline(never)]` to deactivate it.
For the `filter` implementation, we also need to use an experimental feature added in Cairo 2.9 (associated type specification) as for the complex implementation of `map_or` provided as example here: https://community.starknet.io/t/cairo-2-9-0-is-out/115117.

Maybe I should try to describe a little bit more `Fn` and `FnOnce`, wdyt @enitrat @TAdev0 ?


src/ch11-11-closures.md
```diff
@@ -0,0 +1,69 @@
...
+Closures are anonymous functions you can save in a variable or pass as arguments to other functions. You can create the closure in one place and then call the closure elsewhere to evaluate it in a different context. Unlike functions, closures can capture values from the scope in which they’re defined. We’ll demonstrate how these closure features allow for code reuse and behavior customization.
+
+> Note: Closures were introduced in Cairo 2.9 and are still under development.
+> Some new features will be introduced in future versions of Cairo, so this page will evolve accordingly.
+
+## Syntax
+
+Here is a basic example with two simple closures:
+
+```cairo, noplayground
```
enitrat, line 12:
```suggestion
```cairo
```

listings/ch11-advanced-features/listing_closures/src/lib.cairo
```diff
@@ -0,0 +1,80 @@
...
+        F,
+        +Drop<F>,
+        impl func: core::ops::Fn<F, (T,)>[Output: bool],
+        +Drop<func::Output>,
+    >(
+        self: Array<T>, f: F,
+    ) -> Array<T> {
+        let mut output: Array<T> = array![];
+        for elem in self {
+            if f(elem.clone()) {
```
enitrat, line 36: if elem is of type T and T implements copy: do you need an explicit clone here?

enitrat:
@remybar would be nice if you could add some mentions of Fn and FnOnce indeed. Can't really help with that right now, but I think it's pretty similar to Rusts's.

listings/ch11-advanced-features/listing_closures/src/lib.cairo
remybar: Yes, you're right 👍

enitrat (approved):

enitrat:
awesome 💯

PR #1074, opened 2024-12-25
Update ch204-02-02-range-check.md

ShantelPeters:


src/ch204-02-02-range-check.md
```diff
@@ -1 +1,31 @@
 # Range Check Builtin
+A specialized tool for confirming that numerical values fall inside predetermined ranges is Cairo's built-in Range Check. It guarantees that a value falls inside a certain range, like [0, 2^128) or more stringent constraints, such [0, 2^96], and is also a valid field element (feeling). By delegating range validations to this builtin, Cairo allows for efficient proof creation while also ensuring calculation accuracy without introducing overhead to the application logic. It is a crucial part of Cairo programs' constraint enforcement.
+For example, the Range Check builtin only accepts felts and verify that such a felt is within the range [0, 2**128). A program can write a value to the Range Check builtin only if those two constraints hold. Those two constraints represent the validation property of the Range Check builtin.
+![range-check-validation-property](https://github.com/user-attachments/assets/aedb56c1-26ec-4d4d-a688-06bf160a2e7c)
```
enitrat, line 4: images should be uploaded locally along with the `.json` export

src/ch204-02-02-range-check.md
```diff
@@ -1 +1,31 @@
 # Range Check Builtin
+A specialized tool for confirming that numerical values fall inside predetermined ranges is Cairo's built-in Range Check. It guarantees that a value falls inside a certain range, like [0, 2^128) or more stringent constraints, such [0, 2^96], and is also a valid field element (feeling). By delegating range validations to this builtin, Cairo allows for efficient proof creation while also ensuring calculation accuracy without introducing overhead to the application logic. It is a crucial part of Cairo programs' constraint enforcement.
```
enitrat, line 2: > A specialized tool

it's not a tool! it's a builtin of the VM.

enitrat, line 2: > and is also a valid field element (feeling)

feeling ?

enitrat, line 2: > It is a crucial part of Cairo programs' constraint enforcement.

what does this mean?

src/ch204-02-02-range-check.md
```diff
@@ -1 +1,31 @@
...
+        Option::None => false
+    }
+}
+
+// Example usage
+let is_valid = validate_range(123);
+assert!(is_valid, 'Value out of range');
+```
+
+This entire operation takes place in the Range Check Segment, which acts similar to a specialized validation section of memory. Only valid numbers are used in your software because they pass through this part automatically.
```
enitrat, line 24: > Only valid numbers are used in your software because they pass through this part automatically.

src/ch204-02-02-range-check.md
enitrat, line 2: > It guarantees that a value falls inside a certain range, like [0, 2^128) or more stringent constraints, such [0, 2^96]

what makes this possible? is there verification done in the same place? are there actually _two_ builtins?

enitrat, line 2: > By delegating range validations to this builtin, Cairo allows for efficient proof creation while also ensuring calculation accuracy without introducing overhead to the application logic.

how much computation overhead does this save? how many instructions would we need if we were using only pure Cairo code?

src/ch204-02-02-range-check.md
```diff
@@ -1 +1,31 @@
...
+fn validate_range(value: felt252) -> bool {
+    match value.try_into() {
+        Option::Some(_): u128 => true,
+        Option::None => false
+    }
+}
+
+// Example usage
+let is_valid = validate_range(123);
+assert!(is_valid, 'Value out of range');
```
enitrat, line 21: this does not demonstrate the use of the range check builtin. if you wanted to illustrate with code, it should probably be with CASM intstructions or Cairo Zero code directly - but you don't have to

src/ch204-02-02-range-check.md
enitrat: not sure what this means

src/ch204-02-02-range-check.md
```diff
@@ -1 +1,31 @@
...
+```
+
+This entire operation takes place in the Range Check Segment, which acts similar to a specialized validation section of memory. Only valid numbers are used in your software because they pass through this part automatically.
+
+![Range Check (#1060)](https://github.com/user-attachments/assets/6cadb079-e1dd-47b5-83d0-d61b4f84398d)
+
+ In the Cairo VM, the Range Check Segment itself seeks for numerical values as depicted in or the flowchart diagram above.  There are three values depicted: 0x000000000000  (valid, zero), 0xFFFFFFFFFFFF (valid, highest value) and 0x1FFFFFFFFFFFF  (invalid, greater than highest value).  If the value falls outside this range, then it is  considered as an invalid value while any value that falls within the range [0,  2^128-1] is considered of to be valid.
```
enitrat, line 28: > the Range Check Segment itself seeks for numerical values

the range check segment does not seek for anything. you can just write values to it

enitrat (changes requested):
lots of imprecisions that need to be addressed. lots of details missing, too. you're free to keep working on this but I will be opening to other contributors in parallel.

src/ch204-02-02-range-check.md
ShantelPeters: wow, thank you, i will fix this right away

enitrat (changes requested):
### Feedback

1. **Technical Inaccuracies**
- Correctly suggests that there's a single builtins for different ranges, but does not clearly mention which ones are actually in production

2. **Missing Core Components**
- No explanation of the memory segment structure, and expected types
- Missing discussion of the validation rules: this builtin is strictly for validation, and produces no output (simply returns the memory address of the range check segment)
- No implementation references (links welcomed)

3. **Structure Problems**
- Missing illustrative diagrams showing memory layout
- Informal writing style, no structure
- Text is too condensed and not properly structured. No inner titles.

### Specific Recommendations

1. **Add Technical Implementation Section**
- Show validation rule implementation
- Try to give an illustrated example of how this is used during a program run. I write a value in the range check segment; what happens next?

2. **Add Memory Layout Section**
- Describe single cell per instance structure

3. **Add Implementation References**
- Link to various implementations (TS, Rust, Python)
- Document validation rules

4. **Correct Technical Information**
- Two active builtins handling range [0, 2^128) and [0, 2^96) in production
- Instruction cost analysis could benefit from a clearer example with latex formulas


### Style and Organization Issues
- Replace informal queue analogy with technical explanation
- Add proper code examples
- Include diagrams showing:
  - Memory layout and validation flow
- Structure similar to Pedersen chapter

src/ch204-02-02-range-check.md
```diff
@@ -1 +1,52 @@
...
+**Snapshot 2 - Invalid Range Check Segment**:
+
+In the defined cell 2:2, the value is 1, which is valid, the defined cell 2:3, the value is 2^128, which is out of range. If it reads from 2:3, it is going to give an output error because 2^128 is bigger than the defined [0, 2^128).
+
+**Implementation References**
+
+Below are some reference implementations for the Range Check intrinsic:
+- [TypeScript Range Check](https://www.webdevtutor.net/blog/typescript-check-number-is-in-range)
+- [Python Range Check Builtin](https://www.codecademy.com/resources/docs/python/built-in-functions/range)
+- [Rust Range Check Builtin](https://docs.rs/range_check/latest/range_check/)
```
enitrat, line 45: CairoVM references...

src/ch204-02-02-range-check.md
```diff
@@ -1 +1,52 @@
...
+
+- **Execution**: This will happen when the VM reads from the cell or when the following check is executed:
+  - If the value is in the allowed range, it passes the check without problems.
+  - If not, an error will pop up the rest of the program will not be continued.
+
+**Example**
+
+Let us review a program execution by the following memory snapshots:
+
+![Valid Range Check Segment Diagram_](https://github.com/user-attachments/assets/5960b8e5-740d-415b-98b9-5d3e70c45b77)
```
enitrat, line 26: upload in the PR along with the.json file

src/ch204-02-02-range-check.md
```diff
@@ -1 +1,52 @@
...
+
+**Snapshot 1 - Valid Range Check Segment**:
+
+  - Cell 2:0 holds the value 15, which is within the range [0, 2^128).
+  - Cell 2:1 contains the value 93, also within the range [0, 2^128).
+
+Both cells contain valid felts that are within the permitted or defined range; thus, no errors will pop up during the validation process.
+
+![INValid Range Check Segment Diagram 234](https://github.com/user-attachments/assets/c2585808-3985-488e-bde2-c75e0973880e)
```
enitrat, line 35: same

ShantelPeters:
Hi @enitrat please kindly review the requested changes

enitrat:
#1096

PR #1081, opened 2025-01-02
add ecdsa builtin page

### LINKED ISSUE #1061: ECDSA

Symmaque:
Closes #1061


enitrat:
@Symmaque add excalidraw files pls

src/ecdsa-invalid-hash.excalidraw
enitrat, line 1: you can have one single excalidraw scene with all three drawings, grouped in the same .excalidraw file

src/ecdsa-invalid-key.png
enitrat, line 1: i dont really understand the indexing used in the signatures dict: the key is supposed to be the public key, the value is supposed to be the (unique) signature associated to that key

meaning that if you want to verify two signatures for a single public key in the same run, you will need to override the existing signature in the cache

src/ch204-02-03-ecdsa.md
```diff
@@ -1 +1,65 @@
...
+The _ECDSA_ builtin is dedicated to sign a message hash and check the signature with a public key using ECDSA (Elliptic Curve Digital Signature Algorithm).
+
+## Memory organization
+
+The ECDSA builtin has its own segment during a Cairo VM run.
+It has two components : the segment values (storing the public keys and the hashes) and a dictionary mapping the public keys' offsets to the corresponding signatures.
+
+The segment values store the public keys and the hashes which are both Felts.
+When segment values are set, the runner checks that the public key and the hash are valid if there is a corresponding signature in the dictionary.
+If there is no corresponding signature, the runner will sign the hash with the corresponding public key.
```
enitrat, line 12:
```suggestion
The segment values store the public keys and the hashes which are both Felts.
When segment values are set, the runner checks that the public key and the hash are valid if there is a corresponding signature in the dictionary.
If there is no corresponding signature in the dictionary, the runner will throw an error - signatures must be explicitly added inside the signature cache before interacting with the builtin.

```

src/ecdsa-invalid-key.png
Symmaque: No, the key in the dictionary is the **offset** of the public key and not the public key itself

enitrat (approved):

PR #1084, opened 2025-01-03
dev: use owner-gated `transfer_ownership` in component example

realnimish:
`_transfer_ownership` doesn't check the caller address and can allow anyone to transfer ownership.

### DIFF
--- a/listings/ch103-building-advanced-starknet-smart-contracts/listing_03_component_dep/src/counter.cairo
+++ b/listings/ch103-building-advanced-starknet-smart-contracts/listing_03_component_dep/src/counter.cairo
@@ -51,7 +51,7 @@ mod OwnableCounterComponent {
             ref self: ComponentState<TContractState>, new_owner: ContractAddress,
         ) {
             let mut ownable_comp = get_dep_component_mut!(ref self, Owner);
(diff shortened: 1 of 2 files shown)


TAdev0 (changes requested):
hey!
I think you are missing the `ownable_component::Ownable` import to make it compile if you use the public `transfer_ownership` function. 

But maybe this example should be reworked, isnt it redundant to have the `OwnableCounter` impl containing `impl Owner: ownable_component::HasComponent<TContractState>` constraint , meaning the contract that will embed this component will also embed the `ownable_component` component that already has a public `transfer_ownership` function?

enitrat:
if OwnableComponent has an external function `transfer_ownership` then we can use the one from the component instead of defining a new one

enitrat:
> if OwnableComponent has an external function transfer_ownership then we can use the one from the component instead of defining a new one

however because OwnableComponent is a component and not a contract I don't believe you can just embed an implementation. Which means you're right @TAdev0 the contract implementer will want to use `transfer_ownership` from `OwnableCounter`, not `Ownable`.  We can remove `transfer_ownerhsip` (and associated text) without losing any information

PR #1086, opened 2025-01-05
fix: Use scarb `2.9.2` in Hello World chapter

julio4:
Updated hello world chapter to use correct dependencies and show example of modified `Scarb.toml` file for clarity.
Keep `starknet` target makes `cairo-run` fails even without any starknet contract target.
Ensure scarb output is the same as well.


src/ch01-02-hello-world.md
```diff
@@ -106,14 +109,31 @@ The next three lines set the configuration information Scarb needs to compile yo
...
+name = "hello_world"
+version = "0.1.0"
+edition = "2024_07"
+
+[dependencies]
+
+```
+
+{{#label modified-scarb-content}}
+<span class="caption">Listing {{#ref modified-scarb-content}}: Contents of modified _Scarb.toml_</span>
```
TAdev0, line 135: This is not a listing, better to remove the label and listing here?

TAdev0 (changes requested):

src/ch01-02-hello-world.md
julio4: Is scarb-content also not a listing? Should we remove both labels?

enitrat: can you link this to a file under listings/ please? that will avoid future version mishaps

src/ch01-02-hello-world.md
```diff
@@ -158,17 +178,19 @@ From your _hello_world_ directory, build your project by entering the following
 
 ```bash
 $ scarb build
-   Compiling hello_world v0.1.0 (file:///projects/Scarb.toml)
-    Finished release target(s) in 0 seconds
+   Compiling hello_world v0.1.0 (file://hello_world/Scarb.toml)
```
enitrat, line 181: once this is a cairo project is under `listings` we can add an `output.txt` file next to the Scarb.toml with

```
$ scarb build
```

inside, run the `cairo-listings output` command, and reference the output file here

julio4: Is it possible to have two output files in the same directory?

enitrat: might need to modify the output generator with a glob to support all output*.txt files and then do `output-run.txt` and `output-build.txt`

enitrat: https://github.com/cairo-book/cairo-book/pull/1088

julio4:
Let me know if the changes is what was expected.
I didn't push the output*.txt support as I saw #1088, so there's a dependency on it for now

enitrat (dismissed):
💯

PR #1088, opened 2025-01-05
feat(cairo-listings): support multiple outputs

enitrat:


julio4:
On it

julio4 (approved):
Did the same impl locally so LGTM

PR #1093, opened 2025-01-07
Voting contract example

### LINKED ISSUE #1000: feat: update the Voting Contract Example
https://book.cairo-lang.org/ch16-06-01-deploying-and-interacting-with-a-voting-contract.html?highlight=katana#deploying-and-interacting-with-a-voting-contract

Several things should be updated here.

1. The code documentation is done using `/// @dev Core Library Imports for the Traits outside the Starknet Contract` tags - a mix between natspec and rustdoc specification. Adapt to use only rustdoc format. Example:
```rust
use core::starknet::ContractAddress;

/// Trait defining the functions that can be implemented or called by the Starknet Contract
pub trait VoteTrait<T> {
    /// Returns the current vote status
    fn get_vote_status(&self) -> (u8, u8, u8, u8);

    /// Checks if the user at the specified address is allowed to vote
    fn voter_can_vote(&self, user_address: ContractAddress) -> bool;

    /// Checks if the specified address is registered as a voter
    fn is_voter_registered(&self, address: ContractAddress) -> bool;

    /// Allows a user to vote
    fn vote(&mut self, vote: u8);
}
```

Because the trait / interface is already document it, remove code documentation from the implementation.

2. Update all tooling to the most up-to-date version and make sure links are still valid. Example: bump katana
```
$ katana --version
katana katana 1.0.0-alpha.15
```

3. Ensure all commands still work. Ensure the input / output is the same. If not, adapt it with the new required input / output / behavior.

Thibrac:
Closes #1000

- Remove code doc
- Updated .toml for katana use
- Updated starkli commands for katana use


listings/ch103-building-advanced-starknet-smart-contracts/listing_05_vote_contract/Scarb.toml
```diff
@@ -1,8 +1,9 @@
 [package]
 name = "listing_99_12_vote_contract"
-version = "0.1.0"
+version = "1.2.0"
```
enitrat, line 3:
```suggestion
version = "0.1.0"
```

enitrat (approved):

Issue #1059, opened 2024-12-23
Output Builtin


enitrat:

od-hunter:
Hi @enitrat , can I be assigned this?

enitrat:
@od-hunter any update?

od-hunter:
On it @enitrat

Nityam573:
@enitrat Can you assign me on this Issue
I have experience in documentation 
https://app.onlydust.com/u/Nityam573

PR #1094, opened 2025-01-08
doc: Document on CairoVM Output Builtin

od-hunter:
closes [#1059](https://github.com/cairo-book/cairo-book/issues/1059)


src/ch204-02-00-output.md
```diff
@@ -1 +1,70 @@
...
+
+## Cell Organization
+
+The Output builtin follows these rules:
+
+* Each cell must store a felt value - relocatable values (pointers) are not allowed
+* Values must be written to consecutive cells starting at offset 0
+* Attempting to write to a cell at offset n when offset n-1 hasn't been written will result in an error
+* Reading from an unwritten cell will result in an error
+* The total number of outputs must match the number declared when initializing the builtin
```
enitrat, line 25: use complete sentences in paragraph form rather than bullet points

enitrat (changes requested):
After reviewing a bit more carefully, I think there's some content missing regarding mostly the organization of the output_ptr in different pages and the mechanism we can use to attach attributes:

1.  The paging system allows organizing outputs into logical groups with different IDs - the only usage i've seen for this is Starknet where the on-chain data is split in N pages of size `max_page_size = 3800` at most 🤔 it would be investigate a bit more on this to know in which cases this is useful

2. Missing explanation of the attribute system that allows attaching metadata to the output segment - this is from what i've seen used in combination with 1/  to enable custom better deserialization of memory values

3. Incorrect statement about "relocatable values not allowed" - the Output builtin can actually handle both felt values and relocatable values, there's neither a validation nor a deduction propery, you can put any value inside it

4. Im not sure where the "sequential writing" rule is from: can you point me to it exactly? I do believe the VM does allow writing anywhere

src/ch204-02-00-output.md
```diff
@@ -1 +1,57 @@
...
+The Output builtin is dedicated to writing values to an output segment that can be read after the program execution completes. It serves as a way for Cairo programs to communicate results externally.
+
+## Memory Organization
+
+The Output builtin segment accepts both felt values and relocatable values. The segment's memory cells have no validation or deduction properties - any value that can be stored in Cairo memory can be written to an Output builtin cell.
+The segment's memory is organized using a paging system that splits outputs into groups. Each group is identified by a page ID. In Starknet's implementation, this system enforces a maximum page size of 3800 values for on-chain data storage.
+
+## Attribute System
+
+The Output builtin implements an attribute system for memory segment metadata. These attributes, combined with page IDs, define how memory values should be deserialized. The attribute system encodes type information and structural metadata needed for correct interpretation of the segment's contents during proof verification.
```
enitrat, line 12: This lack a bit of accuracy or structure to be able to properly understand the paging system, notably what it could be used for.

src/valid-sequential-output.jpeg
enitrat, line 1: this diagram is not very useful. Perhaps something better would put forward the relation between the output builtin segment and the public memory of a run?

enitrat (changes requested):
Also add some info about the relation between the output builtin used (during a run) and the public memory

enitrat:
##1121

PR #1096, opened 2025-01-10
Add range_check96 builtin

### LINKED ISSUE #1066: Range Check96

saimeunt:
This PR provides initial docs on the range_check96 builtin.

Fix #1066


enitrat:
Thanks @saimeunt - can you make this generic for the RangeCheck builtin, and not its specific instantiation for `96` bits ?

src/ch204-02-08-range-check-96.md
```diff
@@ -1 +1,45 @@
-# Range Check96 Builtin
+# Range Check 96 Builtin
+
+The _Range Check 96_ builtin is used to check whether a field element is in the range \\([0,2^{96}-1]\\).
+It is used when instantiating and comparing the various integer types.
+All arithmetic comparisons use the range check builtin.
+This specific variant operates on a \\(2^{96}-1\\) upper bound while the standard Range Check builtin operates on a \\(2^{128}-1\\) upper bound.
```
enitrat, line 6: If we make it generic you can just say that this builtin comes in two variants: one that operates on a \\(2^{96}-1\\) upper bound and another that operates on a \\(2^{128}-1\\) upper bound.

The entire logic is still the same, so it's not a lot of refactoring. You could even simply specify that for the sake of the example we take the rangecheck96 version and the same principles apply for the rangecheck128, and keep the text the same

src/ch204-02-08-range-check-96.md
```diff
@@ -1 +1,45 @@
...
+
+## Cells organization
+
+The Range Check builtin enforces that all values written to its memory cells are within the range \\([0,2^{96}-1]\\). The memory cells constrained by this builtin are called range-checked cells as every values written to this memory segment must be contained within the range \\([0,2^{96}-1]\\) otherwise an out of range error will be thrown.
+
```
enitrat, line 11: Just need to specify that relocatable value (pointers) cannot be written in this segment.

enitrat (commented):
looks good !

saimeunt:
@enitrat Thanks for your review, I've updated my PR to document the generic RangeCheck builtin instead of the specific RangeCheck96 variant.

enitrat (approved):

enitrat (approved):

PR #1099, opened 2025-01-20
fix(struct): Associated function namespaced using trait

hudem1:
1. Fix a little typo where a trait function was namespaced using the struct instead of the trait.
2. Fix a little grammar typo.


enitrat (approved):
thank you :)

PR #1100, opened 2025-01-20
fix(enums): Improve code readability

hudem1:
Hi, I'm (re-)reading through the Cairo book and come across a few small improvements, such as the one in this suggested PR.

I was wondering if those kinds of suggestions are indeed relevant, and if so, whether it's appropriate to create PRs for changes as minor as this, or if I should maybe group them together to avoid spamming with many small PRs. Please let me know how you'd prefer I handle these!

Thank you :)


enitrat:
I think this will fail the format checks. normally all the snippets are auto formatted.
regarding your question, yes it's very relevant :)

hudem1:
> I think this will fail the format checks. normally all the snippets are auto formatted. regarding your question, yes it's very relevant :)

Oh okay, I should close the PR then ?
Great, thank you, I will create PRs then when I find some potential ideas improvements :)

hudem1:
Oh okay, I should close the PR then ?
Great, thank you, I will create PRs then when I find some potential ideas improvements :)

PR #1106, opened 2025-02-06
add some info about Cairo editions

remybar:
As suggested by FeedTheFed [here](https://discord.com/channels/793094838509764618/1268933187803942913/1336962068439633930), this PR adds some information about each available Cairo editions.


enitrat:
Good initiative - can we have a new appendix page similar to https://doc.rust-lang.org/book/appendix-05-editions.html please ?

PR #1109, opened 2025-02-13
feat: use starkup in installation instructions

enitrat:


supreme2580:
Hi Mathieu can I be assigned to this

supreme2580:
Hi @enitrat can I be assigned to this?

PR #1134, opened 2025-03-22
Updating Bit packing to note that the width is actually 252

JeremyRubin:


enitrat:
hey, I don't think this is true. Do you have anything that contradicts this statement?

src/ch103-01-optimizing-storage-costs.md
```diff
@@ -58,11 +58,11 @@ With these bitwise operators, let's see how to combine two `u8` integers into a
 
 ## Bit-packing in Cairo
 
-The storage of a Starknet smart contract is a map with 2<sup>251</sup> slots, where each slot is a `felt252` which is initialized to 0.
+The storage of a Starknet smart contract is a map with 2<sup>251</sup>+17*2^<sup>192</sup>+1 slots, where each slot is a `felt252` which is initialized to 0.
 
 As we saw earlier, to reduce gas costs due to storage updates, we have to use as few bits as possible, so we have to organize stored variables by packing them.
 
-For example, consider the following `Sizes` struct with 3 fields of different types: one `u8`, one `u32` and one `u64`. The total size is 8 + 32 + 64 = 104 bits. This is less than a slot size (i.e 251 bits) so we can pack them together to be stored into a single slot.
+For example, consider the following `Sizes` struct with 3 fields of different types: one `u8`, one `u32` and one `u64`. The total size is 8 + 32 + 64 = 104 bits. This is less than a slot size (i.e 252 bits) so we can pack them together to be stored into a single slot.
```
enitrat, line 65:
```suggestion
For example, consider the following `Sizes` struct with 3 fields of different types: one `u8`, one `u32` and one `u64`. The total size is 8 + 32 + 64 = 104 bits. This is less than a slot size (i.e the `felt252` size) so we can pack them together to be stored into a single slot.
```

This would probably be more accurate

src/ch103-01-optimizing-storage-costs.md
```diff
@@ -58,11 +58,11 @@ With these bitwise operators, let's see how to combine two `u8` integers into a
 
 ## Bit-packing in Cairo
 
-The storage of a Starknet smart contract is a map with 2<sup>251</sup> slots, where each slot is a `felt252` which is initialized to 0.
+The storage of a Starknet smart contract is a map with 2<sup>251</sup>+17*2^<sup>192</sup>+1 slots, where each slot is a `felt252` which is initialized to 0.
```
enitrat, line 61:
```suggestion
The storage of a Starknet smart contract is a map with 2<sup>251 slots, where each slot is a `felt252` which is initialized to 0.
```

JeremyRubin: if it were 2^251, it would imply not all felts are storable?

src/ch103-01-optimizing-storage-costs.md
JeremyRubin: felt252 size doesn't quite mean anything clear to me

src/ch103-01-optimizing-storage-costs.md
enitrat: yes, see also: https://github.com/starkware-libs/cairo/blob/b663c848d61fe6f082037d3bbf53a650c2eeb10e/corelib/src/starknet/storage_access.cairo#L44-L53

> so i can't e.g., use a poseidon hash for a slot? 

you can, and that's what is done fore corelib types, using pedersen in this example.
https://github.com/starkware-libs/cairo/blob/b663c848d61fe6f082037d3bbf53a650c2eeb10e/corelib/src/starknet/storage.cairo#L397-L409

https://github.com/starkware-libs/cairo/blob/b663c848d61fe6f082037d3bbf53a650c2eeb10e/corelib/src/starknet/storage.cairo#L387-L395

upon finalization, if the value is > 2**251, it wraps around at 0.

enitrat: the `-256` offset only applies to the concept of "StorageBase" - which is the "base address" of a variable in storage. It's a high-level concept made to accomodate storing types of different sizes in storage.

The idea is that given a variable T of size N, T's elements can only be stored in sequential addresses if N < 256 (to reduce collision risks). Thus, given this limit, for all of T's elements to fit in sequential addresses, the base address of T (address of its first element) must not be greater than the maximum storage address - 256

PR #1155, opened 2025-04-24
Update ch01-02-hello-world.md

blurbeast:
i could complete the file here if permitted, while reading i think for newbie it would be best we start with Cairo Test and not something Voluminous , with that we can revamp the beginner friendly as stated in the beginning of the book ... we all know Starknet foundry is a really reliable framework but for beginner i think starting with the Cairo Test will give a strong pathway to the learning , hence , what i am saying is that some setup aren't really needed for start especially for beginners . kindly review maintainer.

### DIFF
--- a/src/ch01-02-hello-world.md
+++ b/src/ch01-02-hello-world.md
@@ -59,11 +59,11 @@ You will be given two options :
   Cairo Test
 ```
 
-In general, we'll prefer using the first one `❯ Starknet Foundry (default)`.
+In general, we'll prefer using the second one `❯ Cairo Test`.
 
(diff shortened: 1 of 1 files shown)


enitrat:
Hi @blurbeast, thanks for your interest!

Unfortunately, although easier to start with, Cairo Test is lacking a lot of useful features for a testing framework, that are only covered by Starknet Foundry. I believe that Cairo Test was (initially) only meant for testing internals while developing the language itself; and I think it would be confusing to switch test runners in the middle of the book.

blurbeast:
okay thank you for the clarification .

PR #1160, opened 2025-05-07
fix: update "Upgradeable Component" example to use last oz version

ncoquelet:
Hi @enitrat,

A small PR, to upgrade update "Upgradeable Component" example to use OZ 1.0.0 instead 0.19.0.
Let me know if all goes well for you ser
Have a nice day


enitrat:
hey @ncoquelet I think you will also need to edit the listings/ch103-building-advanced-starknet-smart-contracts/listing_07_oz_upgrade/Scarb.toml file with the right OZ version!

PR #1176, opened 2025-06-20
docs.cairo-lang.org migration

LandauRaz:
Migrated https://docs.cairo-lang.org/ to the title page.

### DIFF
--- a/src/title-page.md
+++ b/src/title-page.md
@@ -4,6 +4,20 @@ _By the Cairo Community and its [contributors](https://github.com/cairo-book/cai
 
 This version of the text assumes you’re using [Cairo](https://github.com/starkware-libs/cairo) [version 2.11.4](https://github.com/starkware-libs/cairo/releases) and [Starknet Foundry](https://foundry-rs.github.io/starknet-foundry/index.html) [version 0.39.0](https://github.com/foundry-rs/starknet-foundry/releases). See the [Installation](ch01-01-installation.md) section of Chapter {{#chap getting-started}} to install or update Cairo and Starknet Foundry.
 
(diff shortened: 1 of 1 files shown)


src/title-page.md
```diff
@@ -4,6 +4,20 @@ _By the Cairo Community and its [contributors](https://github.com/cairo-book/cai
 
 This version of the text assumes you’re using [Cairo](https://github.com/starkware-libs/cairo) [version 2.11.4](https://github.com/starkware-libs/cairo/releases) and [Starknet Foundry](https://foundry-rs.github.io/starknet-foundry/index.html) [version 0.39.0](https://github.com/foundry-rs/starknet-foundry/releases). See the [Installation](ch01-01-installation.md) section of Chapter {{#chap getting-started}} to install or update Cairo and Starknet Foundry.
 
-While reading this book, if you want to experiment with Cairo code and see how it compiles into Sierra (Intermediate Representation) and CASM (Cairo Assembly), you can use the [cairovm.codes](https://cairovm.codes/) playground.
```
enitrat, line 7: can we keep this ?

LandauRaz: I removed it because:

1. I think it's better to redirect to [the Cairo Playground](https://www.cairo-lang.org/cairovm/) because it has instructions and it's better SEO-wise
2. The Cairo playground is mentioned as one of the additional resources, so I felt weird mentioning it twice

WDYT about adding this as a note under the playground's description? E.g., something like:

* [The Cairo Playground](https://www.cairo-lang.org/cairovm/): A browser-based playground for Cairo, enabling to explore and experiment with Cairo by writing, compiling, debugging, and proving Cairo code without any setup
    > Note: You can use the Cairo Playground to experiment with the book's code snippets and see how they compile into Sierra (Intermediate Representation) and CASM (Cairo Assembly).

enitrat: ah, yes I was not aware they had the Cairo Playground now, sounds good

PR #1193, opened 2025-07-13
Add note on PartialEq usage in ch05-02

sidarth16:
This PR adds a note about the `PartialEq` trait in the `ch05-02` section, where `Into` and `TryInto` conversions are discussed. The note clarifies why `PartialEq` is required for the `assert_eq!` comparisons used in the provided examples giving more clarity to their usage . 

---

## Motivation
The examples involving `assert!(result == expected)` will not compile unless `PartialEq` is derived for the struct being compared. This might confuse new learners, especially since `PartialEq` is not explained prior to this usage. This note bridges that gap and provides a link to the official Cairo book appendix for more information.

---

## What’s Added
A `Note` block is inserted **between** the `Into` and `TryInto` sections:

> **Note**  
> The `PartialEq` trait allows for comparison between instances of a type for equality.  
> When derived on structs, two instances are equal only if all their fields are equal.  
> For more details, see [PartialEq for Equality Comparisons](https://book.cairo-lang.org/appendix-03-derivable-traits.html#:~:text=PartialEq%20for%20Equality%20Comparisons)


enitrat:
Thanks, I think however it would be better to have this directly in 

https://www.starknet.io/cairo-book/ch05-01-defining-and-instantiating-structs.html

after the first example and simply redirect to https://www.starknet.io/cairo-book/appendix-03-derivable-traits.html

enitrat:
the reason for that is that as soon as we introduce structs we mark them with `#[derive(Drop)]`, thus I think that this would be the right place for it

sidarth16:
✅ Done — Added the requested note on derivable traits right after the first example in ch05-01, as suggested:

> Note :
> You can derive additional traits on structs, such as PartialEq for comparison and Debug for debug-printing.
> See the [Appendix on Derivable Traits](https://www.starknet.io/cairo-book/appendix-03-derivable-traits.html) for a complete list and examples.

Also updated the ch05-02 note to briefly mention that PartialEq is a derivable trait and point learners to the appendix, keeping the explanation lightweight and centralized.

Let me know if you'd like any rewording or formatting changes!

src/ch05-01-defining-and-instantiating-structs.md
```diff
@@ -14,6 +14,10 @@ To define a struct, we enter the keyword `struct` and name the entire struct. A
 
 <span class="caption">Listing {{#ref user-struct}}: A `User` struct definition</span>
 
+> **Note :**  
+> You can derive multiple traits on structs, such as `Drop`, `PartialEq` for comparison and `Debug` for debug-printing.  
+> See the [Appendix on Derivable Traits](https://www.starknet.io/cairo-book/appendix-03-derivable-traits.html) for a complete list and examples.
```
enitrat, line 19:
```suggestion
> See the [Appendix on Derivable Traits](./appendix-03-derivable-traits.md) for a complete list and examples.
```

PR #1207, opened 2025-07-31
Update ch01-03-proving-a-prime-number.md

saniksin:
### Description of the change

Added a note under the `scarb new` command section to instruct users to select **Cairo Test** as the test runner during project setup.

This is important because **if you don’t select Cairo Test**, the following line will be missing from the generated `Scarb.toml`:

```toml
cairo_test = "2.11.4"
```

### DIFF
--- a/src/ch01-03-proving-a-prime-number.md
+++ b/src/ch01-03-proving-a-prime-number.md
@@ -14,8 +14,15 @@ Open a terminal in your projects directory and create a new Scarb project:
 
 ```bash
 scarb new prime_prover
+```
+
+>🛠️ During the setup, you’ll be prompted to choose a test runner. Just select Cairo Test.
+```
 cd prime_prover
 ```
+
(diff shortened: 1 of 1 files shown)


saniksin:
@enitrat 

Hi, just wanted to explain the reasoning behind my PR.

In Chapter 1.2, it says it’s preferable to select Starknet Foundry (default) — so that’s what I chose. I followed the book exactly as written.

But then, in the next chapter (1.3), I was surprised that cairo_test = "2.11.4" wasn't present in my Scarb.toml, and I couldn't run tests. It wasn’t clear that this chapter requires Cairo Test to be selected during scarb new.

That’s why I proposed adding a small note — just to avoid this confusion for others who are carefully following the guide step by step. Without that hint, it feels like something’s broken or missing.

enitrat:
ok got it, in that case it doesn't really matter because no test is being run, but we could modify the Scarb file to have the snforge lines instead 👍

PR #1263, opened 2026-07-27
docs: ground the storage packing argument in real gas costs

enitrat:
## Why

`ch103-01` argues for struct packing qualitatively — storage "costs gas", with no sense of how much. It also never stated the fact that actually justifies packing: **a slot is billed as a whole, regardless of how many of its bits you use.**

### DIFF
--- a/src/ch103-01-optimizing-storage-costs.md
+++ b/src/ch103-01-optimizing-storage-costs.md
@@ -7,9 +7,19 @@ expensive.
 
 When writing Cairo smart contracts, it is important to optimize storage usage to
 reduce gas costs. Indeed, most of the cost associated with a transaction is
-related to storage updates; and each storage slot costs gas to write to. This
-means that by packing multiple values into fewer slots, you can decrease the gas
-cost incurred by the users of your smart contract.
+related to storage updates, and each slot is billed as a whole, no matter how
+many of its bits you actually use.
+
+As of Starknet v0.14.3, reading a slot costs 24,000 L2 gas and overwriting a
+slot that already holds a value costs 60,000. The first write to a slot that has
+never been written is by far the most expensive at 462,000, as it permanently
+grows the state that every node has to keep. See the [fee documentation][fees]
+for current figures.
+
+This means that by packing multiple values into fewer slots, you can decrease
+the gas cost incurred by the users of your smart contract.
+
+[fees]: https://docs.starknet.io/learn/protocol/fees/
 
 ## Integer Structure and Bitwise Operators
 
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -16149,9 +16149,19 @@ expensive.
 
 When writing Cairo smart contracts, it is important to optimize storage usage to
 reduce gas costs. Indeed, most of the cost associated with a transaction is
-related to storage updates; and each storage slot costs gas to write to. This
-means that by packing multiple values into fewer slots, you can decrease the gas
-cost incurred by the users of your smart contract.
+related to storage updates, and each slot is billed as a whole, no matter how
+many of its bits you actually use.
+
+As of Starknet v0.14.3, reading a slot costs 24,000 L2 gas and overwriting a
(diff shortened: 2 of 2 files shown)


src/ch103-01-optimizing-storage-costs.md
```diff
@@ -7,9 +7,36 @@ expensive.
 
 When writing Cairo smart contracts, it is important to optimize storage usage to
 reduce gas costs. Indeed, most of the cost associated with a transaction is
-related to storage updates; and each storage slot costs gas to write to. This
-means that by packing multiple values into fewer slots, you can decrease the gas
-cost incurred by the users of your smart contract.
+related to storage updates. Starknet prices storage access to reflect what it
+actually costs the network to prove a state update, and proving a single read or
+write means proving a claim along an entire leaf-to-root path of the state tree.
+The result is that touching storage dwarfs pure computation.
```
enitrat: odd formulation for this last sentence. can we use simpler terms?

src/ch103-01-optimizing-storage-costs.md
```diff
@@ -7,9 +7,36 @@ expensive.
...
+[fee documentation][fees] when you need current figures.
+
+This means that by packing multiple values into fewer slots, you can decrease
+the gas cost incurred by the users of your smart contract. The saving is largest
+when it lets a contract avoid writing to slots it has never used before.
+
+[snip37]:
+  https://community.starknet.io/t/snip-37-revisit-storage-access-cost/116143
+[snip40]: https://community.starknet.io/t/snip-40-more-frequent-blocks/116203
+[fees]: https://docs.starknet.io/learn/protocol/fees/
```
enitrat: remove links and historical references. just keep the [fee documentation] link. simplify to just mention the actual current gas costs for read, write-first, overwrite, be succint

src/ch103-01-optimizing-storage-costs.md
```diff
@@ -75,8 +102,11 @@ the following example:
 The storage of a Starknet smart contract is a map with 2<sup>251</sup> slots,
 where each slot is a `felt252` which is initialized to 0.
 
-As we saw earlier, to reduce gas costs due to storage updates, we have to use as
-few bits as possible, so we have to organize stored variables by packing them.
+As we saw earlier, storage is billed per slot rather than per bit, so what we
+are really minimizing is the number of slots a contract touches. Packing several
+variables into one slot turns what would have been several reads and writes into
+a single one — and, for state a contract has not written before, avoids paying
+the much steeper first-write cost several times over.
```
enitrat: previous formulation was fine

src/ch103-01-optimizing-storage-costs.md
enitrat: Simplified. The leaf-to-root proving explanation is gone — the paragraph now just says most of a transaction's cost is storage updates and that each slot is billed as a whole regardless of how many bits you use.

src/ch103-01-optimizing-storage-costs.md
enitrat: Done — SNIP links and all historical framing removed, only `[fee documentation]` remains, and it now just states the three current numbers.

One substantive result from this: I can now give the first-write cost as a concrete number. My earlier hedge ("close to an order of magnitude more") was because SNIP-40 raised read/overwrite by +33% but never restated the new-cell write, so I had no sourced current value for it.

The blockifier versioned constants resolve it. A first write is the overwrite cost plus a state-allocation surcharge, `allocation_cost.gas_cost.l2_gas = 402_000`, which is identical in `blockifier_versioned_constants_0_14_2/3/4.json`. That reproduces SNIP-37 exactly — 45,000 + 402,000 = 447,000, its published figure — which confirms the composition. Under v0.14.3 it is therefore 60,000 + 402,000 = **462,000**, which is what the text now says.

src/ch103-01-optimizing-storage-costs.md
enitrat: Reverted to the original wording verbatim.

